PR #172, opened 2025-01-27
feat: resend failed certificate

### LINKED ISSUE #47: aggsender: check continously if a cert have an error
We have removed a second goroutine that checked the status of last submitted certificate. It make sense that allow to run this check in background and decide if it's fails send a new one inmediatly

joanestebanr:
Fixes #47

## Description
- The `aggsender`is going to check periodically the status of certificates (`AggSender.CheckStatusCertificateInterval`) and rebuild certificate as soon as `InError`is detected.
- To avoid spamming Agglayer implements a rate limiter
- To avoid submit a certificate too close to end of a epoch avoid to send certificates after reaching a percentage of epoch.

## Configuration
```
[AggSender]
CheckStatusCertificateInterval = "5m"
```
- Rate limit configuration:
```
[AggSender.MaxSubmitCertificateRate]
		NumRequests = 20
		Interval = "1h"
```
It can be disabled setting `NumRequests` to 0

- Configuration of percentatge that we can't submit a new certificate to agglayer:
```
[AggSender]
MaxEpochPercentageAllowedToSendCertificate=80
```
A value of 0 or > 100 disable this check

### DIFF
--- a/aggsender/aggsender.go
+++ b/aggsender/aggsender.go
@@ -345,7 +345,8 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayer.SignedCertif
 	return signedCertificate, nil
 }
 func (a *AggSender) isAllowedSendCertificateEpochPercent() bool {
-	if a.cfg.ForbiddenSendCertificateAfterEpochPercentage == 0 || a.cfg.ForbiddenSendCertificateAfterEpochPercentage >= maxPercent {
+	if a.cfg.ForbiddenSendCertificateAfterEpochPercentage == 0 ||
+		a.cfg.ForbiddenSendCertificateAfterEpochPercentage >= maxPercent {
 		return true
 	}
 	status := a.epochNotifier.GetEpochStatus()
--- a/aggsender/aggsender.go
+++ b/aggsender/aggsender.go
@@ -297,7 +297,7 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayer.SignedCertif
 			rateLimitSleepTime.String(), a.rateLimiter.String())
 		time.Sleep(*rateLimitSleepTime)
 	}
-	if a.isAllowedSendCertificateEpochPercent() {
+	if !a.isAllowedSendCertificateEpochPercent() {
 		return nil, fmt.Errorf("forbidden to send certificate due epoch percentage")
 	}
 
--- a/aggsender/aggsender.go
+++ b/aggsender/aggsender.go
@@ -132,7 +132,7 @@ func (a *AggSender) Start(ctx context.Context) {
 	a.log.Info("AggSender started")
 	a.status.Start(time.Now().UTC())
 	a.checkInitialStatus(ctx)
-	a.sendCertificates(ctx)
+	a.sendCertificates(ctx, 0)
 }
 
 // checkInitialStatus check local status vs agglayer status
@@ -159,7 +159,7 @@ func (a *AggSender) checkInitialStatus(ctx context.Context) {
 }
 
 // sendCertificates sends certificates to the aggLayer
-func (a *AggSender) sendCertificates(ctx context.Context) {
+func (a *AggSender) sendCertificates(ctx context.Context, returnAfterNIterations int) {
 	var checkCertChannel <-chan time.Time
 	if a.cfg.CheckStatusCertificateInterval.Duration > 0 {
 		checkCertTicker := time.NewTicker(a.cfg.CheckStatusCertificateInterval.Duration)
@@ -172,9 +172,11 @@ func (a *AggSender) sendCertificates(ctx context.Context) {
 
 	chEpoch := a.epochNotifier.Subscribe("aggsender")
 	a.status.Status = types.StatusCertificateStage
+	iteration := 0
 	for {
 		select {
 		case <-checkCertChannel:
+			iteration++
 			a.log.Debugf("Checking perodical certificates status (%s)",
 				a.cfg.CheckCertConfigBriefString())
 			checkResult := a.checkPendingCertificatesStatus(ctx)
@@ -190,7 +192,12 @@ func (a *AggSender) sendCertificates(ctx context.Context) {
 					a.log.Infof("Appears an InError cert but skipping send cert because RetryCertInmediatlyAfterInError is false")
 				}
 			}
+			if returnAfterNIterations > 0 && iteration >= returnAfterNIterations {
+				a.log.Warnf("reached number of iterations, so we are going to return")
+				return
+			}
 		case epoch := <-chEpoch:
+			iteration++
 			a.log.Infof("Epoch received: %s", epoch.String())
 			checkResult := a.checkPendingCertificatesStatus(ctx)
 			if !checkResult.thereArePendingCerts {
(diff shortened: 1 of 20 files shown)


aggsender/aggsender.go
```diff
@@ -245,6 +296,12 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayer.SignedCertif
 		return nil, fmt.Errorf("error signing certificate: %w", err)
 	}
 
+	if rateLimitSleepTime := a.rateLimiter.Call("sendCertificate", false); rateLimitSleepTime != nil {
```
goran-ethernal, line 299: This rate limiting and two separate routines of resending `InError` certificates and hitting the correct block in epoch mean that we might send two same certificates one after the other. For example, someone configures rate limit of 2 in 2 minutes, and you see an `InError` certificate, you send it once. Agglayer, sets it again as `InError`, you immediately send it once more (agglayer marks it again as `InError`), hitting the rate limit. Now a new epoch comes, you try to resend it again, since rate limit is reached, you will sleep, but in the mean time, you can also be trying to resend it again using the `checkCertStatus` channel. In the meantime sleep passes for the previous call, it sends it, and agglayer settles it. Sleep passes to the last call, you send it, it marks it as `InError` since you tried to send a wrong height, since the previous call settled it.

I don't know what happens to our logic then, if the last sent certificate was in wrong height.

aggsender/aggsender.go
```diff
@@ -245,6 +296,12 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayer.SignedCertif
...
+		a.log.Warnf("rate limit reached , next cert %s can be submitted after %s so sleeping. Rate:%s",
+			certificate.ID(),
+			rateLimitSleepTime.String(), a.rateLimiter.String())
+		time.Sleep(*rateLimitSleepTime)
```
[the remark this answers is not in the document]
joanestebanr: I agree, what do you suggest? instead of sleeping returning an error?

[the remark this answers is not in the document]
goran-ethernal: Hmm, I am thinking about how can we reuse the epoch notifier for this, and not have this second channel for checking the `InError` certificates.

What do you think about this:

- epoch notifier keeps notifying us when every configured percentage of epoch is finished. But this needs to be a low number, like 10, 20%, and it will give us some object that will contain in which epoch we are, and which block in epoch we are now.
- in the `sendCertificates` loop, we will listen to this channel, and on every notification of the channel, we will check if there are `InError` certificates. If there are, try to resend it.
- if there are no `InError` certificates, we will try to send a new certificate, but to do this, we need to keep track in memory, if we already sent a new certificate in this epoch, by saving the data that the channel provides (epoch number, block number, etc). This way we do not need the rate limiter and two channels.

aggsender/aggsender.go
joanestebanr: You can't send a cert in parallel (only 1 goroutine). Until you don't finish the sleep we are not comming back to main `sendCertificates` select

aggsender/aggsender.go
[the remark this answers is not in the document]
joanestebanr: I implemented a threshold over that you can't send certificate until next epoch

aggsender/aggsender.go
```diff
@@ -146,15 +159,48 @@ func (a *AggSender) checkInitialStatus(ctx context.Context) {
...
 		select {
+		case <-checkCertChannel:
+			iteration++
+			a.log.Debugf("Checking perodical certificates status (%s)",
+				a.cfg.CheckCertConfigBriefString())
+			checkResult := a.checkPendingCertificatesStatus(ctx)
+			if !checkResult.thereArePendingCerts && checkResult.appearsNewInErrorCert {
+				if a.cfg.RetryCertAfterInError {
+					a.log.Infof("Appears an InError cert sending a new one (%s)", a.cfg.CheckCertConfigBriefString())
+					_, err := a.sendCertificate(ctx)
```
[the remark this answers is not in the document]
joanestebanr: I implemented a threshold over that you can't send certificate until next epoch

aggsender/aggsender.go
```diff
@@ -245,6 +291,16 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayer.SignedCertif
...
 	}
 
+	if rateLimitSleepTime := a.rateLimiter.Call("sendCertificate", false); rateLimitSleepTime != nil {
+		a.log.Warnf("rate limit reached , next cert %s can be submitted after %s so sleeping. Rate:%s",
+			certificate.ID(),
+			rateLimitSleepTime.String(), a.rateLimiter.String())
+		time.Sleep(*rateLimitSleepTime)
+	}
+	if a.isAllowedSendCertificateEpochPercent() {
+
```
goran-ethernal, line 301: Are you missing a call here?

ToniRamirezM (approved):

aggsender/aggsender.go
```diff
@@ -146,15 +154,48 @@ func (a *AggSender) checkInitialStatus(ctx context.Context) {
...
+			if !checkResult.existPendingCerts && checkResult.existNewInErrorCert {
+				if a.cfg.RetryCertAfterInError {
+					a.log.Infof("Appears an InError cert sending a new one (%s)", a.cfg.CheckCertConfigBriefString())
+					_, err := a.sendCertificate(ctx)
+					a.status.SetLastError(err)
+					if err != nil {
+						a.log.Error(err)
+					}
+				} else {
+					a.log.Infof("Appears an InError cert but skipping send cert because RetryCertInmediatlyAfterInError is false")
```
goran-ethernal, line 187:
```suggestion
					a.log.Infof("An InError cert exists but skipping send cert because RetryCertImmediatelyAfterInError is false")
```

aggsender/aggsender.go
```diff
@@ -288,6 +344,18 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayer.SignedCertif
 
 	return signedCertificate, nil
 }
+func (a *AggSender) isAllowedSendCertificateEpochPercent() bool {
```
goran-ethernal, line 347:
```suggestion
// isNearEndOfEpoch checks if we are at the configured end of epoch where we are not allowed to send a certificate
func (a *AggSender) isNearEndOfEpoch() bool {
```

config/default.go
```diff
@@ -219,4 +219,12 @@ MaxCertSize = 8388608
 BridgeMetadataAsHash = true
 DryRun = false
 EnableRPC = true
+CheckStatusCertificateInterval = "5m"
+RetryCertInmediatlyAfterInError = true
+# Don't send certificate over 80% of the epoch
+MaxEpochPercentageAllowedToSendCertificate=80
+	[AggSender.MaxSubmitCertificateRate]
+		NumRequests = 20
+		Interval = "1h"
```
goran-ethernal, line 228: Is 20 certificates in 1h too small?

aggsender/config.go
```diff
@@ -44,6 +45,24 @@ type Config struct {
 	DryRun bool `mapstructure:"DryRun"`
 	// EnableRPC is a flag to enable the RPC for aggsender
 	EnableRPC bool `mapstructure:"EnableRPC"`
+	// CheckStatusCertificateInterval is the interval at which the AggSender will check the certificate status in Agglayer
+	CheckStatusCertificateInterval types.Duration `mapstructure:"CheckStatusCertificateInterval"`
+	// RetryCertAfterInError when a cert pass to 'InError'
+	// state the AggSender will try to resend it inmediatly
```
goran-ethernal, line 51:
```suggestion
	// state the AggSender will try to resend it immediately
```

config/default.go
joanestebanr: I think so,  it must be 1 per 1h, so it's 20 times more. If it's wrong we can adjust  in the future

goran-ethernal (approved):

PR #174, opened 2025-01-28
docs: Update aggOracle docs 

rachit77:
## Description

 Update aggOracle docs to include config explanation

### DIFF
--- a/docs/etherman.md
+++ b/docs/etherman.md
@@ -1,0 +1,15 @@
+# Etherman
+Etherman handles the communication with the network.
+
+## Etherman Configuration
+
+| Parameter          | Type     | Description                                                                                       | Example/Default |
+|:-------------------|:---------|:--------------------------------------------------------------------------------------------------|:----------------|
+| `URL`              | `string` | JSON-RPC URL for the network.                                                                     |                 |
+| `MultiGasProvider` | `bool`   | Use multiple gas providers if true.                                                               | `false`         |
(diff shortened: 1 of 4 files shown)


Stefan-Ethernal (approved):

Stefan-Ethernal:
Just an idea:
probably it would be a good idea to describe the configuration parameters for the `EthTxManager` in some separate doc and reference it from there. The same would go for the `Etherman` (because it is nested in the `EthTxManager`). WDYT?

joanestebanr (commented):
Maybe make sense to set the defaults values and mention that is taking advantage of the common vars like: `L1URL`, `L2URL`,....

rachit77:
> Just an idea: probably it would be a good idea to describe the configuration parameters for the `EthTxManager` in some separate doc and reference it from there. The same would go for the `Etherman` (because it is nested in the `EthTxManager`). WDYT?

Will do the same in a separate PR

rachit77:
> Maybe make sense to set the defaults values and mention that is taking advantage of the common vars like: `L1URL`, `L2URL`,....

I have pointed out to `default.go` file which contains all the default values

docs/ethtxmanager.md
```diff
@@ -0,0 +1,23 @@
+# EthTxManager
+
+EthTxManager is responsible for managing transactions
```
temaniarpit27, line 3: can we add a couple of more lines here.

docs/etherman.md
```diff
@@ -0,0 +1,15 @@
+# Etherman
+Etherman handles the communication with the network.
```
temaniarpit27, line 2: a couple of more lines here

temaniarpit27 (approved):
💯 
a few comments

PR #180, opened 2025-02-04
feat: allows reorg detectors configure the what kind of block is a finalized block

### LINKED ISSUE #176: Syncer must configure finalized Block from reorg detector attached.
- Syncer must ask for finalized block to the reorg detector attached instead of hardcoded value (https://github.com/agglayer/aggkit/blob/e141db20f0fbcc49ad046a12987a244b161b3d60/cmd/run.go#L542) : 
    -  This allow a more compact configuration
    -  The reorg detector + syncer is going to be in the same page
- The rorg detector must have this value in configuration
   -  In some cases L2 RPC doesn't have the 'finalized' Block as L2 and returns always a 0.

Also must be a way to disable reorg detector and the code on evmdownloader that request for latest FinalizedBlock.

joanestebanr:
Fixes #176
## Description
- Reorg detector allow to configure the what kind of block is a finalized block (`ReorgDetectorLX.FinalizedBlock`)
- Syncers use finalizedBlock configuration from reorgDetector
- Fix e2e test for bridgesyncer that sometimes fails because a forced reorg affect to a finalized block

## Configuration
```
[ReorgDetectorL1]
DBPath = "{{PathRWData}}/reorgdetectorl1.sqlite"
FinalizedBlock="FinalizedBlock"

[ReorgDetectorL2]
DBPath = "{{PathRWData}}/reorgdetectorl2.sqlite"
FinalizedBlock="LatestBlock"
```

### DIFF
--- a/bridgesync/e2e_test.go
+++ b/bridgesync/e2e_test.go
@@ -60,12 +60,12 @@ func TestBridgeEventE2E(t *testing.T) {
 		expectedBridges = append(expectedBridges, bridge)
 		expectedRoot, err := setup.L1Environment.BridgeContract.GetRoot(nil)
 		require.NoError(t, err)
-		finalizedBlock := GetFinalizedBlockNumber(t, ctx, setup.L1Environment.SimBackend.Client())
+		finalizedBlock := getFinalizedBlockNumber(t, ctx, setup.L1Environment.SimBackend.Client())
 		log.Infof("*** iteration: %d, Bridge Root: %s latestBlock:%d finalizedBlock:%d", i, common.Hash(expectedRoot).Hex(), bn, finalizedBlock)
 		bridgesSent++
 		bn, err = setup.L1Environment.SimBackend.Client().BlockNumber(ctx)
 		require.NoError(t, err)
-		finalizedBlockNumber := GetFinalizedBlockNumber(t, ctx, setup.L1Environment.SimBackend.Client())
+		finalizedBlockNumber := getFinalizedBlockNumber(t, ctx, setup.L1Environment.SimBackend.Client())
 		blocksToReorg := 1 + i%maxReorgDepth
 		// Trigger reorg but prevent to reorg a finalized block
 		if i%reorgEveryXIterations == 0 && bn-uint64(blocksToReorg) > finalizedBlockNumber {
@@ -101,7 +101,7 @@ func TestBridgeEventE2E(t *testing.T) {
 	// Wait for syncer to catch up
 	time.Sleep(time.Second * 2) // sleeping since the processor could be up to date, but have pending reorgs
 
-	lb := GetFinalizedBlockNumber(t, ctx, setup.L1Environment.SimBackend.Client())
+	lb := getFinalizedBlockNumber(t, ctx, setup.L1Environment.SimBackend.Client())
 	helpers.RequireProcessorUpdated(t, setup.L1Environment.BridgeSync, lb)
 
 	// Get bridges
@@ -127,7 +127,7 @@ func TestBridgeEventE2E(t *testing.T) {
 	require.Equal(t, expectedBridges, actualBridges)
 }
 
-func GetFinalizedBlockNumber(t *testing.T, ctx context.Context, client simulated.Client) uint64 {
+func getFinalizedBlockNumber(t *testing.T, ctx context.Context, client simulated.Client) uint64 {
 	t.Helper()
 	lastBlockFinalityType, err := etherman.FinalizedBlock.ToBlockNum()
 	require.NoError(t, err)
--- a/bridgesync/e2e_test.go
+++ b/bridgesync/e2e_test.go
@@ -7,10 +7,12 @@ import (
 	"time"
 
 	"github.com/agglayer/aggkit/bridgesync"
+	"github.com/agglayer/aggkit/etherman"
 	"github.com/agglayer/aggkit/log"
 	"github.com/agglayer/aggkit/test/helpers"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethclient/simulated"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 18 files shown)


reorgdetector/config.go
```diff
@@ -17,6 +18,11 @@ type Config struct {
 
 	// CheckReorgsInterval is the interval to check for reorgs in tracked blocks
 	CheckReorgsInterval types.Duration `mapstructure:"CheckReorgsInterval"`
+	// FinalizedBlockType indicates the status of the blocks that will be queried in order to sync
+	// if finalizedBlock == "LatestBlock" then is like to disable reorg detector
+	// there are no chances to reorg
+	FinalizedBlock etherman.BlockNumberFinality `jsonschema:"enum=LatestBlock, enum=SafeBlock, enum=PendingBlock, enum=FinalizedBlock, enum=EarliestBlock" mapstructure:"FinalizedBlock"` //nolint:lll
```
ToniRamirezM, line 24: Maybe it is like this to be consistent with other components, but I would remove the "block" from all the options: 
```suggestion
	FinalizedBlock etherman.BlockNumberFinality `jsonschema:"enum=Latest, enum=Safe, enum=Pending, enum=Finalized, enum=Earliest" mapstructure:"Finalized"` //nolint:lll
```

ToniRamirezM (approved):

reorgdetector/config.go
joanestebanr: I agree, but  the rest of components use this name to set the block finality

reorgdetector/reorgdetector.go
```diff
@@ -95,8 +115,24 @@ func (rd *ReorgDetector) Start(ctx context.Context) (err error) {
 	return nil
 }
 
+func (rd *ReorgDetector) String() string {
+	if rd == nil {
+		return "ReorgDetector{nil}"
+	}
+	return fmt.Sprintf("ReorgDetector{%s, %v, %s}",
```
goran-ethernal, line 122: Add field names to the log as well.

goran-ethernal, line 122: Add field names to the log as well.

reorgdetector/config.go
```diff
@@ -17,6 +18,11 @@ type Config struct {
 
 	// CheckReorgsInterval is the interval to check for reorgs in tracked blocks
 	CheckReorgsInterval types.Duration `mapstructure:"CheckReorgsInterval"`
+	// FinalizedBlockType indicates the status of the blocks that will be queried in order to sync
+	// if finalizedBlock == "LatestBlock" then is like to disable reorg detector
```
goran-ethernal, line 22:
```suggestion
	// if finalizedBlock == "LatestBlock" then reorg detector is disabled and we assume the network has no chances of reorgs
```

bridgesync/e2e_test.go
```diff
@@ -92,21 +100,38 @@ func TestBridgeEventE2E(t *testing.T) {
...
+	for i := 119; i >= 00; i-- {
+		root, err := setup.L1Environment.BridgeSync.GetExitRootByIndex(ctx, uint32(i))
+		require.NoError(t, err)
+		log.Infof("DepositCount:%d root: %s", i, root.Hash.Hex())
+	}
 	require.Equal(t, common.Hash(expectedRoot).Hex(), root.Hash.Hex())
 	require.Equal(t, expectedBridges, actualBridges)
 }
+
+func GetFinalizedBlockNumber(t *testing.T, ctx context.Context, client simulated.Client) uint64 {
```
goran-ethernal, line 130: No need to export it:

```suggestion
func getFinalizedBlockNumber(t *testing.T, ctx context.Context, client simulated.Client) uint64 {
```

goran-ethernal (approved):

reorgdetector/config.go
Stefan-Ethernal, line 24: The name of this field is a bit strange, personally I would call it `BlockFinality` (or something like that). WDYT?

reorgdetector/reorgdetector.go
```diff
@@ -36,9 +36,12 @@ type EthClient interface {
 }
 
 type ReorgDetector struct {
-	client             EthClient
-	db                 *sql.DB
-	checkReorgInterval time.Duration
+	client               EthClient
+	db                   *sql.DB
+	checkReorgInterval   time.Duration
+	finalizedBlockName   etherman.BlockNumberFinality
```
Stefan-Ethernal, line 42: WDYT?
```suggestion
	blockFinality   etherman.BlockNumberFinality
```

sync/evmdriver.go
```diff
@@ -46,6 +47,8 @@ type processorInterface interface {
 type ReorgDetector interface {
 	Subscribe(id string) (*reorgdetector.Subscription, error)
 	AddBlockToTrack(ctx context.Context, id string, blockNum uint64, blockHash common.Hash) error
+	GetFinalizedBlock() etherman.BlockNumberFinality
```
Stefan-Ethernal, line 50: WDYT?
```suggestion
	GetBlockFinality() etherman.BlockNumberFinality
```

Stefan-Ethernal (approved):

sync/evmdriver.go
joanestebanr: We can't use BlockFinality because is used for the target block of a syncer, and in this case is which block is considered finalized

reorgdetector/reorgdetector.go
joanestebanr: We can't use BlockFinality because is used for the target block of a syncer, and in this case is which block is considered finalized

reorgdetector/config.go
joanestebanr: We can't use BlockFinality because is used for the target block of a syncer, and in this case is which block is considered finalized

reorgdetector/reorgdetector.go
Stefan-Ethernal: wdyt about `finalizedBlockType`?

PR #183, opened 2025-02-05
feat: send `l1infotree` data to the `aggchain prover`

goran-ethernal:
## Description

This PR implements the additional logic of sending `l1infotree` data to the `aggchain prover` that is needed to generate a `aggchain` (`auth`) proof.

The data includes:
- start block from which to build a proof.
- max end block to which a proof needs to be built.
- last finalized `l1InfoTreeRoot`.
- the leaf with the highest index on that block and given finalized `l1infoTreeRoot`.
- merkle proof of the given leaf on given root.

Based on the provided block range and on the finalized l1 info tree data we send to the prover, prover can either return the maximum block range we asked, or return a smaller range.

But, in order to do this, all the syncers need to store block hash as well to the block table. Before this PR, only `l1infotreesyncer` stored this data, but we might need it in all syncers (to be consistent), so this PR expanded all the syncers to store the block hash as well.

Fixes # (issue)


bridgesync/migrations/migrations_test.go
```diff
@@ -56,6 +58,24 @@ func Test001(t *testing.T) {
...
 	`)
 	require.NoError(t, err)
-	err = tx.Commit()
+	require.NoError(t, tx.Commit())
+}
+
+func Test002(t *testing.T) {
+	t.Parallel()
+
+	dbPath := path.Join(t.TempDir(), "bridgesyncTest001.sqlite")
```
Stefan-Ethernal, line 67:
```suggestion
	dbPath := path.Join(t.TempDir(), "bridgesyncTest002.sqlite")
```

Stefan-Ethernal (approved):

bridgesync/migrations/migrations_test.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/183/commits/ae1f6a1e0f33e9b9aac628d534f8d439654d80a8

joanestebanr (commented):
Why is the reason that you need blockHash to all syncers? It looks like is not used on bridesync neither lastgersync.. isn't?

goran-ethernal:
> Why is the reason that you need blockHash to all syncers? It looks like is not used on bridesync neither lastgersync.. isn't?

To be consistent. It is useful for debugging, if we have issues on other syncers with reorg detection.

aggsender/flow_aggchain_prover.go
```diff
@@ -81,7 +93,7 @@ func (a *aggchainProverFlow) GetCertificateBuildParams(ctx context.Context) (*ty
 				aggchainProof.StartBlock, aggchainProof.EndBlock, aggchainProof.Proof,
 				lastSentCertificateInfo.FromBlock, lastSentCertificateInfo.ToBlock)
 
-			proof = aggchainProof.Proof
+			aggProof = aggchainProof.Proof
 
 			if aggchainProof.EndBlock < lastSentCertificateInfo.ToBlock {
 				// aggchain prover can return a proof for a smaller range than requested
```
joanestebanr, line 99: maybe you need to request again the claims and bridges to include in this cert?

lastgersync/migrations/lastgersync0002.sql
```diff
@@ -0,0 +1,5 @@
+-- +migrate Down
+ALTER TABLE block DROP COLUMN hash;
+
+-- +migrate Up
+ALTER TABLE block ADD COLUMN hash VARCHAR;
```
joanestebanr, line 5: the pre existing data on Database (blocks already there) how are going to populate the hash value for this blocks?

aggsender/flow_aggchain_prover.go
```diff
@@ -3,18 +3,24 @@ package aggsender
...
 	"github.com/agglayer/aggkit/agglayer"
 	"github.com/agglayer/aggkit/aggsender/db"
 	"github.com/agglayer/aggkit/aggsender/grpc"
 	"github.com/agglayer/aggkit/aggsender/types"
+	"github.com/agglayer/aggkit/etherman"
+	treeTypes "github.com/agglayer/aggkit/tree/types"
 	"github.com/ethereum/go-ethereum/common"
 )
 
+var finalizedBlockBigInt = big.NewInt(int64(etherman.Finalized))
```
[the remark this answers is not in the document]
goran-ethernal: It returns an error as well, and having this in a global variable is useful for tests and the call for finalized block.

aggsender/flow_aggchain_prover.go
goran-ethernal: If it is lower, we adjust the bridges and claims on line 117 in `adjustBlockRange` function.

lastgersync/migrations/lastgersync0002.sql
goran-ethernal: Added some logic around this: https://github.com/agglayer/aggkit/pull/183/commits/3cdebe019a3599a7202614a1f1811df1543888a9

joanestebanr (approved):

PR #191, opened 2025-02-07
feat: `bridge_getTokenMappings` endpoint

### LINKED ISSUE #190: Provide bridge_getTokenMappings endpoint
# Description
Implement `bridge_getTokenMappings` JSON RPC endpoint. It should provide all the `NewWrappedToken` events on the specific network (identified by the network id) in descending order of `blockNumber`.
The results should be paginated and determined by the page number and page size

## Example request
```bash
/bridge_getTokenMappings?networkID=1&page=2&pageSize=10
```

## Example response
```json
{
   "tokenMappings":[
      {
         "originTokenNetwork":0,
         "originTokenAddress":"0x928855b2c73a3574684fe7a4d55ae8bae212c92f",
         "wrappedTokenAddress":"0x11136b8af575c1bf11435f14d4a008982bc48b97",
         "transactionHash":"0x1e35755a2e1d4e13fb57380ed2dbc390383d72066bd0316f693029ab96d454c8",
         "timestamp":"2025-01-20T07:50:57.000Z",
         "blockNumber":4315009,
         "metadata":"000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000A00000000000000000000000000000000000000000000000000000000000000012000000000000000000000000000000000000000000000000000000000000000D477265656E20426974636F696E0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044742544300000000000000000000000000000000000000000000000000000000",
         "callData":"0x"
      }
   ],
   "totalTokenMappings":1
}
```

Stefan-Ethernal:
## Description

Expose the JSON RPC endpoint to retrieve the token mappings (`bridge_getTokenMappings`). It relies on indexing the `NewWrappedToken` [event](https://github.com/0xPolygonHermez/zkevm-contracts/blob/21d3fd6ec0881731de49f1a6133fb97ed863a7ab/contracts/v2/PolygonZkEVMBridgeV2.sol#L130-L135) from the bridge contract, which is captured in the local SQL lite db. 

The `bridge_getTokenMappings` provides the paginated token mapping results.

Fixes #190


rpc/bridge.go
```diff
@@ -70,6 +72,65 @@ func NewBridgeEndpoints(
...
+
+// GetTokenMappings returns the token mappings for the given network
+func (b *BridgeEndpoints) GetTokenMappings(networkID uint32, pageNumber, pageSize *uint32) (interface{}, rpc.Error) {
+	b.logger.Debugf("GetTokenMappings invoked (network id=%d, page number=%v, page size=%v)",
+		networkID, pageNumber, pageSize)
+
+	ctx, cancel := context.WithTimeout(context.Background(), b.readTimeout)
+	defer cancel()
+
+	c, merr := b.meter.Int64Counter("get_token_mappings")
```
temaniarpit27, line 89: @Stefan-Ethernal whats the usecase for this counter? do we track anything with this?

common/common.go
```diff
@@ -18,34 +18,49 @@ var (
...
-
-	bytes := make([]byte, uint64ByteSize)
+	bytes := make([]byte, Uint64ByteSize)
 	binary.BigEndian.PutUint64(bytes, num)
 
 	return bytes
 }
 
 // BytesToUint64 converts a byte slice to a uint64
 func BytesToUint64(bytes []byte) uint64 {
```
temaniarpit27, line 35: Any reason why we are making changes here?

temaniarpit27 (approved):
LGTM

common/common.go
Stefan-Ethernal: I have noticed [this](https://github.com/agglayer/aggkit/pull/191/files#diff-18eb3102bf480ea624d689abce486ee67aa3ec80ca545c2d7952ff12dfffeea6L391-L393) private function and decided it is better to reuse the one from the `common` pkg. However the code in the `common.go`  needed some changes (to be reused for both converting the byte array with less than 4 bytes and the one that is exactly 4 bytes in length) and better code coverage and that's how I ended up affecting this logic.

rpc/bridge.go
Stefan-Ethernal: I have just followed what we have in the already existing [functions](https://github.com/agglayer/aggkit/blob/509353b885176e54a0e2f0a8b95e29a53a5646cc/rpc/bridge.go#L80) since we have exposed metrics for the bridge service, which can be useful in the future, once it is being run in the production (to have some metrics that indicate the API invocations rate).

rachit77 (approved):

PR #201, opened 2025-02-10
feat: prometheus

### LINKED ISSUE #195: [FEATURE] Add monitoring using prometheus

ToniRamirezM:
## Description

Adds prometheus metrics endpoint.

Closes #195 

## Related PR on Kurtosis repo:

https://github.com/0xPolygon/kurtosis-cdk/pull/476


prometheus/prometheus.go
```diff
@@ -0,0 +1,677 @@
+package prometheus
```
vcastellm, line 1: From where is this code comming from? Do we need all of this functionalities?

ToniRamirezM: It comes from the metric package in zkevm-node. We just need counters now, but if in the future we need any other kind of metric, it will be already there ready to be used.

goran-ethernal (approved):

joanestebanr (approved):

PR #202, opened 2025-02-10
feat: add support for external signer

### LINKED ISSUE #194: [FEATURE] add support for external signer
have a mechanism to avoid having the private keys directly on the container, specially for AggSender.
- Allow to choose the signing mechanism: local / eth-sign 
- Support consensys KMS as eth-sign

NOTE: 
- To add support to this feature for AggOracle is need to implement it on ethtxManager library (#245)

## References: 
- https://docs.web3signer.consensys.io/reference/api/json-rpc#eth_signtransaction
- https://docs.web3signer.consensys.io/how-to/store-keys
- https://github.com/ethereum/remote-signing-api?tab=readme-ov-file

joanestebanr:
## Description
How to adopt new method of signing: 
- Change the struct type  `types.KeystoreFileConfig` for configuring the key to `signer.SignerConfig`
-  Define the field `Method` to: `local` or `web3signer`. If not set is `local`  
- The config file  are 100% compatible with previous  `types.KeystoreFileConfig`, so no user friction
- An old configuration like this is mapped to `local`: 
```
{Path = "/app/keystore/claimsopnsor.keystore", Password = "testonly"}
```
- New `web3signer`: This use a server web3signer to store the keys an invoke the sign process using a API. To configure it you need to set: 
```
{Method = "web3signer", URL = "http://localhost:8545", Address = "0x1234567890abcdef"}
```
- `Method` (mandatory): sign method
- `URL` (mandatory): web3signer server
- `Address` (optional): Public key to use to sign, if this field is empty it try to guess the key to use, so if the server only have 1 key that this one is used.
- 
Fixes #194


goran-ethernal (approved):

aggsender/config.go
```diff
@@ -59,6 +60,12 @@ type Config struct {
 	MaxEpochPercentageAllowedToSendCertificate uint `mapstructure:"MaxEpochPercentageAllowedToSendCertificate"`
 	// MaxSubmitCertificateRate is the maximum rate of certificate submission allowed
 	MaxSubmitCertificateRate common.RateLimitConfig `mapstructure:"MaxSubmitCertificateRate"`
+
+	Signer SignerConfig
+}
+
+type SignerConfig struct {
+	Method string `jsonschema:"enum=local, enum=SafeBlock, enum=PendingBlock, enum=FinalizedBlock, enum=EarliestBlock" mapstructure:"BlockFinality"` //nolint:lll
```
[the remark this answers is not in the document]
joanestebanr: This was the remains from a previous implementation. deleted!

Stefan-Ethernal (approved):
Overall LGTM, nice.

signer/local.go
```diff
@@ -0,0 +1,122 @@
...
+	return SignerConfig{
+		Method: MethodLocal,
+		Config: map[string]interface{}{
+			FieldPath:     path,
+			FieldPassword: pass,
+		},
+	}
+}
+
+// NewLocalConfig creates a LocalSignerConfig (specific config) from a SignerConfig
```
Stefan-Ethernal, line 42:
```suggestion
// NewLocalConfig creates a KeystoreFileConfig (specific config) from a SignerConfig
```

signer/local.go
```diff
@@ -0,0 +1,122 @@
...
+
+const (
+	MethodLocal   = "local"
+	FieldPath     = "path"
+	FieldPassword = "password"
+)
+
+// // AggsenderPrivateKey is the private key which is used to sign certificates
+// AggsenderPrivateKey types.KeystoreFileConfig `mapstructure:"AggsenderPrivateKey"`
+type LocalSign struct {
```
Stefan-Ethernal, line 23:
```suggestion
type LocalSigner struct {
```

signer/local.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+const (
+	MethodLocal   = "local"
+	FieldPath     = "path"
+	FieldPassword = "password"
+)
+
+// // AggsenderPrivateKey is the private key which is used to sign certificates
+// AggsenderPrivateKey types.KeystoreFileConfig `mapstructure:"AggsenderPrivateKey"`
```
Stefan-Ethernal, line 22: It looks like these comments are by mistake here?
```suggestion
```

signer/local.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	res = types.KeystoreFileConfig{
+		Path:     pathStr,
+		Password: passStr,
+	}
+	return res, nil
+}
+
+// NewLocalSign creates a new LocalSign based on config
+func NewLocalSign(name string, logger commontypes.Logger, file types.KeystoreFileConfig) *LocalSign {
```
Stefan-Ethernal, line 66:
```suggestion
func NewLocalSigner(name string, logger commontypes.Logger, file types.KeystoreFileConfig) *LocalSign {
```

signer/local.go
```diff
@@ -0,0 +1,122 @@
...
+	return &LocalSign{
+		name:   name,
+		logger: logger,
+		file:   file,
+	}
+}
+
+// NewLocalSignFromPrivateKey creates a new LocalSign based on a private key
+func NewLocalSignFromPrivateKey(name string,
```
Stefan-Ethernal, line 75:
```suggestion
func NewLocalSignerFromPrivateKey(name string,
```

signer/local.go
```diff
@@ -0,0 +1,122 @@
...
+func (e *LocalSign) SignHash(ctx context.Context, hash common.Hash) ([]byte, error) {
+	if e.privateKey == nil {
+		return nil, fmt.Errorf("%s private key is nil", e.logPrefix())
+	}
+	return crypto.Sign(hash.Bytes(), e.privateKey)
+}
+
+func (e *LocalSign) PublicAddress() common.Address {
+	return e.publicAddress
+}
```
Stefan-Ethernal, line 115: Nitpick: add new line
```suggestion
}

```

signer/local.go
```diff
@@ -0,0 +1,122 @@
...
+func (e *LocalSign) String() string {
+	return fmt.Sprintf("singer: %s path:%s, pubAddr: %s", e.logPrefix(), e.file, e.publicAddress.String())
+}
+
+func (e *LocalSign) logPrefix() string {
+	return fmt.Sprintf("singer:%s[%s]: ", MethodLocal, e.name)
```
Stefan-Ethernal, line 121:
```suggestion
	return fmt.Sprintf("signer:%s[%s]: ", MethodLocal, e.name)
```

signer/web3signer_sign.go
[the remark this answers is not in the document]
joanestebanr: I agree, but maybe make more snse to  rename local to local_signer?, WDYT?

signer/local.go
joanestebanr: Why do you suggest this? the object is called `LocalSign`

signer/local.go
joanestebanr: The same as before, the produced object is `LocalSign`

signer/web3signer_sign.go
```diff
@@ -0,0 +1,132 @@
...
+		return Web3SignerConfig{}, fmt.Errorf("config %s: field %s is not string %v",
+			MethodWeb3Signer, FieldURL, cfg.Config["url"])
+	}
+	return Web3SignerConfig{
+		URL:     urlStr,
+		Address: addr,
+	}, nil
+}
+
+type Web3SignerSign struct {
```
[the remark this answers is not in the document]
Stefan-Ethernal: IMHO, the `Web3Signer` (or `Web3SignService` etc.) is still the best candidate (at least IMO). The `Web3Sign` or `Web3SignerSign` sound a bit awkward.

While it is true that the convention states that interfaces should be named with an `-er` suffix, it does not necessarily mean that structs cannot be named with exactly the same suffix (in [the golang guideliness](https://go.dev/doc/effective_go#names), there is nothing explicitly written for the struct naming).

For the reference: https://github.com/ethereum/go-ethereum/blob/756310fa43b8beab12b4fbe5fb663bc79285c8d2/core/types/transaction_signing.go#L568.

However, if you think it is still better to name it `Web3Sign` and `LocalSign`, feel free to name it that way. It is more important to me to name the similar/same concepts consistently. :slightly_smiling_face:

signer/local.go
Stefan-Ethernal: IMO renaming to `LocalSigner` would make more sense (semantically). But if you are against this suggestion, leave it like this.

signer/local.go
Stefan-Ethernal: IMO renaming it to `LocalSigner` would make more sense (semantically). But if you are against this suggestion, leave it like this.

signer/local.go
joanestebanr: The suffix `-er` suggest that is an interface, that is the reason that I choose `Sign`

signer/web3signer_client/client.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/0xPolygon/cdk-rpc/rpc"
+	"github.com/ethereum/go-ethereum/common"
+)
+
```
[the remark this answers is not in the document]
joanestebanr: It's a very clever solution to check the interface!
AFAIK,in that case,  from a point of view of golang idiomatic, make more sense to declare the interface in the place where is going to be used to reduce the coupling with the client (maybe the client could have more methods in the future that are going to be used for somebody else)

signer/web3signer_sign.go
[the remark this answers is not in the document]
joanestebanr: I think that I will keep the name   `Web3SignerSign`  instead of `Web3Sign`  because it's a reference to the underlaying  service used that is called [Web3Signer](https://docs.web3signer.consensys.io/).  WDYT?

PR #204, opened 2025-02-11
feat: send `GER` inclusion proofs to `aggchain prover`

goran-ethernal:
## Description

This PR adds the functionality of sending `GlobalExitRoot` inclusion proofs to the `aggchain prover` since it is needed to generate a correct `auth` proof.

To do this, when `aggsender` is in `AggchainProof` mode, we collect injected `GER`s in `L2` by directly calling the `L2` contract to collect `InsertGlobalExitRoot` events in a given block range.

Before calling the `aggchain prover` to generate the `auth` proof, `aggsender` will get the last finalized `l1 info tree root`. After that, it will get the all injected `GERs` from the `L2` contracts for the block range for which it is building the certificate, and it will generate the inclusion proofs for each injected `GER` based on the last finalized `l1 info tree root`.

Also, it will provide the prover with imported bridge exits we are adding to the certificate, but only data in imported bridge exits like:
- bridge exit
- token info
- global index.

After that it will call the `aggchain prover` so that it can generate the `auth` proof.

Also, important note is, we now will use the last finalized `l1 info tree root` in the `AggchainProof` mode to generate all the inclusion proofs for every claim (`imported bridge exit`), since this is a prerequesite, since the `aggchain prover` uses that root to generate the proof.

This PR adds a new field called `custom_chain_data` to the certfificate. It is a random byte array returned by the `aggchain prover`, and we need to send it to the `agglayer` in the certificate.

Fixes # (issue)


arnaubennassar (changes requested):
If I remember well, the `lastgersync` only stores the GERs that are "latest" in terms of index, but not necessarily all the injections...

Please review if the logic of this sycner does what you expect.
We may need to heavily change how the syncer works, if this is the case just go ahead and simplify the syncer to "just" index the GERs injections, and we will figure out later how we deal with the birdge service, as this syncer was meant to be used by it originally

goran-ethernal:
> If I remember well, the `lastgersync` only stores the GERs that are "latest" in terms of index, but not necessarily all the injections...
> 
> Please review if the logic of this sycner does what you expect. We may need to heavily change how the syncer works, if this is the case just go ahead and simplify the syncer to "just" index the GERs injections, and we will figure out later how we deal with the birdge service, as this syncer was meant to be used by it originally

@arnaubennassar I think we are good here. What the `lastGER` syncer does is it constantly gets what is the last index it has in its `db`, then gets what is the last root the `l1 info tree` has, and gets all the indexes it missed. After that it does sanity check to see if all of the `GERs` it got are injected on L2.

aggsender/l2_etherman.go
```diff
@@ -0,0 +1,82 @@
...
+
+// L2GERManager is an interface to interact with the GlobalExitRootManager contract
+type L2GERManager interface {
+	BridgeAddress(*bind.CallOpts) (common.Address, error)
+	FilterInsertGlobalExitRoot(opts *bind.FilterOpts, newGlobalExitRoot [][32]byte, newHashChainValue [][32]byte) (
+		*globalexitrootmanagerl2sovereignchain.Globalexitrootmanagerl2sovereignchainInsertGlobalExitRootIterator, error)
+}
+
+// L2Etherman is a component used to interact with L2 contracts
+type L2Etherman struct {
```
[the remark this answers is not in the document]
goran-ethernal: It makes sense. I will move it to a new `chaingerreader` package in the `aggoracle`.

config/default.go
```diff
@@ -228,5 +228,6 @@ MaxEpochPercentageAllowedToSendCertificate=80
 	[AggSender.MaxSubmitCertificateRate]
 		NumRequests = 20
 		Interval = "1h"
+GlobalExitRootL2="{{L2Config.GlobalExitRootAddr}}"
```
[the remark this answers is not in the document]
goran-ethernal: Done: https://github.com/agglayer/aggkit/pull/204/commits/0057047b1a1f298233964c0cce737398b6d1fc27

aggsender/l2_etherman.go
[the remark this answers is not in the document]
goran-ethernal: Done: https://github.com/agglayer/aggkit/pull/204/commits/0057047b1a1f298233964c0cce737398b6d1fc27

aggsender/flow_aggchain_prover.go
```diff
@@ -131,40 +155,74 @@ func (a *aggchainProverFlow) GetCertificateBuildParams(ctx context.Context) (*ty
...
+	proofs := make(map[common.Hash]treeTypes.Proof, len(injectedGERs))
+
+	for _, gerHash := range injectedGERs {
+		info, err := a.l1InfoTreeSyncer.GetInfoByGlobalExitRoot(gerHash)
+		if err != nil {
+			return nil, fmt.Errorf("aggchainProverFlow - error getting L1 Info tree leaf by global exit root %s: %w",
+				gerHash.String(), err)
+		}
+
+		proof, err := a.l1InfoTreeSyncer.GetL1InfoTreeMerkleProofFromIndexToRoot(ctx,
```
joanestebanr, line 179: I think that `GetL1InfoTreeMerkleProofFromIndexToRoot` doesn't prevent from doing a proof for a leaf not included on root. Maybe make sense to check it before

aggsender/flow_aggchain_prover.go
```diff
@@ -97,21 +108,33 @@ func (a *aggchainProverFlow) GetCertificateBuildParams(ctx context.Context) (*ty
...
+	// set the root from which to generate merkle proofs for each claim
+	// this is crucial since Aggchain Prover will use this root to generate the proofs as well
+	buildParams.L1InfoTreeRootFromWhichToProve = root
+
+	injectedGERsProofs, err := a.getInjectedGERsProofs(ctx, root, buildParams.FromBlock, buildParams.ToBlock)
+	if err != nil {
+		return nil, fmt.Errorf("aggchainProverFlow - error getting injected GERs proofs: %w", err)
+	}
+
+	importedBridgeExits, err := a.getImportedBridgeExitsForProver(buildParams.Claims)
```
joanestebanr, line 130: Claims are included in L1InfoTree root?

aggsender/flow_aggchain_prover.go
```diff
@@ -97,21 +108,33 @@ func (a *aggchainProverFlow) GetCertificateBuildParams(ctx context.Context) (*ty
...
+	if err != nil {
+		return nil, fmt.Errorf("aggchainProverFlow - error getting imported bridge exits for prover: %w", err)
+	}
+
 	aggchainProof, err := a.aggchainProofClient.GenerateAggchainProof(
```
joanestebanr, line 135: The field aggchainProof.LocalExitRoot is discarded... maybe we can take advantage of that and check it against the certificate

aggsender/flow_aggchain_prover.go
```diff
@@ -122,6 +145,7 @@ func (a *aggchainProverFlow) GetCertificateBuildParams(ctx context.Context) (*ty
 		buildParams.FromBlock, buildParams.ToBlock)
 
 	buildParams.AggchainProof = aggchainProof.Proof
+	buildParams.CustomChainData = aggchainProof.CustomChainData
 
 	buildParams, err = adjustBlockRange(buildParams, buildParams.ToBlock, aggchainProof.EndBlock)
```
joanestebanr, line 150: After adjusting blockRange maybe len(Bridges) == 0, and then we can't send this certificate

aggsender/flow_pp.go
```diff
@@ -146,7 +152,7 @@ func (f *baseFlow) buildCertificate(ctx context.Context,
 	}
 
 	bridgeExits := f.getBridgeExits(certParams.Bridges)
-	importedBridgeExits, err := f.getImportedBridgeExits(ctx, certParams.Claims)
+	importedBridgeExits, err := f.getImportedBridgeExits(ctx, certParams.Claims, certParams.L1InfoTreeRootFromWhichToProve)
```
joanestebanr, line 155: for a PP which is the value of certParams.L1InfoTreeRootFromWhichToProve?

joanestebanr (commented):
Some comments: 
- Make sense to store on DB the L1InfoTreeRoot choosen for proves.. may be it can be extracted from certificate, I'm not sure

aggsender/flow_aggchain_prover.go
goran-ethernal: Which is fine I guess. Maybe all those bridges are not part of the finalized l1 info root.

aggsender/flow_pp.go
goran-ethernal: It is nil. Which will make us calculate it in the pp flow based on imported bridge exits that go in the certificate. This is checked and dobe in the getImportedBridgeExits function.

aggsender/flow_aggchain_prover.go
goran-ethernal: So claim, or imported bridge exit, has a bridge exit on L1, so they are included in the l1 info tree.

aggsender/flow_aggchain_prover.go
goran-ethernal: We had a little discussion, and we are good to ask for the merkle proof here with the given root, because, on this [line](https://github.com/agglayer/aggkit/blob/e47f778502f2cb4e80c27f62e61c85b4e026c419/aggsender/flow_aggchain_prover.go#L125), we are asking for merkle proofs of injected GERs on L2, based on that L1 info root, before we call the `aggchain prover`, and before we generate merkle proofs for claims. So it will fail on this point if the `aggoracle` didn't inject all the `GERs` on L1, or if it has some bug.

aggsender/flow_aggchain_prover.go
goran-ethernal: Plus, to add to this, if it was an exit on L2, and it was claimed on L1, the `aggoracle` will inject that `GER` to L2, after it sees that L1 info tree has changed.

aggsender/flow_aggchain_prover.go
goran-ethernal: I modified this a bit, to ask after `adjustBlockRange` if we have any bridges. Fix:

https://github.com/agglayer/aggkit/pull/204/commits/024e1ac57642fc86d7ced9422bfd3abeaf7adee6

aggsender/flow_aggchain_prover.go
goran-ethernal: I am unsure which LER will be returned by the `aggchain prover`. I would leave it like this for now, until we have a clear picture of `api` response by the prover, since the structure of the `aggchain proof` will most definitely be changed.

Stefan-Ethernal (approved):

aggoracle/chaingerreader/evm_test.go
```diff
@@ -0,0 +1,93 @@
...
+		mockL2GERManager := mocks.NewL2GERManagerContract(t)
+		mockL2GERManager.On("FilterInsertGlobalExitRoot", &bind.FilterOpts{
+			Context: ctx,
+			Start:   1,
+			End:     &toBlock,
+		}, mock.Anything, mock.Anything).Return(nil, errors.New("failed to create iterator"))
+
+		l2Etherman := &EVMChainGERReader{l2GERManager: mockL2GERManager}
+
+		_, err := l2Etherman.GetInjectedGERsForRange(ctx, 1, toBlock)
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/204/commits/4a0c9f146d8cfbc02b031b211f2e42b2c1baaf18

aggoracle/chaingerreader/evm_test.go
```diff
@@ -0,0 +1,93 @@
...
+	invalidAddress := common.Address{}
+
+	t.Run("success", func(t *testing.T) {
+		t.Parallel()
+		mockL2GERManager := mocks.NewL2GERManagerContract(t)
+		mockL2GERManager.On("BridgeAddress", (*bind.CallOpts)(nil)).Return(validAddress, nil)
+
+		l2Etherman, err := newEVMChainGERReader(mockL2GERManager, validAddress)
+		assert.NoError(t, err)
+		assert.NotNil(t, l2Etherman)
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/204/commits/4a0c9f146d8cfbc02b031b211f2e42b2c1baaf18

aggoracle/chaingerreader/evm.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return fmt.Errorf("fail sanity check GlobalExitRootManagerL2(%s) Contract. Err: %w", contractAddr.String(), err)
+	}
+	log.Infof("sanity check GlobalExitRootManagerL2(%s) OK. bridgeAddress: %s", contractAddr.String(), bridgeAddr.String())
+	return nil
+}
+
+// GetInjectedGERsForRange returns the injected GlobalExitRoots for the given block range
+func (e *EVMChainGERReader) GetInjectedGERsForRange(ctx context.Context, fromBlock, toBlock uint64) ([]common.Hash, error) {
+	iter, err := e.l2GERManager.FilterInsertGlobalExitRoot(
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/204/commits/4a0c9f146d8cfbc02b031b211f2e42b2c1baaf18

joanestebanr:
> > Some comments:
> > 
> > * Make sense to store on DB the L1InfoTreeRoot choosen for proves.. may be it can be extracted from certificate, I'm not sure
> 
> I mean we could, but we would only use it for resending the `InError` certificate, and its basically just a call to the l1 info tree sync db. Maybe it is not worth it, since it is already stored in another db. WDYT?

In which DB do you mean? I think that for debug can be useful to known the choosen root.. WDYT?

goran-ethernal:
> > > Some comments:
> > > 
> > > * Make sense to store on DB the L1InfoTreeRoot choosen for proves.. may be it can be extracted from certificate, I'm not sure
> > 
> > 
> > I mean we could, but we would only use it for resending the `InError` certificate, and its basically just a call to the l1 info tree sync db. Maybe it is not worth it, since it is already stored in another db. WDYT?
> 
> In which DB do you mean? I think that for debug can be useful to known the choosen root.. WDYT?

I agree, I will add it.

aggsender/flow_aggchain_prover.go
joanestebanr: My point is that you choose in a way the L1InfoRoot and in another way the Claims / Bridges. Maybe make sense to check that are in the same page. WDYT?

aggsender/flow_pp.go
joanestebanr: Maybe this code make sense to be in FlowPP class, and the base assume that you must set this param L1InfoTreeRootFromWhichToProve... maybe is going to be more clear

aggsender/flow_aggchain_prover.go
joanestebanr: I think that stil exists the case, let's duscus later

aggsender/flow_aggchain_prover.go
goran-ethernal: You are right. We need to add that check.

aggsender/flow_pp.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/204/commits/04d50936683bd1030213cfc9f8df4d8baabc8885

aggsender/flow_aggchain_prover.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/204/commits/c99394cc408532226d9b203b80527c0dbd85cc47

joanestebanr (approved):

Issue #186, opened 2025-02-06
Bridge API


temaniarpit27:
- [x] API
- [x] E2E testcase
- [x] Doc

PR #205, opened 2025-02-12
feat: bridge api

temaniarpit27:
#186


rachit77 (approved):

bridgesync/migrations/bridgesync0003.sql
Stefan-Ethernal, line 1: Let's move this to the `bridgesync0002.sql` migration file, as we will ship the entire bridge API simultaneously (at least the read API).

bridgesync/processor.go
```diff
@@ -211,6 +215,45 @@ func (p *processor) GetClaims(
...
+	if err != nil {
+		return nil, 0, err
+	}
+	if depositCount > 0 {
+		whereClause = fmt.Sprintf("WHERE deposit_count = %d", depositCount)
+		pageNumber = 1
+		pageSize = 1
+	}
+	offset := (pageNumber - 1) * pageSize
+	rows, err := p.queryPaged(tx, offset, pageSize, "bridge", orderBy, order, whereClause)
```
Stefan-Ethernal, line 243: 1. Do we need to call `rows.Close` in deferred execution?

2. Use the const for the bridge table name
```suggestion
	rows, err := p.queryPaged(tx, offset, pageSize, bridgeTableName, orderBy, order, whereClause)
```

bridgesync/processor.go
```diff
@@ -211,6 +215,45 @@ func (p *processor) GetClaims(
...
+	count, err := p.GetTotalNumberOfRecords("bridge")
+	if err != nil {
+		return nil, 0, err
+	}
+	if depositCount > 0 {
+		whereClause = fmt.Sprintf("WHERE deposit_count = %d", depositCount)
+		pageNumber = 1
+		pageSize = 1
+	}
+	offset := (pageNumber - 1) * pageSize
```
Stefan-Ethernal, line 242: What would happen if the `offset` value is greater than the `count`? Consider adding necessary validations.

bridgesync/processor.go
```diff
@@ -211,6 +215,45 @@ func (p *processor) GetClaims(
...
+	}
+	defer func() {
+		if err := tx.Rollback(); err != nil {
+			log.Warnf("error rolling back tx: %v", err)
+		}
+	}()
+	orderBy := "deposit_count"
+	order := "DESC"
+	whereClause := ""
+	count, err := p.GetTotalNumberOfRecords("bridge")
```
Stefan-Ethernal, line 233:
```suggestion
	count, err := p.GetTotalNumberOfRecords(bridgeTableName)
```

rpc/bridge.go
```diff
@@ -209,6 +209,59 @@ func (b *BridgeEndpoints) InjectedInfoAfterIndex(networkID uint32, l1InfoTreeInd
...
+// BridgesResult contains the bridges and the total count of bridges
+type BridgesResult struct {
+	Bridges []*bridgesync.Bridge `json:"bridges"`
+	Count   int                  `json:"count"`
+}
+
+func (b *BridgeEndpoints) GetBridges(
+	networkID uint32,
+	pageNumber, pageSize *uint32,
+	depositCount uint64,
```
Stefan-Ethernal, line 221: Should it be nullable, because AFAIK it should be an optional parameter

```suggestion
	depositCount *uint64,
```

rpc/bridge.go
```diff
@@ -209,6 +209,59 @@ func (b *BridgeEndpoints) InjectedInfoAfterIndex(networkID uint32, l1InfoTreeInd
...
+	if err != nil {
+		return nil, rpc.NewRPCError(rpc.InvalidRequestErrorCode, err.Error())
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), b.readTimeout)
+	defer cancel()
+
+	c, merr := b.meter.Int64Counter("getBridges")
+	if merr != nil {
+		b.logger.Warnf("failed to create get_token_mappings counter: %s", merr)
```
Stefan-Ethernal, line 233:
```suggestion
	c, merr := b.meter.Int64Counter("get_bridges")
	if merr != nil {
		b.logger.Warnf("failed to create get_bridges counter: %s", merr)
```

rpc/bridge.go
```diff
@@ -209,6 +209,59 @@ func (b *BridgeEndpoints) InjectedInfoAfterIndex(networkID uint32, l1InfoTreeInd
...
+	var (
+		bridges []*bridgesync.Bridge
+		count   int
+	)
+
+	switch {
+	case networkID == 0:
+		bridges, count, err = b.bridgeL1.GetBridgesPaged(ctx, pageNumberU32, pageSizeU32, depositCount)
+		if err != nil {
+			return nil, rpc.NewRPCError(rpc.DefaultErrorCode, fmt.Sprintf("failed to get deposit, error: %s", err))
```
Stefan-Ethernal, line 246:
```suggestion
			return nil, rpc.NewRPCError(rpc.DefaultErrorCode, fmt.Sprintf("failed to get bridges, error: %s", err))
```

rpc/bridge.go
```diff
@@ -209,6 +209,59 @@ func (b *BridgeEndpoints) InjectedInfoAfterIndex(networkID uint32, l1InfoTreeInd
...
+		}
+	case networkID == b.networkID:
+		bridges, count, err = b.bridgeL2.GetBridgesPaged(ctx, pageNumberU32, pageSizeU32, depositCount)
+		if err != nil {
+			return nil, rpc.NewRPCError(rpc.DefaultErrorCode, fmt.Sprintf("failed to get deposit, error: %s", err))
```
Stefan-Ethernal, line 251:
```suggestion
			return nil, rpc.NewRPCError(rpc.DefaultErrorCode, fmt.Sprintf("failed to get bridges, error: %s", err))
```

bridgesync/migrations/migrations.go
Stefan-Ethernal, line 1: Consider adding a unit test to `migrations_test`, that would test this migration as well.

bridgesync/processor.go
```diff
@@ -52,8 +52,18 @@ type Bridge struct {
...
 	DepositCount       uint32         `meddler:"deposit_count"`
+	BlockTimestamp     uint64         `meddler:"block_timestamp"`
+	TxHash             common.Hash    `meddler:"tx_hash,hash"`
+	FromAddress        common.Address `meddler:"from_address,address"`
 }
 
+// BridgeResponse is the representation of a bridge event with additional fields
+type BridgeResponse struct {
+	Bridge
+	BridgeHash common.Hash `meddler:"bridge_hash,hash"`
```
Stefan-Ethernal, line 63: If we don't want to persist the `BridgeHash` to the db, we can remove the `meddler` tag. Otherwise, let's rename the field to `Hash` and move it to the `Bridge` struct.

```suggestion
	BridgeHash common.Hash
```

bridgesync/downloader.go
```diff
@@ -74,6 +74,9 @@ func buildAppender(client EthClienter, bridge common.Address, syncFullClaims boo
 			Amount:             bridge.Amount,
 			Metadata:           bridge.Metadata,
 			DepositCount:       bridge.DepositCount,
+			BlockTimestamp:     b.Timestamp,
+			TxHash:             l.TxHash,
+			FromAddress:        l.Address,
```
Stefan-Ethernal, line 79: Rename it per Notion doc (or the `ContractAddress`), wdyt?
```suggestion
			EventInitiatorAddress:        l.Address,
```

Stefan-Ethernal (approved):

bridgesync/downloader.go
temaniarpit27: i would like to go with fromAddress. it can be used by frontend in whatever way they want then. wdyt?

Stefan-Ethernal:
@temaniarpit27 I have just remembered: please update the openrpc json file (https://github.com/agglayer/aggkit/blob/docs/bridge-service/rpc/openrpc.json) with the `bridge_getBridges` endpoint.

PR #225, opened 2025-02-20
feat: ger leaf

temaniarpit27:


vcastellm (approved):
LGTM

aggsender/proto/aggchain_proof_generation.proto
```diff
@@ -17,13 +21,13 @@ message GenerateAggchainProofRequest {
   // The max end block for which the aggchain proof is requested.
   uint64 max_end_block = 2;
   // L1 Info tree root. (hash)
-  bytes l1_info_tree_root_hash = 3;
+  FixedBytes32 l1_info_tree_root_hash = 3;
```
joanestebanr, line 24: What is the reason behind this change? FixedBytes32 = bytes... it looks almost the same. FixedBytes32 is a  'Hash'?

temaniarpit27: we want to keep it consistent with agglayer proto and provers proto
so moved according to what it is used everywhere.
going forward we will try to also directly use proto files defined at other places instead of copy pasting it here

joanestebanr (approved):

PR #236, opened 2025-02-24
test: e2e testcase

temaniarpit27:


rpc/bridge.go
```diff
@@ -268,6 +268,7 @@ func (b *BridgeEndpoints) GetBridges(
 			fmt.Sprintf("this client does not support network %d", networkID),
 		)
 	}
+	fmt.Println("------------- bridges", bridges)
```
Stefan-Ethernal, line 271: Do we need this? In case we do, I'm suggesting adding a debug log. WDYT?
```suggestion
```

test/bats/pp/bridge-e2e.bats
```diff
@@ -64,6 +64,12 @@ setup() {
     assert_success
     local bridge_tx_hash=$output
 
+    echo "------- bridge_getBridges API testcase"
+    local aggkit_node_url=$(kurtosis port print $enclave cdk-node-001 rpc)
+    run get_bridges $aggkit_node_url $l1_rpc_network_id
```
Stefan-Ethernal, line 69: Can we return bridges as a return value from the `get_bridges` function and assert that some values are expected: `origin_network`, `origin_address`, `destination_network`, `destination_address`, `amount`, etc.
WDYT?

test/bats/helpers/lxly-bridge.bash
```diff
@@ -89,6 +89,55 @@ function bridge_asset() {
...
+
+    echo "------ bridges_result ------"
+    echo "$bridges_result"
+    echo "------ bridges_result ------"
+
+    count=$(jq -r '.count' <<< "$bridges_result")
+    if [ "$count" -ne 1 ]; then
+        echo "Assertion failed. 'count' should be 1. Actual: $count"
+        exit 1
+    fi
```
Stefan-Ethernal, line 106: I'd say, let's search for a specific `bridge_tx_hash` and return it instead. If it is not found, then return an error. 

The reason I'm saying this is that then we could reuse this function in other tests as well (imagine running multiple tests against the same cluster, which is the case on the CI, and therefore there could be more than one bridges). WDYT?

test/bats/helpers/lxly-bridge.bash
```diff
@@ -89,6 +89,55 @@ function bridge_asset() {
...
+            echo "Assertion failed. Missing or null '$field' in the first bridge object."
+            exit 1
+        fi
+    done
+
+    origin_network=$(jq -r '.bridges[0].origin_network' <<< "$bridges_result")
+    if [ "$origin_network" -ne 0 ]; then
+        echo "Assertion failed. 'origin_network' should be 0. Actual: $origin_network"
+        exit 1
+    fi
```
Stefan-Ethernal, line 138: This should be also asserted on the outside (in the particular test). This helper function, IMO should just return a `bridge` response, based on a provided `tx_hash` (or error in case it is not found).

Stefan-Ethernal (approved):
LGTM.

Maybe we could also try to retrieve the bridge using the `deposit_count` value?

PR #242, opened 2025-02-25
Feat/reorg detector logs 121

joanestebanr:
## Description

Please include a summary of the change and which issue is fixed. Also, include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)


reorgdetector/migrations/reorgdetector0002.sql
```diff
@@ -0,0 +1,14 @@
+-- +migrate Down
+DROP TABLE IF EXISTS reorg_event;
+
+-- +migrate Up
+CREATE TABLE reorg_event (
```
goran-ethernal, line 5: Do we need some primary key for the table?

reorgdetector/reorgdetector.go
```diff
@@ -223,17 +223,35 @@ func (rd *ReorgDetector) detectReorgInTrackedList(ctx context.Context) error {
...
 				}
-
+				event := ReorgEvent{
+					DetectedAt:   startTime,
+					FromBlock:    hdr.Num,
+					ToBlock:      headers[len(headers)-1].Num,
+					SubscriberID: id,
+					CurrentHash:  currentHeader.Hash(),
+					TrackedHash:  hdr.Hash,
+					ExtraData: struct {
```
goran-ethernal, line 233: What is the usage of `ExtraData` field? I don't see its value if it is only used on this place. I would say, just make these two regular fields in the `ReorgEvent`.

reorgdetector/reorgdetector_db.go
```diff
@@ -78,3 +82,42 @@ func (rd *ReorgDetector) removeTrackedBlockRange(id string, fromBlock, toBlock u
...
+	DetectedAt   time.Time
+	FromBlock    uint64
+	ToBlock      uint64
+	SubscriberID string
+	TrackedHash  common.Hash
+	CurrentHash  common.Hash
+	ExtraData    interface{}
+}
+
+type eventReorgRow struct {
```
goran-ethernal, line 96: I think the point of meddler is to not have struct conversion before saving it to the `db`, because he is the one that does that in the background. So, I would say we don't need this struct.

reorgdetector/migrations/reorgdetector0002.sql
joanestebanr: You're right, I would set (detected_at, subscriber_id, from_block, to_block)

reorgdetector/reorgdetector.go
joanestebanr: The idea is to be able to grow, modify this extraData without modifying the database struct. Right now I set network, and blocktimestamp, but it's not clear what extra context data we will add. This fields doesn't make sense to be real fields on DB because doesn't belong to the key data. Also, we avoid migration process if we decide to modify it. WDYT?

reorgdetector/reorgdetector_db.go
joanestebanr: I think that is a bad practice, because you're coupling runtime data with DB structs:
-  Force the type to the same as DB, like using uint64 instead of time.Time, 
-  Fields like `version` that only exists on DB must exists on `ReorgEvent` struct 

Anyway. I change it to remove the DB struct and the event directly can be use in meddler

reorgdetector/reorgdetector_db_test.go
```diff
@@ -0,0 +1,56 @@
...
+		SubscriberID: "test",
+		TrackedHash:  common.Hash{},
+		CurrentHash:  common.Hash{},
+		Version:      "1.0",
+		ExtraData:    "extra",
+	}
+
+	err = reorgDetector.insertReorgEvent(event)
+	require.NoError(t, err)
+	row := reorgDetector.db.QueryRow("SELECT * FROM reorg_event WHERE subscriber_id = $1", "test")
```
goran-ethernal, line 37: You could replace these query row and scan with:

```go
var eventFromDB ReorgEvent

err := meddler.QueryRow(db, &eventFromDB,
		"SELECT * FROM reorg_event WHERE subscriber_id = $1;", "test")

require.NoError(t, err)
require.Equal(t, event, eventFromDB)
```

reorgdetector/reorgdetector_db_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"time"
+
+	aggkittypes "github.com/agglayer/aggkit/config/types"
+	"github.com/agglayer/aggkit/etherman"
+	common "github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/require"
+)
+
+func TestInsertReorgEvent(t *testing.T) {
+	// Create test DB dir
```
goran-ethernal, line 14: Should we uncomment this, and use this one as `testDir`?

reorgdetector/reorgdetector.go
Stefan-Ethernal: I would agree with @goran-ethernal here. IMO either have data present in columns, or capture them in a separate table (if we want to have better modularity) and have the foreign key to it, have some kind of IS-A hierarchy (if it is supported by the SQL lite) etc.

This way, we are restricted and we cannot manipulate with those data on a database level (but rather on an application level).

reorgdetector/reorgdetector_db_test.go
joanestebanr: Thanks! good catch!!

goran-ethernal (approved):

Stefan-Ethernal (approved):

PR #243, opened 2025-02-25
test: e2e`bridge_getClaims`

rachit77:
## Description

 Added bats e2e for`bridge_getClaims` endpoint

### DIFF
--- a/test/bats/helpers/lxly-bridge.bash
+++ b/test/bats/helpers/lxly-bridge.bash
@@ -90,11 +90,10 @@ function bridge_asset() {
 }
 
 function get_bridge() {
-    local aggkit_node_url=$1
-    local network_id=$2
-    local expected_tx_hash=$3
-    local max_attempts=$4
-    local poll_frequency=$5
+    local network_id="$1"
+    local expected_tx_hash="$2"
+    local max_attempts="$3"
+    local poll_frequency="$4"
 
     local attempt=0
 
@@ -320,10 +319,9 @@ function check_claim_revert_code() {
 }
 
 function wait_for_expected_token() {
-    local aggkit_node_url="$1"
-    local expected_origin_token="$2"
-    local max_attempts="$3"
-    local poll_frequency="$4"
+    local expected_origin_token="$1"
+    local max_attempts="$2"
+    local poll_frequency="$3"
 
     local attempt=0
     local token_mappings_result
@@ -359,17 +357,16 @@ function wait_for_expected_token() {
 }
 
 function get_claim() {
-    local aggkit_node_url=$1
-    local l1_rpc_network_id=$2
-    local expected_global_index=$3
-    local max_attempts=$4
-    local poll_frequency=$5
+    local network_id="$1"
+    local expected_global_index="$2"
+    local max_attempts="$3"
+    local poll_frequency="$4"
     local attempt=0
 
     while true; do
         ((attempt++))
         log "🔍 Attempt $attempt: fetching claims from the RPC ("$aggkit_node_url")..."
-        claims_result=$(cast rpc --rpc-url "$aggkit_node_url" "bridge_getClaims" "$l1_rpc_network_id")
+        claims_result=$(cast rpc --rpc-url "$aggkit_node_url" "bridge_getClaims" "$network_id")
         log "------ claims_result ------"
         log "$claims_result"
 
--- a/test/bats/helpers/lxly-bridge.bash
+++ b/test/bats/helpers/lxly-bridge.bash
@@ -89,48 +89,6 @@ function bridge_asset() {
     fi
 }
 
-function get_bridge() {
-    local network_id="$1"
-    local expected_tx_hash="$2"
-    local max_attempts="$3"
-    local poll_frequency="$4"
-
-    local attempt=0
-
-    while true; do
-        ((attempt++))
-        log "Attempt $attempt: fetching bridges from the RPC..."
-
-        # Fetch bridges from the RPC
-        bridges_result=$(cast rpc --rpc-url "$aggkit_node_url" "bridge_getBridges" "$network_id")
-
-        log "------ bridges_result ------"
-        log "$bridges_result"
-        log "------ bridges_result ------"
-
-        # Extract the elements of the 'bridges' array one by one
-        for row in $(echo "$bridges_result" | jq -c '.bridges[]'); do
-            # Parse out the tx_hash from each element
-            tx_hash=$(echo "$row" | jq -r '.tx_hash')
-
-            if [[ "$tx_hash" == "$expected_tx_hash" ]]; then
-                log "Found expected bridge with tx hash: $tx_hash"
-                echo "$row"
-                return 0
-            fi
-        done
-
-        # Fail test if max attempts are reached
-        if [[ "$attempt" -ge "$max_attempts" ]]; then
-            echo "Error: Reached max attempts ($max_attempts) without finding expected bridge with tx hash." >&2
-            return 1
-        fi
-
-        # Sleep before the next attempt
-        sleep "$poll_frequency"
-    done
-}
-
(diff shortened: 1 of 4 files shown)


temaniarpit27 (approved):

test/bats/helpers/lxly-bridge.bash
```diff
@@ -313,3 +314,47 @@ function wait_for_expected_token() {
...
+            exit 1
+        fi
+    done
+
+    origin_network=$(jq -r '.claims[0].origin_network' <<< "$claims_result")
+    if [ "$origin_network" -ne 0 ]; then
+        echo "Assertion failed. 'origin_network' should be 0. Actual: $origin_network"
+        exit 1
+    fi
+}
```
Stefan-Ethernal, line 360: Similarly to what I have written to @temaniarpit27 [here](https://github.com/agglayer/aggkit/pull/236#discussion_r1973335179), IMO this should return a certain claim based on provided either `global_index`, `tx_hash` and return it back to the caller. Then we can reuse this function in many tests and make certain assertions on a given test, however we want.

test/bats/helpers/lxly-bridge.bash
```diff
@@ -356,3 +357,60 @@ function wait_for_expected_token() {
         sleep "$poll_frequency"
     done
 }
+
+function get_claims() {
```
Stefan-Ethernal, line 361:
```suggestion
function get_claim() {
```

test/bats/helpers/lxly-bridge.bash
```diff
@@ -356,3 +357,60 @@ function wait_for_expected_token() {
...
+    local aggkit_node_url=$1
+    local l1_rpc_network_id=$2
+    local expected_global_index=$3
+    local max_attempts=$4
+    local poll_frequency=$5
+    local attempt=0
+
+    while true; do
+        ((attempt++))
+        echo "🔍 Attempt $attempt: fetching claims from the RPC..." >&3
```
Stefan-Ethernal, line 371: FYI it is better to use the `log` function, because it includes timestamp as well (also in that case `>&3` redirection should be removed)

```suggestion
        log "🔍 Attempt $attempt: fetching claims from the RPC ("$aggkit_node_url")..."
```

test/bats/pp/bridge-e2e.bats
```diff
@@ -151,6 +152,17 @@ setup() {
...
+    # Validate the bridge_getClaims API
+    echo "------- bridge_getClaims API testcase --------"
+    run get_claims "$aggkit_node_url" "$l1_rpc_network_id" "$global_index" 10 15
+    assert_success
+
+    local origin_network="$(echo "$output" | jq -r '.origin_network')"
+    local destination_network="$(echo "$output" | jq -r '.destination_network')"
+    assert_equal "$l1_rpc_network_id" "$origin_network"
+    assert_equal "$l2_rpc_network_id" "$destination_network"
+    echo "🚀🚀 bridge_getClaims API testcase passed" >&3
```
Stefan-Ethernal, line 164: Just thinking out loud and want to discuss with you:
should we for now both `bridge_getBridges` and `bridge_getClaims` assertions make in a single e2e test (for example the `Native gas token deposit to WETH` one).

Eventually, probably we can make some helper functions and assert bridges and claims in each test (if I'm not mistaken, we are going to do this anyway once we substitute the `zkevm-bridge-service` with the aggkit one, so no rush with this).

test/bats/helpers/lxly-bridge.bash
```diff
@@ -356,3 +357,60 @@ function wait_for_expected_token() {
...
+        claims_result=$(cast rpc --rpc-url "$aggkit_node_url" "bridge_getClaims" "$l1_rpc_network_id")
+        echo "------ claims_result ------" >&3
+        echo "$claims_result" >&3
+
+        for row in $(echo "$claims_result" | jq -c '.claims[]'); do
+            global_index=$(jq -r '.global_index' <<< "$row")
+
+            if [[ "$global_index" == "$expected_global_index" ]]; then
+                echo "🎉 Success: Expected global_index '$expected_global_index' found. Exiting loop." >&3
```
Stefan-Ethernal, line 380:
```suggestion
                log "🎉 Success: Expected global_index '$expected_global_index' found. Exiting loop."
```

test/bats/helpers/lxly-bridge.bash
```diff
@@ -356,3 +357,60 @@ function wait_for_expected_token() {
...
+    local max_attempts=$4
+    local poll_frequency=$5
+    local attempt=0
+
+    while true; do
+        ((attempt++))
+        echo "🔍 Attempt $attempt: fetching claims from the RPC..." >&3
+        claims_result=$(cast rpc --rpc-url "$aggkit_node_url" "bridge_getClaims" "$l1_rpc_network_id")
+        echo "------ claims_result ------" >&3
+        echo "$claims_result" >&3
```
Stefan-Ethernal, line 374:
```suggestion
        log "------ claims_result ------"
        log "$claims_result"
```

test/bats/helpers/lxly-bridge.bash
```diff
@@ -356,3 +357,60 @@ function wait_for_expected_token() {
...
+                done
+                
+                echo "$row"
+                return 0
+            fi
+        done
+
+        # Fail test if max attempts are reached
+        if [[ "$attempt" -ge "$max_attempts" ]]; then
+            echo "❌ Error: Reached max attempts ($max_attempts) without finding expected claim with global index." >&2
```
Stefan-Ethernal, line 409: Log the expected global index, wdyt?
```suggestion
            echo "❌ Error: Reached max attempts ($max_attempts) without finding expected claim with global index ($expected_global_index)." >&2
```

Stefan-Ethernal (approved):

test/bats/pp/bridge-e2e.bats
```diff
@@ -151,6 +152,17 @@ setup() {
     run claim_tx_hash "$timeout" "$bridge_tx_hash" "$destination_addr" "$l2_rpc_url" "$bridge_api_url"
     assert_success
 
+    # Validate the bridge_getClaims API
+    echo "------- bridge_getClaims API testcase --------"
+    run get_claims "$aggkit_node_url" "$l1_rpc_network_id" "$global_index" 10 15
```
Stefan-Ethernal, line 157: Also bear in mind that `global_index` is not set
```suggestion
    run get_claims "$aggkit_node_url" "$l2_rpc_network_id" "$global_index" 10 15
```

rachit77: isn't `global_index` being set in `claim_tx_hash` [here] ??(https://github.com/agglayer/aggkit/blob/f9014dee8cce00694fa459eedf3f82f7cf2f7633/test/bats/helpers/lxly-bridge.bash#L189-L190)

Stefan-Ethernal: Yes it is, my bad. Although we should query the `$l2_rpc_network_id`. I have pushed the fix https://github.com/agglayer/aggkit/pull/243/commits/7a28ea904879a26a34a39afbd3054dd72a154964 and let's see.

test/bats/pp/bridge-e2e.bats
rachit77: I will move `bridge_getClaims` assertions to `Native gas token deposit to WETH` where we already have assertions for `bridges`

test/bats/helpers/lxly-bridge.bash
```diff
@@ -356,3 +357,59 @@ function wait_for_expected_token() {
         sleep "$poll_frequency"
     done
 }
+
+function get_claim() {
+    local l1_rpc_network_id="$1"
```
Stefan-Ethernal, line 362: Rename this, as it is not going to be an L1 network always, but it also can be an L2 one.
```suggestion
    local network_id="$1"
```

PR #244, opened 2025-02-25
fix: `bridge_getClaims` endpoint results

### LINKED ISSUE #227: `bridge_getClaims` endpoint returns incorrect of order of `GlobalIndex`
The `bridge_getClaims` endpoint does not return claims in the correct descending order of `GlobalIndex`.

SQLite only supports integers up to 8 bytes, while GlobalIndex is a big.Int.
Currently, GlobalIndex is stored as TEXT in the claim db, which leads to incorrect sorting for large numbers.

Expected Behavior:
Claims should be sorted numerically in descending order of GlobalIndex.

rachit77:
Fixes #227

### DIFF
--- a/bridgesync/processor.go
+++ b/bridgesync/processor.go
@@ -319,7 +319,7 @@ func (p *processor) GetClaimsPaged(
 		return nil, 0, db.ErrNotFound
 	}
 
-	orderBy := "global_index + 0"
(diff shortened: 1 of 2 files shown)


temaniarpit27 (approved):

Stefan-Ethernal (approved):
:clap: :100:

PR #254, opened 2025-02-28
feat: check runtime vs database db

### LINKED ISSUE #165: aggsender: store in database relevant execution data to detect a missmatch between running and database
The idea it's to detect this situation soon as possible: 
- I run aggsender with contract A that is networkID=5
- I stop it and change the contract to B that is networkID=15
The databases (aggsender and L2Brigdesyncer) contains invalid data, so must emit a fatal

Add a run flag to override this check

joanestebanr:
## Description

- Each database have a new key/value table  (`key_value`) 
- The system, at startup, checks that each database is compatible with this run. To achieve this:
   -  It checks previous data from key/value table, if it's empty it stores the revelant data to database (in case of syncer is chainID plus the addresses, in case of aggsender is networkID)
   - If the database data and runtime are compatible (is checked with  function `IsCompatible(storage T) error`  can run, if  not it's a fail (you can override that setting config variable `RequireStorageContentCompatibility` to false)
- The evmdriver for syncer check that the runtime environment match what is stored on DB: 
- Aggsender use the networkID

## Config changes
- There are a new param `RequireStorageContentCompatibility` (by default is `true`) as global var that is injected in:  `BridgeL1Sync`,  `BridgeL2Sync`, `LastGERSync`, `AggSender`:
```

[BridgeL1Sync]

[BridgeL2Sync]

[LastGERSync]

[AggSender]
```

Fixes #165

## Review comments:
- The interface `Logger` have been moved from `aggsender/types`  to  `common` packages because it's required by multiples modules
- The integration of this feature in syncer is: 
- The compatibility data is `sync.RuntimeData`: 
```
type RuntimeData struct {
	ChainID   uint64
	Addresses []common.Address
}
```
- `sync.EVMDriver` have a new fields `compatibilityChecker compatibility.CompatibilityChecker` that implements the logic of this feature
- The current (runtime) data is provided by `sync.downloader` interface using  `GetRuntimeData` 
- The stored data is provided by `sync.processorInterface`  implementing interface `compatibility.CompatibilityDataStorager[RuntimeData]`: There  are a helper objects to allow provide this funcionality directly:  that is `compatibility.NewKeyValueToCompatibilityStorage[T]`

The objects in compatibility module: 

```mermaid
---
title: CompatibilityCheck class diagram
---
classDiagram
   
    CompatibilityCheck <|-- RuntimeDataGetter
    CompatibilityCheck <|-- CompatibilityDataStorager
    
    class CompatibilityCheck{
        +Check(ctx context.Context, tx db.Querier)
    }
    class CompatibilityDataStorager{
        + GetCompatibilityData(ctx context.Context, tx db.Querier) (bool, T, error)
        + SetCompatibilityData(ctx context.Context, tx db.Querier, data T) error
    }

    class RuntimeDataGetter{
        + func(ctx context.Context) (T, error)
    }
```
To implement ` CompatibilityDataStorager` you can use: 

## PR Checklist:
 - [X] I have performed a self-review of my code
 - [X] I have commented my code, particularly in hard-to-understand areas
- [X]  I have made corresponding changes to the documentation
- [X]  I have added or updated tests that comprehensively prove my change is effective or that my feature works


bridgesync/bridgesync.go
```diff
@@ -49,6 +49,7 @@ func NewL1(
 	maxRetryAttemptsAfterError int,
 	originNetwork uint32,
 	syncFullClaims bool,
+	requireStorageContentCompatibility bool,
```
vcastellm, line 52: Why make this optional?

bridgesync/config.go
```diff
@@ -24,4 +24,7 @@ type Config struct {
 	MaxRetryAttemptsAfterError int `mapstructure:"MaxRetryAttemptsAfterError"`
 	// WaitForNewBlocksPeriod time that will be waited when the synchronizer has reached the latest block
 	WaitForNewBlocksPeriod types.Duration `mapstructure:"WaitForNewBlocksPeriod"`
+	// RequireStorageContentCompatibility is true it's mandatory that data stored in the database
+	// is compatible with the running environment
+	RequireStorageContentCompatibility bool `mapstructure:"RequireStorageContentCompatibility"`
```
vcastellm, line 29: Should this be optional and configurable?

sync/common.go
```diff
@@ -6,14 +6,21 @@ import (
 	"time"
 )
 
+var LogFatalf = log.Fatalf
```
vcastellm, line 9: Why this aliasing?

joanestebanr: It's for be able to check it in unittest. log.Fatalf procuce a os.Exit() that is not easy to unittest. You can check the usage of it in file `evmdriver_test.go`

bridgesync/config.go
joanestebanr: What do you mean? this is set by default to `true`. It's not necessary to set it specifically

sync/evmdriver.go
```diff
@@ -78,7 +124,16 @@ reset:
 		attempts           int
 		err                error
 	)
-
+	for {
+		err = d.compatibilityChecker.Check(ctx, nil)
```
goran-ethernal, line 128: Why do we check it here for the syncers? Shouldn't we check this on the start? I think every syncer has a `Start` function. Plus this will keep checking until the retry handler says it is enough, even though it will fail every time if its not compatible.

rachit77 (approved):
LGTM

bridgesync/config.go
vcastellm: I mean, do we want anyone to disable it?

sync/evmdownloader.go
```diff
@@ -244,6 +260,14 @@ func NewEVMDownloaderImplementation(
 	}
 }
 
+func (d *EVMDownloaderImplementation) GetChainID(ctx context.Context) (uint64, error) {
+	chainID, err := d.ethClient.ChainID(ctx)
+	if err != nil {
+		return 0, err
+	}
+	return chainID.Uint64(), nil
+}
```
vcastellm, line 269: IMO we can directly return here without any treatment, the err can be popped up to the called and the data type too:
```suggestion
func (d *EVMDownloaderImplementation) GetChainID(ctx context.Context) (*big.Int, error) {
	return d.ethClient.ChainID(ctx)
}
```

sync/evmdownloader.go
```diff
@@ -104,6 +108,18 @@ func (d *EVMDownloader) setStopDownloaderOnIterationN(iteration int) {
...
+// GetRuntimeData returns the runtime data: chainID + addresses to query
+func (d *EVMDownloader) GetRuntimeData(ctx context.Context) (RuntimeData, error) {
+	chainID, err := d.GetChainID(ctx)
+	if err != nil {
+		return RuntimeData{}, err
+	}
+	return RuntimeData{
+		ChainID:   chainID,
+		Addresses: d.adressessToQuery,
+	}, nil
```
vcastellm, line 120: Related to the other comment:
```suggestion
	chainID, err := d.ChainID(ctx)
	if err != nil {
		return RuntimeData{}, err
	}
	return RuntimeData{
		ChainID:   chainID.Uint64,
		Addresses: d.adressessToQuery,
	}, nil
```

sync/evmdownloader.go
```diff
@@ -104,6 +108,18 @@ func (d *EVMDownloader) setStopDownloaderOnIterationN(iteration int) {
 	d.stopDownloaderOnIterationN = iteration
 }
 
+// GetRuntimeData returns the runtime data: chainID + addresses to query
+func (d *EVMDownloader) GetRuntimeData(ctx context.Context) (RuntimeData, error) {
```
vcastellm, line 112:
```suggestion
func (d *EVMDownloader) RuntimeData(ctx context.Context) (RuntimeData, error) {
```

bridgesync/config.go
joanestebanr: In some scenarios you may want to skip this check because you kown that the data is compatible. So I think that is good to have a way to skip if it's required.

sync/evmdriver.go
joanestebanr: The `syncer.Start` it's 1 line that calls to `driver.Sync(ctx)` (the code that you mention). Some examples,[ brdigesync](https://github.com/agglayer/aggkit/blob/336ffecdbacc05953fc998aa8364d726a92c685e/bridgesync/bridgesync.go#L210), [l1infotreesync](https://github.com/agglayer/aggkit/blob/336ffecdbacc05953fc998aa8364d726a92c685e/l1infotreesync/l1infotreesync.go#L107)
So, eventually, this is checked on syncer.Start function but, to avoid duplications, it's moved the code to driver

goran-ethernal: Then, I would remove this `for` loop and just check it once since it will return an error every time if the `db` is not compatible.

sync/evmdownloader.go
joanestebanr: This is to use `uint64`  that is a more friendly type than `*big.Int`

sync/evmdriver.go
joanestebanr: I'm not sure about that. I have copied the pattern of the rest of checks (line 137 calls `GetLastProcessedBlock` and have the same `for`). The check can involve calls to external services so, make sense, to retry if fails. 

 Inside the `for` there are a call to `d.rh.Handle("Sync", attempts)` if fails that is going to stop process depending of the configuration of retries.

WDYT?

aggsender/aggsender.go
```diff
@@ -131,9 +142,20 @@ func (a *AggSender) Start(ctx context.Context) {
...
 	a.checkInitialStatus(ctx)
 	a.sendCertificates(ctx, 0)
 }
+func (a *AggSender) checkDBCompatibility(ctx context.Context) {
+	if a.compatibilityStoragedChecker == nil {
+		a.log.Warnf("compatibilityStoragedChecker is nil, so we are not going to check the compatibility")
+		return
+	}
+	err := a.compatibilityStoragedChecker.Check(ctx, nil)
+	if err != nil {
```
goran-ethernal, line 155:
```suggestion
	if err := a.compatibilityStoragedChecker.Check(ctx, nil); err != nil {
```

db/compatibility/compatibility_db_contents.go
```diff
@@ -0,0 +1,114 @@
...
+		RequireStorageContentCompatibility: requireStorageContentCompatibility,
+		RuntimeDataGetter:                  runtimeDataGetter,
+		Storage:                            storage,
+	}
+}
+
+// Check checks the compatibility between the runtime data and the data stored in the storage
+func (s *CompatibilityCheck[T]) Check(ctx context.Context, tx db.Querier) error {
+	err := s.initialize()
+	if err != nil {
```
goran-ethernal, line 67: And on other places to lower affected lines number:

```suggestion
	if err := s.initialize(); err != nil {
```

db/compatibility/helper_storage.go
```diff
@@ -0,0 +1,81 @@
...
+	aggkitcommon.AGGSENDER),
+
+*/
+
+const (
+	// compatibilityContentKey is the key used to store the compatibility data in storage key/value table
+	compatibilityContentKey = "compatibility_content"
+)
+
+type KeyValueStorager interface {
```
goran-ethernal, line 38: Add comments on structs and functions.

goran-ethernal (approved):

db/compatibility/compatibility_db_contents.go
```diff
@@ -0,0 +1,114 @@
...
+	s.Logger.Infof("compatibilityCheck: data in DB[%s] is compatible with [%s]",
+		storageData.String(), runtimeData.String())
+
+	return nil
+}
+
+func (s *CompatibilityCheck[T]) initialize() error {
+	if s.Logger == nil {
+		s.Logger = log.WithFields("module", "compatibilityCheck")
+	}
```
Stefan-Ethernal, line 106: Why don't we move this to the constructor function (`NewCompatibilityCheck`)?

db/compatibility/compatibility_db_contents.go
```diff
@@ -0,0 +1,114 @@
...
+			return nil
+		}
+	}
+	s.Logger.Infof("compatibilityCheck: data in DB[%s] is compatible with [%s]",
+		storageData.String(), runtimeData.String())
+
+	return nil
+}
+
+func (s *CompatibilityCheck[T]) initialize() error {
```
Stefan-Ethernal, line 103: WDYT, in order to be semantically more correct?
```suggestion
func (s *CompatibilityCheck[T]) validate() error {
```

aggsender/aggsender.go
```diff
@@ -131,9 +142,20 @@ func (a *AggSender) Start(ctx context.Context) {
...
+		a.log.Fatalf("error checking compatibility data in DB, you can bypass this check using config file")
```
Stefan-Ethernal, line 156: Incorporate the error message to the log as well?
```suggestion
		a.log.Fatalf("error checking compatibility data in DB, you can disable this check using config file: %s", err)
```

sync/evmdriver.go
```diff
@@ -78,7 +124,16 @@ reset:
...
+		if err != nil {
+			attempts++
+			d.log.Error("error checking compatibility data between downloader(runtime) and processor (db): ", err)
```
Stefan-Ethernal, line 131:
```suggestion
			d.log.Error("error checking compatibility data between downloader (runtime) and processor (db): ", err)
```

sync/evmdriver.go
Stefan-Ethernal: It came as a surprise to me as well, checking the compatility on each iteration, as I would also expect to check this only once and move on if valid, otherwise halt (or whatever we find appropriate)...

db/compatibility/compatibility_db_contents.go
```diff
@@ -0,0 +1,114 @@
...
+			return fmt.Errorf("compatibilityCheck: data on DB is [%s] != runtime [%s]. Err: %w",
+				storageData.String(), runtimeData.String(), err)
+		} else {
+			s.Logger.Warnf("compatibilityCheck: data on DB is [%s] != runtime [%s]. Err: %w",
+				storageData.String(), runtimeData.String(), err)
+			return nil
+		}
+	}
+	s.Logger.Infof("compatibilityCheck: data in DB[%s] is compatible with [%s]",
+		storageData.String(), runtimeData.String())
```
Stefan-Ethernal, line 98: Lower the verbosity to debug, wdyt?

```suggestion
	s.Logger.Debugf("compatibilityCheck: data in DB[%s] is compatible with [%s]",
		storageData.String(), runtimeData.String())
```

sync/evmdriver.go
```diff
@@ -20,23 +23,59 @@ type Block struct {
...
+func (r RuntimeData) IsCompatible(other RuntimeData) error {
+	if r.ChainID != other.ChainID {
+		return fmt.Errorf("chain ID mismatch: %d != %d", r.ChainID, other.ChainID)
+	}
+	if len(r.Addresses) != len(other.Addresses) {
+		return fmt.Errorf("addresses len mishmatch: %d != %d", len(r.Addresses), len(other.Addresses))
+	}
+	for i, addr := range r.Addresses {
+		if addr != other.Addresses[i] {
+			return fmt.Errorf("addresses[%d] mishmatch: %s != %s", i, addr.String(), other.Addresses[i].String())
```
Stefan-Ethernal, line 67: Typo
```suggestion
			return fmt.Errorf("addresses[%d] mismatch: %s != %s", i, addr.String(), other.Addresses[i].String())
```

sync/evmdriver.go
```diff
@@ -20,23 +23,59 @@ type Block struct {
...
+	}
+	return res
+}
+
+func (r RuntimeData) IsCompatible(other RuntimeData) error {
+	if r.ChainID != other.ChainID {
+		return fmt.Errorf("chain ID mismatch: %d != %d", r.ChainID, other.ChainID)
+	}
+	if len(r.Addresses) != len(other.Addresses) {
+		return fmt.Errorf("addresses len mishmatch: %d != %d", len(r.Addresses), len(other.Addresses))
```
Stefan-Ethernal, line 63: Typo
```suggestion
		return fmt.Errorf("addresses len mismatch: %d != %d", len(r.Addresses), len(other.Addresses))
```

db/compatibility/compatibility_db_contents.go
joanestebanr: In fact it's doing both things: the field Logger is initialized and the rest checked... I don't mind if you prefer `validate`  for me it's ok. WDYT?

aggsender/aggsender.go
joanestebanr: About adding the Error ok!, about changing 'bypass' to 'disable' I think  that `bypass` is more accurated, beucase you execute the test, but just ignore the error. WDTY?

sync/evmdriver.go
joanestebanr: This is exactly how it works: 
- Check compatibility
- Get last processed block 
- launch downloader

The @goran-ethernal point is that instead of just executing the check and if fails do a Fatal I retry using the `retryHandler` object

WDTY?

db/compatibility/compatibility_db_contents.go
joanestebanr: I prefer a lazy initialization: if I create the object but I don't use it doesn't produce any error, but if you try to use it you need to fullfil the parameters. 

I case that your question if over  populating  `s.Logger`. It's for the same reason, if I don't need Logger I don't need to populate until I'm going to use it, also, in the first version, not setting the Logger variable was an error

WDTY?

aggsender/aggsender.go
Stefan-Ethernal: Ok, leave it as a bypass, but consider adding an error to the log.

PR #258, opened 2025-03-03
docs: l1 info tree index for bridge

temaniarpit27:


goran-ethernal (approved):

rachit77 (approved):

docs/bridge_service.md
```diff
@@ -267,7 +267,27 @@ Failed response (`rpc.Error`)
...
+| `networkID`    | `uint32` | ID of the network to fetch L1 info tree index.                       | Yes          | `0` for L1(otherwise L2)        |
+| `depositCount` | `uint32` | Deposit count of the bridge for we have requested L1 info tree index | Yes          |                                 |
+
+---
+
+#### Return value
+
+Successful response
+
+- uint32
```
rachit77, line 285: on a side note: should we include return type of response param in other APIs docs as well?
cc: @Stefan-Ethernal

Stefan-Ethernal: We have it, aren't we?
https://github.com/agglayer/aggkit/blob/b34f4e887537222ed2de9a4f769bb638cd6fb7b7/docs/bridge_service.md?plain=1#L115-L155

Stefan-Ethernal (approved):

docs/bridge_service.md
temaniarpit27: in this API
there is just an error or uint return so here i have just mentioned this.
Other apis, there is json response so we have included accordingly

PR #276, opened 2025-03-06
fix: Multiplatform docker image

vcastellm:
Add platform param

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ COPY . .
 RUN make build-aggkit build-tools
 
 # CONTAINER FOR RUNNING BINARY
-FROM debian:bookworm-slim
+FROM --platform=${BUILDPLATFORM} debian:bookworm-slim
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
     ca-certificates \


goran-ethernal (approved):

Ekleog-Polygon:
I can confirm that with this change (on top of my local checkout of a632969272775a27381b6a7a3eebb32170d9a4ac), I can run:
```
% docker build -t local-aggkit .
% docker run --rm -ti local-aggkit:latest
```
And it does display the help message of aggkit :)

Ekleog-Polygon (approved):

Ekleog-Polygon:
This being said, even without these changes, on the same commit if I run the same commands it also works. There's just this command that doesn't work on my machine:
```
% docker run --rm -ti ghcr.io/agglayer/aggkit:0.1.0-beta2@sha256:a07da8cb29b0c04c801c84f19510fa1d89863e39a8a8384d23291467e5e80f98
```

Maybe it's just the dockerhub image that's broken and would need rebuilding?

Freyskeyd:
@Ekleog-Polygon to explain the problem:

The aggkit job that handle the `arm` build of the docker image define this element in that [matrix](https://github.com/agglayer/aggkit/blob/main/.github/workflows/release.yml#L30) that is used to configure the docker build-and-push [here](https://github.com/agglayer/aggkit/blob/main/.github/workflows/release.yml#L80)

But the whole thing is building on an `amd64` as defined in this job: https://github.com/agglayer/aggkit/actions/runs/13634610135/job/38110157731
Interesting context is: `Build and push by digest > Docker info` which defines the Arch and Architecture in used.
If docker encounter a `FROM` without an explicit `--platform` it will use the target platform which can be found using: 
`docker system info --format '{{.OSType}}/{{.Architecture}}'`

On the CI, this `OSType.Architecture` seems to be `linux/amd64`. That's why, when building locally on M1 (arm), the build succeed and we do not have the error, because the target platform is `linux/arm64`

PR #279, opened 2025-03-06
chore: bump versions

rachit77:
## Description

Update Go to `1.23.7 `and go-ethereum to `v1.15.5`


l1infotreesync/e2e_test.go
```diff
@@ -214,36 +216,41 @@ func TestWithReorgs(t *testing.T) {
...
+
+	// Commit block 7, 8, 9 after the fork
+	helpers.CommitBlocks(t, client, 5, time.Millisecond*500)
 
-	// Assert rollup exit root after the fork - should be zero since there are no events in the block after the fork
+	// Assert rollup exit root after commiting new blocks on the fork
 	expectedRollupExitRoot, err = verifySC.GetRollupExitRoot(&bind.CallOpts{Pending: false})
 	require.NoError(t, err)
 	actualRollupExitRoot, err = syncer.GetLastRollupExitRoot(ctx)
-	require.ErrorContains(t, err, "not found") // rollup exit tree reorged, it does not have any exits in it
```
temaniarpit27, line 228: can we have a testcase similar to this?
currently it seems we have removed this flow

rachit77: resolving this as discussed

joanestebanr (approved):

bridgesync/e2e_test.go
```diff
@@ -63,35 +60,8 @@ func TestBridgeEventE2E(t *testing.T) {
...
 		bridgesSent++
-		bn, err = setup.L1Environment.SimBackend.Client().BlockNumber(ctx)
-		require.NoError(t, err)
-		finalizedBlockNumber := getFinalizedBlockNumber(t, ctx, setup.L1Environment.SimBackend.Client())
-		blocksToReorg := 1 + i%maxReorgDepth
-		// Trigger reorg but prevent to reorg a finalized block
-		if i%reorgEveryXIterations == 0 && bn-uint64(blocksToReorg) > finalizedBlockNumber {
-			log.Infof("*** Reorg iteration: %d, Reorging %d blocks. From block: %d to %d. finalizedBlockNumber: %d",
-				i, blocksToReorg, bn, bn-uint64(blocksToReorg), finalizedBlockNumber)
-			helpers.Reorg(t, setup.L1Environment.SimBackend, uint64(blocksToReorg))
```
goran-ethernal, line 74: Why did we remove this part of the e2e test?

rachit77: @goran-ethernal there are some changes in simulated backend in new version of geth particularly this new method `SetCanonical` here brings back the transactions from discarded blocks to the txpool again after forking which was not the case previously as it was using `SetHead`
To tackle this we have to remove reorg logics for testing

https://github.com/ethereum/go-ethereum/blob/master/eth/catalyst/simulated_beacon.go#L329
https://github.com/ethereum/go-ethereum/blob/v1.14.5/eth/catalyst/simulated_beacon.go#L292

goran-ethernal: But then that means we are not testing the reorg logic. I think that is the point of the test 😕

rachit77: yes this change makes this useless at this point
i don't see any other way to do this with geth simulated backend new version 🫠

goran-ethernal (approved):

PR #282, opened 2025-03-10
chore: remove unwanted files

temaniarpit27:
#281


state/batch.go
```diff
@@ -29,5 +28,4 @@ type Batch struct {
 	WIP     bool
 	ChainID uint64
 	ForkID  uint64
-	Type    datastream.BatchType
```
Stefan-Ethernal, line 32: Can we remove the Batch struct altogether? Maybe there are some more leftovers.

Stefan-Ethernal (approved):

goran-ethernal (approved):

PR #307, opened 2025-03-14
fix: v0.2.0-beta1

rachit77:
## Description

Update migration files for `v0.2.0-beta1`


joanestebanr (approved):
There are any reason to create a new migration file for bridge? I think that is not release yet, so you can use previous sql file.

rachit77:
> There are any reason to create a new migration file for bridge? I think that is not release yet, so you can use previous sql file.

actually we released `v0.2.0-beta1` yesterday with modification on existing sql file and that was throwing error on network upgrades.

PR #316, opened 2025-03-18
fix: Reorgdetector 

rachit77:
## Description

The current implementation of `getTrackedBlocks` fails to correctly handle the scenario where the `SubscriberID` changes on the last element of `headersWithID`. Rather than creating a new `HeadersList` for the new `SubscriberID`, it mistakenly adds it to the existing `HeadersList` of a different `SubscriberID`.

### DIFF
--- a/reorgdetector/reorgdetector_db.go
+++ b/reorgdetector/reorgdetector_db.go
@@ -26,27 +26,22 @@ func (rd *ReorgDetector) getTrackedBlocks() (map[string]*headersList, error) {
 	}
 	currentID := headersWithID[0].SubscriberID
 	currentHeaders := []header{}
-	for i := 0; i < len(headersWithID); i++ {
-		if i == len(headersWithID)-1 {
-			currentHeaders = append(currentHeaders, header{
-				Num:  headersWithID[i].Num,
-				Hash: headersWithID[i].Hash,
-			})
+	for _, row := range headersWithID {
+		// If the subscriber ID changes, save the current group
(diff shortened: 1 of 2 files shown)


reorgdetector/reorgdetector_db.go
```diff
@@ -26,28 +26,23 @@ func (rd *ReorgDetector) getTrackedBlocks() (map[string]*headersList, error) {
...
 	currentID := headersWithID[0].SubscriberID
 	currentHeaders := []header{}
-	for i := 0; i < len(headersWithID); i++ {
-		if i == len(headersWithID)-1 {
-			currentHeaders = append(currentHeaders, header{
-				Num:  headersWithID[i].Num,
-				Hash: headersWithID[i].Hash,
-			})
+	for _, row := range headersWithID {
+		// If the subscriber ID changes, save the current group
```
goran-ethernal, line 30: Just curious, how can subscriber ID change?

goran-ethernal: I've took a look again to the code, and I understand it now. I will resolve the comment.

goran-ethernal (approved):
LGTM. But I would suggest adding a UT that tests this situation.

rachit77:
> LGTM. But I would suggest adding a UT that tests this situation.

added [9163ee2](https://github.com/agglayer/aggkit/pull/316/commits/9163ee2f1aab654d7de0cbb5805ca8e01197c8f4)

Issue #328, opened 2025-03-24
migrate e2e cases from aggkit to e2e repo


temaniarpit27:
currently the e2e/bats tests sit in aggkit repo. we want to move these tests to a common repo `agglayer/e2e` for more common work and use case. we will call the tests from aggkit repo with required params and the tests will kick off.

PR #333, opened 2025-03-27
feat: migrate tests

temaniarpit27:
#328


.github/workflows/test-e2e-single-chain.yml
```diff
@@ -1,77 +1,64 @@
...
+            "gas_token_address": "",
+            "agglayer_prover_sp1_key": "",
+            "zkevm_use_real_verifier": false,
+            "agglayer_prover_primary_prover": "mock-prover",
+            "enable_normalcy": true,
+            "zkevm_rollup_id": 1,
+            "binary_name": "aggkit",
+            "components": "aggsender,bridge"
+          }
+        }
```
Stefan-Ethernal, line 54: IMO it is cleaner if we could keep the Kurtosis CDK params in the yml file.  WDYT?

Stefan-Ethernal (commented):
We should probably adapt these `make` commands, or remove it altogether?
https://github.com/agglayer/aggkit/blob/8fb235f3edddd3ca4d2ad4fa70bb396ade7f25f5/test/Makefile#L7-L19.

Also figure out how to support the use case of local debugging the aggkit node and adapt the (readme file)[https://github.com/agglayer/aggkit/blob/8fb235f3edddd3ca4d2ad4fa70bb396ade7f25f5/docs/aggsender.md#L154-L155] accordingly

.github/workflows/test-e2e-single-chain.yml
temaniarpit27: will change

PR #334, opened 2025-03-27
refactor: LastGERSync

rachit77:
## Description

Issue #294


lastgersync/evmdownloader.go
```diff
@@ -45,7 +45,7 @@ func newDownloader(
 	gerContract, err := globalexitrootmanagerl2sovereignchain.NewGlobalexitrootmanagerl2sovereignchain(
 		l2GERAddr, l2Client)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to initialize L2 GER contract: %w", err)
```
Stefan-Ethernal, line 48: WDYT?
```suggestion
		return nil, fmt.Errorf("failed to initialize L2 GER manager contract: %w", err)
```

lastgersync/lastgersync.go
```diff
@@ -69,16 +72,23 @@ func New(
 	}, nil
 }
 
-func (s *LastGERSync) Start(ctx context.Context) {
+// Start initiates the synchronization process.
+func (s *LastGERSync) Start(ctx context.Context) error {
+	if s.driver == nil {
+		return fmt.Errorf("driver is not initialized")
+	}
```
Stefan-Ethernal, line 79: This should never happen, because from what I can see we are always invoking the `New` constructor function (which initializes the [driver](https://github.com/agglayer/aggkit/blob/99fae0611f85694e931c3eae0efa85d7101646c8/lastgersync/lastgersync.go#L64-L67)). WDYT?

```suggestion
```

This also means, that we wouldn't need to propagate an error from `Start` function.

lastgersync/processor.go
```diff
@@ -106,22 +85,57 @@ func (p *processor) ProcessBlock(ctx context.Context, block sync.Block) error {
...
+
+// GetLastIndex retrieves the highest L1InfoTreeIndex recorded in the imported_global_exit_root table
+func (p *processor) getLastIndex() (uint32, error) {
+	var lastIndex uint32
+	row := p.db.QueryRow(`
+		SELECT l1_info_tree_index 
+		FROM imported_global_exit_root 
+		ORDER BY l1_info_tree_index DESC LIMIT 1;
+	`)
+	err := row.Scan(&lastIndex)
```
Stefan-Ethernal, line 117: Optional nitpick, move the declaration closer to its usage.

```suggestion
	row := p.db.QueryRow(`
		SELECT l1_info_tree_index 
		FROM imported_global_exit_root 
		ORDER BY l1_info_tree_index DESC LIMIT 1;
	`)
    var lastIndex uint32
	err := row.Scan(&lastIndex)
```

lastgersync/processor.go
```diff
@@ -106,22 +85,57 @@ func (p *processor) ProcessBlock(ctx context.Context, block sync.Block) error {
...
+
+// Reorg removes all blocks and associated data starting from a specific block number from lastgersync database
 func (p *processor) Reorg(ctx context.Context, firstReorgedBlock uint64) error {
-	_, err := p.db.Exec(`DELETE FROM block WHERE num >= $1;`, firstReorgedBlock)
-	return fmt.Errorf("error processing reorg: %w", err)
+	_, err := p.db.ExecContext(ctx, `DELETE FROM block WHERE num >= $1;`, firstReorgedBlock)
+	if err != nil {
+		return fmt.Errorf("error processing reorg: %w", err)
+	}
+	return nil
```
Stefan-Ethernal, line 130: :raised_hands: :100:

Stefan-Ethernal (approved):
LGTM and nice overall. Leaving a couple of nitpick comments to consider.

As a side note, it would be worth considering if we can make it last GER syncer follow the same pattern as the other syncers (to declare `LogAppenderMap` as it is done in the [bridge syncer](https://github.com/agglayer/aggkit/blob/99fae0611f85694e931c3eae0efa85d7101646c8/bridgesync/downloader.go#L42-L174)). However it can be a follow up task.

lastgersync/lastgersync.go
rachit77: You are right, I will remove this check

PR #348, opened 2025-04-02
feat: use external  signer library to support GCP KMS

joanestebanr:
## Description

- The code for signer have been moved to a [go_signer repo](https://github.com/agglayer/go_signer) . So `aggkit` must use it
- Adding version 0.0.5 also add support to google cloud KMS


aggsender/aggsender.go
```diff
@@ -77,7 +78,7 @@ func New(
 	if err != nil {
 		return nil, err
 	}
-	signer, err := signer.NewSigner(aggkitcommon.AGGSENDER, logger, ctx, cfg.AggsenderPrivateKey)
+	signer, err := signer.NewSigner(ctx, 0, cfg.AggsenderPrivateKey, aggkitcommon.AGGSENDER, logger)
```
goran-ethernal, line 81: Do we need the chain ID here? I know we are not signing a transaction de facto, but I can not recall if chain ID is needed. Probably not, but just to be sure, I have to ask 😄 .

goran-ethernal (approved):
LGTM

aggsender/aggsender.go
joanestebanr: No, for signing a hash the chainID is not used

go.mod
```diff
@@ -36,24 +36,46 @@ require (
...
+	cloud.google.com/go/iam v1.2.2 // indirect
+	cloud.google.com/go/kms v1.20.1 // indirect
+	cloud.google.com/go/longrunning v0.6.2 // indirect
 	github.com/0xPolygonHermez/zkevm-synchronizer-l1 v1.0.7 // indirect
+	github.com/BurntSushi/toml v1.4.0 // indirect
 	github.com/DataDog/zstd v1.5.6 // indirect
 	github.com/Microsoft/go-winio v0.6.2 // indirect
-	github.com/StackExchange/wmi v1.2.1 // indirect
 	github.com/VictoriaMetrics/fastcache v1.12.2 // indirect
+	github.com/agglayer/go_signer v0.0.5 // indirect
```
Stefan-Ethernal, line 51: Run the `go mod tidy`, as this should be a direct dependency

joanestebanr: done! good catch!

Stefan-Ethernal (approved):

PR #354, opened 2025-04-03
feat: update `aggsender` based on update to `GenerateAggchainProof` API on the `aggkit prover`

goran-ethernal:
## Description

This PR updates the `aggsender` based on latest updates on the `aggkit prover` API.

The `aggkit prover` now returns additional data in the `GenerateAggchainProof`:
- `CustomChainData` - `[]byte`
- `AggchainParams` - `common.Hash`
- `Context` - `map[string][]byte'
- `SP1StarkProof` - which contains the `Proof ([]byte)`, `VKey ([]byte)` and `Version (string)`.

All of this data will be sent to the `agglayer` in the `certificate`.

`GenerateAggchainProof` request also does not need the entire data from the `imported bridge exits`, so a new struct is created that will be sent to the `prover` to generate the proof:

```go
type ImportedBridgeExitWithBlockNumber struct {
     BlockNumber uint64 // blockNumber on which claim (importedBridgeExit) event was seen on L2
     GlobalIndex common.Hash // global index of the claim (importedBridgeExit)
     BridgeExitHash common.Hash // (hash of the corresponding bridgeExit of given claim/importedBridgeExit)
}
```

Fixes # (issue)


vcastellm (commented):
LGTM

vcastellm (approved):

vcastellm (approved):

aggsender/db/migrations/0002.sql
```diff
@@ -5,7 +5,7 @@ ALTER TABLE certificate_info_history DROP COLUMN aggchain_proof;
 ALTER TABLE certificate_info_history DROP COLUMN finalized_l1_info_tree_root;
 
 -- +migrate Up
-ALTER TABLE certificate_info ADD COLUMN aggchain_proof VARCHAR;
+ALTER TABLE certificate_info ADD COLUMN aggchain_proof BLOB;
```
temaniarpit27, line 8: @goran-ethernal should we add some default value?
i think db fetching might give an error for entries with null value

aggsender/grpc/aggchain_proof_client.go
```diff
@@ -126,122 +128,18 @@ func (c *AggchainProofClient) GenerateAggchainProof(
 	convertedImportedBridgeExitsWithBlockNumber := make([]*aggkitProverV1Proto.ImportedBridgeExitWithBlockNumber,
 		len(importedBridgeExitsWithBlockNumber))
 	for i, importedBridgeExitWithBlockNumber := range importedBridgeExitsWithBlockNumber {
-		convertedImportedBridgeExit := &agglayerInteropTypesV1Proto.ImportedBridgeExit{
-			BridgeExit: &agglayerInteropTypesV1Proto.BridgeExit{
```
temaniarpit27, line 130: dont we need bridge exits now?

aggsender/db/migrations/0002.sql
goran-ethernal: It doesn't return an error, because I wrote a custom meddler for this field.

aggsender/grpc/aggchain_proof_client.go
goran-ethernal: No, just a bridge exit hash.

aggsender/types/certificate_build_params.go
```diff
@@ -23,8 +23,7 @@ type CertificateBuildParams struct {
 	RetryCount                     int
 	LastSentCertificate            *CertificateInfo
 	L1InfoTreeRootFromWhichToProve *treetypes.Root
-	AggchainProof                  []byte
-	CustomChainData                []byte
```
temaniarpit27, line 27: is this removed?

aggsender/types/types.go
```diff
@@ -168,15 +220,15 @@ func (c *CertificateInfo) String() string {
...
 		time.Unix(int64(c.CreatedAt), 0),
 		time.Unix(int64(c.UpdatedAt), 0),
-		c.AggchainProof,
+		aggchainProof,
 		finalizedL1InfoTreeRoot,
 	)
 }
 
 // ID returns a string with the unique identifier of the cerificate (height+certificateID)
 func (c *CertificateInfo) ID() string {
```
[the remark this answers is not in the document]
goran-ethernal: Function is on the pointer of `Ceritificate` info, so it is safe to call it on a nil object. That is why we have this condition in the function:

```
if c == nil {
    return "nil"
}
```

aggsender/types/certificate_build_params.go
goran-ethernal: It is not removed, just moved to the new field:

```
AggchainProof *AggchainProof
```

Which is basically everything the `prover` returned on `GenerateAggchainProof` request.

temaniarpit27 (approved):

aggsender/db/migrations/0002.sql
```diff
@@ -5,7 +5,7 @@ ALTER TABLE certificate_info_history DROP COLUMN aggchain_proof;
...
 ALTER TABLE certificate_info ADD COLUMN finalized_l1_info_tree_root VARCHAR;
-ALTER TABLE certificate_info_history ADD COLUMN aggchain_proof VARCHAR;
+ALTER TABLE certificate_info_history ADD COLUMN aggchain_proof BLOB;
```
Stefan-Ethernal, line 10: Shall these changes be part of a new migration file? It is the same doubt we had for bridge service: when deploying these changes, are they doing the fresh start, or just make an upgrade (use the newer binaries against the old data)?

aggsender/types/types.go
```diff
@@ -16,6 +16,8 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 )
 
+const nilStr = "nil"
```
Stefan-Ethernal, line 19: Nitpick, consider exporting this constant and using it instead of this [one](https://github.com/agglayer/aggkit/blob/cca60035aa19700242b316f15b433e109db6b63a/aggsender/aggsender_initial_state.go#L20), in the `aggsender` pkg.

db/meddler.go
```diff
@@ -57,6 +57,11 @@ func SlicePtrsToSlice(slice interface{}) interface{} {
 	return res.Interface()
 }
 
+// RegisterMeddlerType registers a new meddler type with the given name
+func RegisterMeddler(name string, meddlerType meddler.Meddler) {
+	meddler.Register(name, meddlerType)
+}
```
Stefan-Ethernal, line 63: Nitpick, why not directly using the `meddler.Register` function, instead of this wrapper function?

```suggestion
```

aggsender/prover/proof_generation_tool.go
```diff
@@ -104,7 +104,7 @@ func (a *AggchainProofGenerationTool) GetRPCServices() []rpc.Service {
 // GenerateAggchainProof generates an Aggchain proof
 func (a *AggchainProofGenerationTool) GenerateAggchainProof(
```
Stefan-Ethernal, line 105: Should we rename it to `GenerateSP1StarkProof` perhaps (also update the log messages)?

Stefan-Ethernal (approved):
:100: Nice overall.

aggsender/db/migrations/0002.sql
goran-ethernal: No. This is not an official release.

db/meddler.go
goran-ethernal: Like I wrote in the comment where this is called, I need to ensure that `db` package and his `init` function is called before calling the register of a custom meddler.

aggsender/prover/proof_generation_tool.go
goran-ethernal: I want to be consistent with the prover API, a their function is called `GenerateAggchainProof`.

aggsender/types/types.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/354/commits/00fabf68972e6a8690351f09bfcadccbe0fb4e9d

PR #355, opened 2025-04-04
feat: update `agglayer` API on `aggsender`

goran-ethernal:
## Description

This PR updates `agglayer` API for sending certificates. Now the `agglayer` receives `Vkey` and `Version` alongside the `SP1StarkProof` in the certificate.

This data is returned by the `aggkit-prover` to `aggsender` and `aggsender` packs it to a certificate and sends it to `agglayer`.

Fixes # (issue)

### DIFF
--- a/agglayer/grpc/agglayer_grpc_client.go
+++ b/agglayer/grpc/agglayer_grpc_client.go
@@ -6,8 +6,9 @@ import (
 	"strings"
 
 	node "buf.build/gen/go/agglayer/agglayer/grpc/go/agglayer/node/v1/nodev1grpc"
+	v1nodetypes "buf.build/gen/go/agglayer/agglayer/protocolbuffers/go/agglayer/node/types/v1"
 	v1 "buf.build/gen/go/agglayer/agglayer/protocolbuffers/go/agglayer/node/v1"
-	v1Types "buf.build/gen/go/agglayer/agglayer/protocolbuffers/go/agglayer/protocol/types/v1"
+	v1types "buf.build/gen/go/agglayer/interop/protocolbuffers/go/agglayer/interop/types/v1"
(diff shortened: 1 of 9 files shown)


agglayer/grpc/agglayer_grpc_client.go
```diff
@@ -213,45 +218,48 @@ func convertProtoCertificateHeader(response *v1Types.CertificateHeader) *types.C
...
-			OriginTokenAddress: &v1Types.FixedBytes20{
+			OriginTokenAddress: &v1types.FixedBytes20{
 				Value: be.TokenInfo.OriginTokenAddress.Bytes(),
 			},
 		},
 	}
 
+	if be.Amount != nil {
+		protoBridgeExit.Amount = &v1types.FixedBytes32{
+			Value: common.BigToHash(be.Amount).Bytes(),
```
Stefan-Ethernal, line 242: Optional nitpick, but for sake of semantical correctness (although I'm aware that under the hood, this is the same), maybe we can add the following function to aggkit's common package, and use it here, wdyt?:

```go
func BigIntToBytes32(n *big.Int) [32]byte {
	var b32 [32]byte
	bytes := n.FillBytes(make([]byte, 32)) // fills a 32-byte slice with big-endian representation
	copy(b32[:], bytes)
	return b32
}

```

Stefan-Ethernal (approved):

Stefan-Ethernal (approved):

agglayer/grpc/agglayer_grpc_client.go
goran-ethernal: Its nice, but, why implement something we already have. As far as for logic is concerned, I checked it, the existing logic is not more complex than this one.

PR #377, opened 2025-04-14
feat: add block index to injected GERs we send to the prover

goran-ethernal:
## Description

This PR adds the `block index` field to the `ProvenInsertedGERWithBlockNumber` and sends it to the `aggkit prover` so that the prover can properly sort the injected GERs when building a proof.

Fixes # (issue)

### DIFF
--- a/aggoracle/chaingerreader/evm.go
+++ b/aggoracle/chaingerreader/evm.go
@@ -58,12 +58,13 @@ func checkGlobalExitRootManagerContract(l2GERManager types.L2GERManagerContract,
 
 // GetInjectedGERsForRange returns the injected GlobalExitRoots for the given block range
 func (e *EVMChainGERReader) GetInjectedGERsForRange(ctx context.Context,
-	fromBlock, toBlock uint64) (map[uint64][]InjectedGER, error) {
+	fromBlock, toBlock uint64) (map[common.Hash]InjectedGER, error) {
 	if fromBlock > toBlock {
 		return nil, fmt.Errorf("invalid block range: fromBlock(%d) > toBlock(%d)", fromBlock, toBlock)
 	}
 
-	iter, err := e.l2GERManager.FilterUpdateHashChainValue(
+	// first get all inserted GERs in the block range
+	insertIter, err := e.l2GERManager.FilterUpdateHashChainValue(
 		&bind.FilterOpts{
 			Context: ctx,
 			Start:   fromBlock,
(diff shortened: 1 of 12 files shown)


aggoracle/chaingerreader/evm.go
```diff
@@ -11,6 +11,16 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 )
 
+// InjectedGER is a struct that represents the injected GlobalExitRoot event
+type InjectedGER struct {
+	// BlockNumber is the block number of the event
+	BlockNumber uint64
+	// BlockIndex is the index of the event in the block
+	BlockIndex uint
```
vcastellm, line 19: hmm, sorry for not commenting it in the other PR, but should this be `EventIndex`?

goran-ethernal: I mean it could. On `aggkit` we call these fields block index or even block position. That is why I went with this.

aggoracle/chaingerreader/evm.go
```diff
@@ -73,7 +83,11 @@ func (e *EVMChainGERReader) GetInjectedGERsForRange(ctx context.Context,
 
 		eventBlockNum := iter.Event.Raw.BlockNumber
 
-		injectedGERs[eventBlockNum] = append(injectedGERs[eventBlockNum], iter.Event.NewGlobalExitRoot)
+		injectedGERs[eventBlockNum] = append(injectedGERs[eventBlockNum], InjectedGER{
+			BlockNumber:    eventBlockNum,
+			BlockIndex:     iter.Event.Raw.Index,
+			GlobalExitRoot: iter.Event.NewGlobalExitRoot,
+		})
 	}
```
Stefan-Ethernal, line 92: IMO we should handle the UpdateHashChainRemoval events and return only those GERs that are not removed in a meantime (https://github.com/agglayer/agg-contracts-internal/blob/feature/ongoing-v0.3.0/contracts/v2/sovereignChains/GlobalExitRootManagerL2SovereignChain.sol#L48-L51).

aggsender/flows/flow_aggchain_prover_test.go
```diff
@@ -120,7 +121,7 @@ func Test_AggchainProverFlow_GetCertificateBuildParams(t *testing.T) {
 					nil,
 				)
 				mockL1InfoDataQuery.On("CheckIfClaimsArePartOfFinalizedL1InfoTree", mock.Anything, mock.Anything).Return(nil)
-				mockChainGERReader.On("GetInjectedGERsForRange", ctx, uint64(1), uint64(10)).Return(map[uint64][]common.Hash{}, nil)
+				mockChainGERReader.On("GetInjectedGERsForRange", ctx, uint64(1), uint64(10)).Return(map[uint64][]chaingerreader.InjectedGER{}, nil)
```
Stefan-Ethernal, line 124: Nitpick, you could use expecter (`EXPECT()`) here, as it is more developer friendly.

aggoracle/chaingerreader/evm.go
Stefan-Ethernal: Yes it is strange (I don't like the way we are naming them in syncers, but it is what it is). It would be straightforward to call it as @vcastellm suggested.

Stefan-Ethernal (approved):
:100: LGTM, aside from some minor suggestions.

vcastellm (approved):
LGTM aside from comment for upcoming PR

aggoracle/chaingerreader/evm.go
goran-ethernal: Since this PR already got merged on the prover side, lets leave it as is for now. I'll write this down so we can change it when we get the happy path working.

aggsender/flows/flow_aggchain_prover_test.go
goran-ethernal: I agree, but I would leave that for some cleanup PR, so we can change all UTs to use expecter.

aggoracle/chaingerreader/evm.go
goran-ethernal: Agreed. I will add that logic.

goran-ethernal: Done: https://github.com/agglayer/aggkit/pull/377/commits/689daf2ad01a379704262ac491e30292843344d5

aggsender/flows/flow_aggchain_prover_test.go
goran-ethernal: Done: https://github.com/agglayer/aggkit/pull/377/commits/689daf2ad01a379704262ac491e30292843344d5

PR #378, opened 2025-04-14
feat: allow empty certificates for `FEP` (aggchain proof) flow

goran-ethernal:
## Description

This PR allows sending of empty certificates (certificates without bridge exits and claims) to the `agglayer` in case of `AggchainProof` mode on `aggsender`. This is needed so that `FEP` can work properly on the network.

Fixes # (issue)


Stefan-Ethernal (approved):
:100: LGTM

PR #386, opened 2025-04-15
fix: resend proof in `AggchainProof` flow

goran-ethernal:
## Description

This PR fixes the way `InError` certificate is resent to the `agglayer` in the new `FEP` (`AggchainProof`) mode on `aggsender`.

Since proof was already generated by the prover, we just need to resend it back to the `agglayer`. No need to call the prover again.

Fixes # (issue)

### DIFF
--- a/aggsender/aggsender.go
+++ b/aggsender/aggsender.go
@@ -311,6 +311,12 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certifi
 	}
 	certificateHash, err := a.aggLayerClient.SendCertificate(ctx, certificate)
 	if err != nil {
+		raw, unmarshalErr := json.Marshal(certificate)
+		if unmarshalErr == nil {
+			// we ignore the marshal error, since marshaled certificate is only needed for logging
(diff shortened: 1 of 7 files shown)


aggsender/aggsender.go
```diff
@@ -311,6 +311,12 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certifi
 	}
 	certificateHash, err := a.aggLayerClient.SendCertificate(ctx, certificate)
 	if err != nil {
+		raw, unmarshalErr := json.Marshal(certificate)
+		if unmarshalErr == nil {
```
Stefan-Ethernal, line 315:
```suggestion
		raw, marshalErr := json.Marshal(certificate)
		if marshalErr == nil {
```

vcastellm (approved):
LGTM

Stefan-Ethernal (approved):
:100: LGTM

aggsender/aggsender.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/386/commits/09ed15af3168a60e446af5f20288f9a212694fa3

PR #388, opened 2025-04-15
fix: new `LER` on empty certificate

goran-ethernal:
## Description

This PR fixes issue when in new `FEP` (`AggchainFlow`), we have a certificate which is empty (has no bridge transactions), a wrong `new_local_exit_root` is sent to the `agglayer`. Basically, it is a certificate that has no change in the `LocalExitRoot`, and we need to return the `previous_ler` in this case.

The error that happens is:

```
Mismatch on the certificate new local exit root. declared: 1c9a7bedb395b0debf452130c2de1fbc3d1e5c43f720713cfc6ab98ff0277f6e, computed: 27ae5ba08d7291c96c8cbddcc148bf48a6d68c7974b94356f53754ef6171d757"
```

Fixes # (issue)


vcastellm (approved):
LGTM

PR #390, opened 2025-04-15
feat: send finalized `l1_info_root` in cert

goran-ethernal:
## Description

This PR implements sending finalized `l1 info root` in the certificate to the agglayer.

This finalized `l1 info root` is the same `aggsender` sends to the `aggkit prover` to generate the `aggchain proof`, and the one that is used to generate merkle proofs for each `imported bridge exit` in the certificate.

Fixes # (issue)


Stefan-Ethernal (approved):
:100:

PR #391, opened 2025-04-16
feat: bridge service return consistent error codes

### LINKED ISSUE #392: Return consistent error codes from bridge service

Stefan-Ethernal:
## Description

- Propagate consistent error codes in all endpoints exposed by the bridge service.
- Return empty result set, instead of error in case the result is not found.
- Update the Open RPC definition with the missing `bridge_getLegacyTokenMigrations` function.

Fixes #392

### DIFF
--- a/bridgesync/processor.go
+++ b/bridgesync/processor.go
@@ -418,6 +418,10 @@ func (p *processor) GetBridges(
 	}()
 	rows, err := p.queryBlockRange(tx, fromBlock, toBlock, bridgeTableName)
 	if err != nil {
+		if errors.Is(err, db.ErrNotFound) {
+			p.log.Debugf("no bridges were found for block range [%d..%d]", fromBlock, toBlock)
+			return []Bridge{}, nil
+		}
 		return nil, err
 	}
 	bridgePtrs := []*Bridge{}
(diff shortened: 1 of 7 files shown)


rachit77 (approved):
💯

bridgesync/processor.go
```diff
@@ -485,13 +493,19 @@ func (p *processor) GetBridgesPaged(
...
 	}
 	offset := (pageNumber - 1) * pageSize
-	if offset >= uint32(count) {
-		p.log.Debugf("offset is larger than total bridges (page number=%d, page size=%d, total bridges=%d)",
-			pageNumber, pageSize, count)
-		return nil, 0, db.ErrNotFound
+	if offset >= uint32(bridgesCount) {
+		p.log.Debugf("offset is greater than total bridges (page number=%d, page size=%d, total bridges=%d)",
+			pageNumber, pageSize, bridgesCount)
+		return nil, 0, fmt.Errorf("provided offset is greater than the total bridges count (offset=%d, total count=%d)",
```
goran-ethernal, line 499: I would just change the messages in these `offset >= count` errors:

```suggestion
		return nil, 0, fmt.Errorf("provided page number is invalid for given page size and total number of bridges (page number=%d, page size=%d, total count=%d)",
```

goran-ethernal (approved):
LGTM 💯

bridgesync/processor.go
Stefan-Ethernal: Fixed https://github.com/agglayer/aggkit/pull/391/commits/c0cea262cc91f7adf7ea18708f48cef10fbf46fb

PR #396, opened 2025-04-17
feat: code quality improvement by introducing `bridgeQuerier` and `gerQuerier` on `aggsender`

goran-ethernal:
## Description

This PR introduces a new interface `BridgeDataQuerier` which encapsulates all calls to the bridge syncer needed on `aggsender` in order to have better code separation, code quality and easier testing (mocking).

It also introduces `GERQuerier` interface which encapsulates all calls to the chain GER reader and generating proofs for injected GERs,  in order to have better code separation, code quality and easier testing (mocking).

This PR also introduced a `flow factory`, which will return an instance of `AggsenderFlow` based on the `aggsender` configuration.

Fixes # (issue)


aggsender/flows/flow_base.go
```diff
@@ -42,9 +42,9 @@ func (f *baseFlow) getBridgesAndClaims(
 	fromBlock, toBlock uint64,
 	allowEmptyCert bool,
 ) ([]bridgesync.Bridge, []bridgesync.Claim, error) {
-	bridges, err := f.l2Syncer.GetBridges(ctx, fromBlock, toBlock)
+	bridges, claims, err := f.l2BridgeQuerier.GetBridgesAndClaims(ctx, fromBlock, toBlock)
```
joanestebanr: maybe make more sense to divide this call to getBridges, check the condition and GetClaims. because if the next if `len(bridges)` is true you have already waste a time getting the Claims. WDTY?

aggsender/flows/flow_base.go
```diff
@@ -183,7 +178,7 @@ func (f *baseFlow) buildCertificate(ctx context.Context,
 	)
 
 	return &agglayertypes.Certificate{
-		NetworkID:           f.l2Syncer.OriginNetwork(),
+		NetworkID:           f.l2BridgeQuerier.OriginNetwork(),
```
joanestebanr, line 156: In aggsender struct you set this value in the constructor. maybe make sense to do the same here instead of calling to `l2BridgeQuerier`

aggsender/query/bridge_query.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/agglayer/aggkit/aggsender/types"
+	"github.com/agglayer/aggkit/bridgesync"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+var _ types.BridgeDataQuerier = (*bridgeDataQuerier)(nil)
+
+// bridgeDataQuerier is a struct that holds the logic to query the bridge data
+type bridgeDataQuerier struct {
```
joanestebanr, line 17: It's a bit confusing that the interface and the implementation have the same name, WDYT?

aggsender/query/ger_query.go
```diff
@@ -0,0 +1,86 @@
...
+	agglayertypes "github.com/agglayer/aggkit/agglayer/types"
+	"github.com/agglayer/aggkit/aggsender/types"
+	treetypes "github.com/agglayer/aggkit/tree/types"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+var _ types.GERQuerier = (*gerQuerier)(nil)
+
+// gerQuerier is a struct that holds the logic to query the GER (Global Exit Root) data
+type gerQuerier struct {
```
joanestebanr: Same name interface and implementation

aggsender/types/types.go
```diff
@@ -61,6 +61,17 @@ type L2BridgeSyncer interface {
 	GetLastProcessedBlock(ctx context.Context) (uint64, error)
 }
 
+// BridgeDataQuerier is an interface defining functions that an BridgeDataQuerier should implement
+type BridgeDataQuerier interface {
```
joanestebanr: maybe this interfaces make sense to have his own file

joanestebanr (commented):
The change of name of incoming params to internal ones like: 
- aggsender constructor receive `l2Syncer`  that is passed to `flows.NewFlow` that receives as `l2BridgeSyncer` make a bit difficult to track variables. WDYT?
-

aggsender/flows/flow_base.go
goran-ethernal: It becomes very difficult to mock the `newAggsenderTestData` function. Because, now the base flow would need to get this through the constructor, and we want to control the function mocks in the test itself, and not in this `newAggsenderTestData` function. I would leave it as it is, since it is basically, just a field in the `l2BridgeQuerier`. Its not something that is read from some contract or a `db`.

aggsender/types/types.go
goran-ethernal: Moved all the interfaces from `types.go` to `interfaces.go`:

https://github.com/agglayer/aggkit/pull/396/commits/17d6d20218fb2de21fb3370fda229b6fedae17ed

aggsender/query/ger_query.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/396/commits/17d6d20218fb2de21fb3370fda229b6fedae17ed

aggsender/query/bridge_query.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/396/commits/17d6d20218fb2de21fb3370fda229b6fedae17ed

aggsender/flows/flow_base.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/396/commits/17d6d20218fb2de21fb3370fda229b6fedae17ed

Stefan-Ethernal (approved):
:100: :fire: :fire_extinguisher: LGTM!

vcastellm (approved):
LGTM :100:

PR #397, opened 2025-04-19
fix: panic on resend `InError` cert

goran-ethernal:
## Description

This PR fixes a panic in `AggchainProof` mode on `aggsender` that happens on resending an `InError` certificate when the `aggsender db` gets deleted.

What happened is that the `agglayer` marked a certificate as `InError`, after which the user delted the `aggsender db` and started `aggsender` again. `aggsender` will get the last certificate on `agglayer` and save it to its `db` since it has a cleaned state, and it needs to know from where to start. While saving this, it has no information about the proof and finalized l1 info tree root that was used to send that certificate that was marked as `InError`, so when it tried to resend it, it panicked since it didn't have the proof. 

What this PR does is it recognizes this case, and calls the prover again to get the proof.

Fixes # (issue)

### DIFF
--- a/aggsender/flows/flow_aggchain_prover.go
+++ b/aggsender/flows/flow_aggchain_prover.go
@@ -100,18 +100,32 @@ func (a *AggchainProverFlow) GetCertificateBuildParams(ctx context.Context) (*ty
 			return nil, fmt.Errorf("aggchainProverFlow - error getting bridges and claims: %w", err)
 		}
 
-		return &types.CertificateBuildParams{
-			FromBlock:                      lastSentCertificateInfo.FromBlock,
-			ToBlock:                        lastSentCertificateInfo.ToBlock,
-			RetryCount:                     lastSentCertificateInfo.RetryCount + 1,
-			Bridges:                        bridges,
-			Claims:                         claims,
-			LastSentCertificate:            lastSentCertificateInfo,
-			CreatedAt:                      lastSentCertificateInfo.CreatedAt,
-			AggchainProof:                  lastSentCertificateInfo.AggchainProof,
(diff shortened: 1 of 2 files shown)


vcastellm (approved):
LGTM 💯

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -125,6 +139,13 @@ func (a *AggchainProverFlow) GetCertificateBuildParams(ctx context.Context) (*ty
...
+	return a.verifyBuildParamsAndGenerateProof(ctx, buildParams)
+}
+
+// verifyBuildParams verifies the certificate build params and returns an error if they are not valid
+// it also calls the prover to get the aggchain proof
+func (a *AggchainProverFlow) verifyBuildParamsAndGenerateProof(
+	ctx context.Context, buildParams *types.CertificateBuildParams) (*types.CertificateBuildParams, error) {
 	if err := a.verifyBuildParams(buildParams); err != nil {
 		return nil, fmt.Errorf("aggchainProverFlow - error verifying build params: %w", err)
 	}
```
Stefan-Ethernal, line 151: Just a thought:
I would prefer having the `verifyBuildParams` invoked outside of this function, to have better separation of concerns (this function IMO should only be in charge of generating the proof).

Stefan-Ethernal (approved):
:100: LGTM

aggsender/flows/flow_aggchain_prover.go
goran-ethernal: We need to do it before calling the prover, to not waste computational resources when the proof won't be generated if the data we sent was not verified first. Which can easily be done on aggsender before calling the prover.

Issue #360, opened 2025-04-04
Add a native token indicator to bridge event


Stefan-Ethernal:
In case `originTokenAddress == gasTokenAddress`, mark it as native token.

PR #407, opened 2025-04-22
feat: add native token field

temaniarpit27:
#360

### DIFF
--- a/bridgesync/downloader.go
+++ b/bridgesync/downloader.go
@@ -97,6 +97,11 @@ func buildBridgeEventHandler(contract *polygonzkevmbridgev2.Polygonzkevmbridgev2
 			return fmt.Errorf("error parsing BridgeEvent log %+v: %w", l, err)
 		}
 
+		gasTokenAddress, err := contract.GasTokenAddress(nil)
+		if err != nil {
+			return fmt.Errorf("error parsing gas token address: %w", err)
+		}
+
 		foundCall, err := extractCall(client, bridgeAddr, l.TxHash)
 		if err != nil {
 			return fmt.Errorf("failed to extract bridge event calldata (tx hash: %s): %w", l.TxHash, err)
@@ -117,6 +122,7 @@ func buildBridgeEventHandler(contract *polygonzkevmbridgev2.Polygonzkevmbridgev2
 			Amount:             bridgeEvent.Amount,
 			Metadata:           bridgeEvent.Metadata,
 			DepositCount:       bridgeEvent.DepositCount,
+			GasTokenAddress:    gasTokenAddress,
 		}})
 		return nil
 	}
--- a/bridgesync/downloader.go
+++ b/bridgesync/downloader.go
@@ -42,6 +42,7 @@ var (
 	claimMessageEtrogMethodID    = common.Hex2Bytes("f5efcd79")
 	claimAssetPreEtrogMethodID   = common.Hex2Bytes("2cffd02e")
 	claimMessagePreEtrogMethodID = common.Hex2Bytes("2d2c9d94")
+	zeroAddress                  = common.HexToAddress("0x0")
 )
 
 const (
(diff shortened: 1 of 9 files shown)


bridgesync/migrations/bridgesync0002.sql
```diff
@@ -27,5 +28,6 @@ ALTER TABLE claim ADD COLUMN tx_hash VARCHAR;
 ALTER TABLE bridge ADD COLUMN block_timestamp INTEGER;
 ALTER TABLE claim ADD COLUMN block_timestamp INTEGER;
 ALTER TABLE bridge ADD COLUMN from_address VARCHAR;
+ALTER TABLE bridge ADD COLUMN gas_token_address VARCHAR;
```
rachit77, line 31: Also update the Migration tests to account for the new column.

rachit77 (approved):
LGTM 💯

bridgesync/processor.go
```diff
@@ -506,8 +508,9 @@ func (p *processor) GetBridgesPaged(
 	bridgeResponsePtrs := make([]*BridgeResponse, len(bridgePtrs))
 	for i, bridgePtr := range bridgePtrs {
 		bridgeResponsePtrs[i] = &BridgeResponse{
-			Bridge:     *bridgePtr,
-			BridgeHash: bridgePtr.Hash(),
+			Bridge:      *bridgePtr,
+			BridgeHash:  bridgePtr.Hash(),
+			NativeToken: bridgePtr.GasTokenAddress == bridgePtr.OriginAddress,
 		}
```
Stefan-Ethernal, line 514: We should check if the `OriginAddress` is `0x0`, in that case it is also a native token bridge transfer.
As a side note, consider extracting the `common.HexToAddress("0x0")` to some global var.

```suggestion
		isNativeToken := bridgePtr.OriginAddress == bridgePtr.GasTokenAddress ||
			bridgePtr.OriginAddress == common.HexToAddress("0x0")

		bridgeResponsePtrs[i] = &BridgeResponse{
			Bridge:      *bridgePtr,
			BridgeHash:  bridgePtr.Hash(),
			NativeToken: isNativeToken,
		}
```

bridgesync/processor.go
```diff
@@ -72,6 +72,7 @@ type Bridge struct {
 	Amount             *big.Int       `meddler:"amount,bigint" json:"amount"`
 	Metadata           []byte         `meddler:"metadata" json:"metadata"`
 	DepositCount       uint32         `meddler:"deposit_count" json:"deposit_count"`
+	GasTokenAddress    common.Address `meddler:"gas_token_address" json:"gas_token_address"`
```
Stefan-Ethernal, line 75: When disucssing should we persist the native token indicator in the database, I wasn't thinking to put the `GasTokenAddress` value, but `IsNativeToken bool`. I think that it is not necessary to persist the GasTokenAddress, as this should be queried only once (on aggkit startup) and we can probably preserve it in memory, e.g. `processor` (as it is defined through an [initializer](https://github.com/agglayer/agg-contracts-internal/blob/b5f9e96eb556c51b80a5799849128eeb81eec6c4/contracts/v2/PolygonZkEVMBridgeV2.sol#L183) and should not be changed).

WDYT?

bridgesync/downloader.go
```diff
@@ -102,6 +108,13 @@ func buildBridgeEventHandler(contract *polygonzkevmbridgev2.Polygonzkevmbridgev2
 			return fmt.Errorf("failed to extract bridge event calldata (tx hash: %s): %w", l.TxHash, err)
 		}
 
+		isNativeToken := false
+
+		if gasTokenAddress == bridgeEvent.OriginAddress ||
+			bridgeEvent.OriginAddress == common.Address(aggkitCommon.ZeroHash.Bytes()) {
```
Stefan-Ethernal, line 114: Although it works this way as well, having the variable of type `common.Address` that corresponds to zero address would be nicer IMO (from a semantical point of view it is incorrect).

bridgesync/downloader.go
```diff
@@ -102,6 +108,13 @@ func buildBridgeEventHandler(contract *polygonzkevmbridgev2.Polygonzkevmbridgev2
 			return fmt.Errorf("failed to extract bridge event calldata (tx hash: %s): %w", l.TxHash, err)
 		}
 
+		isNativeToken := false
```
Stefan-Ethernal, line 111: Nitpick, you could inline this and spare a couple of lines if you want.

```suggestion
		isNativeToken := bridgeEvent.OriginAddress == gasTokenAddress ||
			bridgeEvent.OriginAddress == common.Address(aggkitCommon.ZeroHash.Bytes())
```

Stefan-Ethernal (approved):
:100: LGTM, leaving a couple of comments to consider.

PR #411, opened 2025-04-23
feat: certificate status checker

goran-ethernal:
## Description

This PR moves everything in regards to certificate status checking to a separate interface called `CertificateStatusChecker`, to improve the code quality and to have a more cleaner code separation and easier testing.

Fixes # (issue)


aggsender/aggsender.go
```diff
@@ -157,52 +157,18 @@ func (a *AggSender) checkDBCompatibility(ctx context.Context) {
...
-	}
+	checkCertChannel := a.certStatusChecker.StartStatusChecking(ctx, a.cfg.CheckStatusCertificateInterval.Duration)
 
 	chEpoch := a.epochNotifier.Subscribe("aggsender")
 	a.status.Status = types.StatusCertificateStage
 	iteration := 0
 	for {
 		select {
-		case <-checkCertChannel:
+		case checkResult := <-checkCertChannel:
```
joanestebanr, line 169: doing in that way you can have 2 checks at the same time when is called synchronously when reach an epoch. The previous version avoid this situation

aggsender/statuschecker/cert_status_checker.go
```diff
@@ -0,0 +1,294 @@
...
+		}
+
+		c.log.Debugf("agglayerClient.GetCertificateHeader status [%s] of certificate %s  elapsed time:%s",
+			certificateHeader.Status,
+			certificateHeader.ID(),
+			certificateLocal.ElapsedTimeSinceCreation())
+		appearsNewInErrorCert = appearsNewInErrorCert ||
+			(!certificateLocal.Status.IsInError() && certificateHeader.Status.IsInError())
+
+		if err := c.updateCertificateStatus(ctx, certificateLocal, certificateHeader); err != nil {
```
joanestebanr, line 169: there is no db transaction, so it can be modified by other gorutine

joanestebanr (commented):
From my point of view it's a good idea to split the code, but I would no generate a goroutine to do the check. I would left the integration in main `for` as a ticker. 

There also a launch of a gorutine internally meanwhile the rest of goroutines are launched by run.go

aggsender/aggsender.go
goran-ethernal: Reverted to the old logic: https://github.com/agglayer/aggkit/pull/411/commits/85378045dd2c58c730b9f8f0e14845c8957f3bef

aggsender/statuschecker/cert_status_checker.go
goran-ethernal: Reverted to the old logic: https://github.com/agglayer/aggkit/pull/411/commits/85378045dd2c58c730b9f8f0e14845c8957f3bef

goran-ethernal:
> From my point of view it's a good idea to split the code, but I would no generate a goroutine to do the check. I would left the integration in main `for` as a ticker.
> 
> There also a launch of a gorutine internally meanwhile the rest of goroutines are launched by run.go

Reverted to the old logic. There is no more added routines. Now it behaves syncronously.

Stefan-Ethernal (approved):
LGTM

joanestebanr (approved):

vcastellm (approved):
LGTM

PR #413, opened 2025-04-23
feat: enable `pprof`

goran-ethernal:
## Description

This PR enables `pprof` profiling on `aggkit` app so that we can run different profiles for easier finding of leaks, weak points, and non-optimized parts of the application.

## Config

A new config is added to the `aggkit`:

```
[Profiling]
ProfilingHost = "localhost"
ProfilingPort = 6060
ProfilingEnabled = true
```

If `ProfilingEnabled` is true, the `pprof` server will start on designated port in the `aggkit` app.

**IMPORTANT NOTE**: The code is taken from `cdk-validium-node`.

Fixes # (issue)

### DIFF
--- a/config/default.go
+++ b/config/default.go
@@ -253,4 +253,9 @@ AggchainProofURL = "{{AggchainProofURL}}"
 SovereignRollupAddr = "{{L1Config.polygonZkEVMAddress}}"
 GlobalExitRootL2 = "{{L2Config.GlobalExitRootAddr}}"
 GenerateAggchainProofTimeout="{{GenerateAggchainProofTimeout}}"
+
+[Profiling]
+ProfilingHost = "localhost"
+ProfilingPort = 6060
+ProfilingEnabled = true
 `
--- a/config/default.go
+++ b/config/default.go
@@ -257,5 +257,5 @@ GenerateAggchainProofTimeout="{{GenerateAggchainProofTimeout}}"
 [Profiling]
 ProfilingHost = "localhost"
 ProfilingPort = 6060
-ProfilingEnabled = true
+ProfilingEnabled = false
 `
--- a/pprof/pprof.go
+++ b/pprof/pprof.go
@@ -1,6 +1,7 @@
 package pprof
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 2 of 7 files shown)


joanestebanr (commented):
can you explain howto use  this profiling? and what data would provide?

goran-ethernal:
> can you explain howto use this profiling? and what data would provide?

`pprof` is a profiling tool that helps to analyze the performance of a Go application. Things like CPU usage, memory allocation, goroutine activity, and more.

https://pkg.go.dev/net/http/pprof

We basically expose a server on some port that runs in our `aggkit` app, and then we can just ping its endpoints to get reports on before mentioned analysis. Its super helpful to profile Go applications.

config/default.go
```diff
@@ -253,4 +253,9 @@ AggchainProofURL = "{{AggchainProofURL}}"
 SovereignRollupAddr = "{{L1Config.polygonZkEVMAddress}}"
 GlobalExitRootL2 = "{{L2Config.GlobalExitRootAddr}}"
 GenerateAggchainProofTimeout="{{GenerateAggchainProofTimeout}}"
+
+[Profiling]
+ProfilingHost = "localhost"
+ProfilingPort = 6060
+ProfilingEnabled = true
```
joanestebanr, line 260: maybe the default must be disabled, WDYT?

joanestebanr: Also I suggest to add new config params to [this test](https://github.com/agglayer/aggkit/blob/336ffecdbacc05953fc998aa8364d726a92c685e/config/config_test.go#L22) to be sure that read the expected params.

goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/413/commits/619cbaae252649cf4a35cf39bf2333a6fa8ebc51

pprof/pprof.go
```diff
@@ -0,0 +1,55 @@
...
+//   - Handles server closure gracefully, logging a warning if the server is stopped
+//     or an error if the connection is closed unexpectedly.
+func StartProfilingHTTPServer(c Config) {
+	const two = 2
+	mux := http.NewServeMux()
+	address := fmt.Sprintf("%s:%d", c.ProfilingHost, c.ProfilingPort)
+	lis, err := net.Listen("tcp", address)
+	if err != nil {
+		log.Errorf("failed to create tcp listener for profiling: %v", err)
+		return
```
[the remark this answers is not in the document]
goran-ethernal: To me it shouldn't. Its just a debug thing. Same as for metrics.

pprof/pprof.go
```diff
@@ -0,0 +1,55 @@
...
+	mux.HandleFunc(prometheus.ProfilingCmdEndpoint, pprof.Cmdline)
+	mux.HandleFunc(prometheus.ProfilingSymbolEndpoint, pprof.Symbol)
+	mux.HandleFunc(prometheus.ProfilingTraceEndpoint, pprof.Trace)
+	profilingServer := &http.Server{
+		Handler:           mux,
+		ReadHeaderTimeout: two * time.Minute,
+		ReadTimeout:       two * time.Minute,
+	}
+	log.Infof("profiling server listening on port %d", c.ProfilingPort)
+	if err := profilingServer.Serve(lis); err != nil {
```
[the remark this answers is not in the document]
goran-ethernal: Nice one 💯 

Fix: https://github.com/agglayer/aggkit/pull/413/commits/c1f173772a8d9954c7fb0af17dc1036caa6be7d1

Stefan-Ethernal (approved):
:100:

joanestebanr (approved):

PR #421, opened 2025-04-25
fix: `PP` to `FEP` upgrade `startBlock`

goran-ethernal:
## Description

This PR fixes the case where in `PP` to `FEP` upgrade we have a gap between the last settled block in `PP` and the start block for `FEP` that is defined in the `FEP` rollup contract (this start block defines a block from which spans are created).

Fixes # (issue)


Stefan-Ethernal (approved):

rachit77 (approved):

vcastellm (approved):
LGTM

PR #422, opened 2025-04-25
chore: remove `GetBridgesPublished` and use `GetBridges` instead

### LINKED ISSUE #423: chore: remove `GetBridgesPublished` from bridge syncer and fix mockery config

Stefan-Ethernal:
## Description

The `GetBridgesPublished` is just a wrapper around the `GetBridges` in the bridge syncer, and it is useless. Therefore, this PR removes the `GetBridgesPublished` and uses the `GetBridges` directly instead.

Also fixes the mockery configuration, since there were some mocks that aren't used and others that were removed in a meantime (e.g. `evmDownloaderFull`).

Fixes #423


goran-ethernal (approved):
LGTM 💯

PR #424, opened 2025-04-25
feat: remove check end-epoch and fix kurtosis version

### LINKED ISSUE #420: feat: a certificate send at end of an epoch can skip to next epoch
# Feature Request
Previous verion of agglayer discourage `aggsender` to send a certificate in the end of an epoch. For this there was a code to prevent this. 

Right now, last version, doesn't have this limit so we are going to remove this checks from aggsender. Also we have to take in consideration to skipping epoch

To be more precise here, we have some code on the aggsender that if we approach an end of the epoch, we won't send the certificate. Is this check safe to remove from the aggsender? What will happen if we send a certificate near the end of the epoch? I remember we had some limitation on the old flow in agglayer. 
[@Monir](https://0xpolygon.slack.com/team/U075MKM0SEL)
 
[@Simon](https://0xpolygon.slack.com/team/U074X143J2F)
yeah, last summer the limitation was the fact that we were constrained with one PP settled per epoch (still the case today), so if we receive one certificate near the end of epoch N, then high chances were that the process of this certificate would slip into epoch N+1 and so being considered for epoch N+1. So the next certificate received in N+1 would be considered for N+2, leading to some certificates which could have spike in latency that we tried to avoid at that time
Now I think in this new FEP (op-stack) flow we might not need to do this on the aggsender even if the certificate does not settle on the agglayer because it is in between epochs, we will just resend the same certificate in the new epoch.
Yeah it seems that trying to make some fine-tuned timing isn't perfect either, trying to optimize the timing further for our current setup might not make sense especially if we move away from the 1cert/epoch/network for the next version, so yeah I dont see problems with ditching this custom logic

joanestebanr:
## Description
- Remove check of end-of-epoch before sending certificate
- Add support for  kurtosis-core > 1.7.0
- Fix issue #434: setup L2 client with op-node support

Fixes #420

### DIFF
--- a/aggsender/aggsender.go
+++ b/aggsender/aggsender.go
@@ -299,7 +299,7 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certifi
 			rateLimitSleepTime.String(), a.rateLimiter.String())
 		time.Sleep(*rateLimitSleepTime)
 	}
-	a.log.Infof("certificate ready to be send to AggLayer: %s start: %s , end: %s",
+	a.log.Infof("certificate ready to be sent to AggLayer: %s start: %s , end: %s",
 		certificate.Brief(), startEpochStatus.String(), a.epochNotifier.GetEpochStatus().String())
 	metrics.CertificateBuildTime(time.Since(start).Seconds())
 
--- a/aggsender/aggsender.go
+++ b/aggsender/aggsender.go
@@ -299,7 +299,8 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certifi
 			rateLimitSleepTime.String(), a.rateLimiter.String())
 		time.Sleep(*rateLimitSleepTime)
 	}
-	a.log.Infof("certificate ready to be send to AggLayer: %s start: %s , end: %s", certificate.Brief(), startEpochStatus.String(), a.epochNotifier.GetEpochStatus().String())
+	a.log.Infof("certificate ready to be send to AggLayer: %s start: %s , end: %s",
+		certificate.Brief(), startEpochStatus.String(), a.epochNotifier.GetEpochStatus().String())
 	metrics.CertificateBuildTime(time.Since(start).Seconds())
 
 	if a.cfg.DryRun {
--- a/aggsender/aggsender.go
+++ b/aggsender/aggsender.go
@@ -274,7 +274,8 @@ func (a *AggSender) sendCertificates(ctx context.Context, returnAfterNIterations
 
 // sendCertificate sends certificate for a network
 func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certificate, error) {
-	a.log.Infof("trying to send a new certificate...")
+	startEpochStatus := a.epochNotifier.GetEpochStatus()
+	a.log.Infof("trying to send a new certificate... %s", startEpochStatus.String())
 
 	start := time.Now()
 
@@ -298,11 +299,7 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certifi
 			rateLimitSleepTime.String(), a.rateLimiter.String())
 		time.Sleep(*rateLimitSleepTime)
 	}
-	if !a.isAllowedSendCertificateEpochPercent() {
-		return nil, fmt.Errorf("forbidden to send certificate due epoch percentage")
-	}
-
-	a.log.Infof("certificate ready to be send to AggLayer: %s", certificate.Brief())
+	a.log.Infof("certificate ready to be send to AggLayer: %s start: %s , end: %s", certificate.Brief(), startEpochStatus.String(), a.epochNotifier.GetEpochStatus().String())
 	metrics.CertificateBuildTime(time.Since(start).Seconds())
 
 	if a.cfg.DryRun {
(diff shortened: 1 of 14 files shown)


config/default.go
```diff
@@ -234,8 +234,6 @@ Mode = "PessimisticProof"
 CheckStatusCertificateInterval = "5m"
 RetryCertAfterInError = false
 GlobalExitRootL2="{{L2Config.GlobalExitRootAddr}}"
-# Don't send certificate over 80% of the epoch
-MaxEpochPercentageAllowedToSendCertificate=80
```
Stefan-Ethernal, line 238: Remove this param from the [aggsender.md](https://github.com/agglayer/aggkit/blob/8e28051fa30de5952a3aa6a88b7818d6e9c8ccf4/docs/aggsender.md#L171) as well.

aggsender/types/epoch_notifier.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 )
 
+const perCent = 100.0
```
Stefan-Ethernal, line 8: Consider keeping reusing this constant, instead of `maxPercent` (https://github.com/agglayer/aggkit/blob/8e28051fa30de5952a3aa6a88b7818d6e9c8ccf4/aggsender/epoch_notifier_per_block.go#L13). WDYT?

aggsender/types/epoch_notifier_test.go
```diff
@@ -0,0 +1,16 @@
+package types
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestXxx(t *testing.T) {
```
Stefan-Ethernal, line 9: WDYT?
```suggestion
func TestEpochStatus_String(t *testing.T) {
```

aggsender/epoch_notifier_per_block.go
```diff
@@ -169,7 +169,7 @@ func (e *EpochNotifierPerBlock) step(status internalStatus,
 	if needNotify {
 		logFunc = e.logger.Infof
 	}
-	logFunc("New block seen [finality:%s]: %d. blockRate:%s Epoch:%d Percent:%f%% notify:%v config:%s",
+	logFunc("New block seen [finality:%s]: %d. blockRate:%s Epoch:%d Percent:%.3f%% notify:%v config:%s",
```
Stefan-Ethernal, line 172: Two decimals should be more than enough, wdyt?
```suggestion
	logFunc("New block seen [finality:%s]: %d. blockRate:%s Epoch:%d Percent:%.2f%% notify:%v config:%s",
```

aggsender/types/epoch_notifier.go
joanestebanr: I'll keep `perCent` const because the packages are not the same

Stefan-Ethernal (approved):
:100: LGTM

test/bats/helpers/common-multi_cdk-setup.bash
```diff
@@ -6,10 +6,10 @@ _common_multi_setup() {
...
     readonly target_address=0xbecE3a31343c6019CDE0D5a4dF2AF8Df17ebcB0f
     readonly target_private_key=0x51caa196504216b1730280feb63ddd8c5ae194d13e57e58d559f1f1dc3eda7c9
-
-    kurtosis service exec $enclave contracts-001 "cat /opt/zkevm/combined-001.json" | tail -n +2 | jq '.' >combined-001.json
-    kurtosis service exec $enclave contracts-002 "cat /opt/zkevm/combined-002.json" | tail -n +2 | jq '.' >combined-002.json
-
+    
+    kurtosis_download_file_exec_method $enclave "contracts-001" "/opt/zkevm/combined-001.json" | jq '.' >combined-001.json
+    kurtosis_download_file_exec_method $enclave "contracts-002" "/opt/zkevm/combined-001.json" | jq '.' >combined-002.json
+
```
Stefan-Ethernal: Nitpick, some extra spaces, that can be removed
```suggestion

    kurtosis_download_file_exec_method $enclave "contracts-001" "/opt/zkevm/combined-001.json" | jq '.' >combined-001.json
    kurtosis_download_file_exec_method $enclave "contracts-002" "/opt/zkevm/combined-001.json" | jq '.' >combined-002.json

```

aggsender/aggsender.go
```diff
@@ -298,11 +299,8 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certifi
 			rateLimitSleepTime.String(), a.rateLimiter.String())
 		time.Sleep(*rateLimitSleepTime)
 	}
-	if !a.isAllowedSendCertificateEpochPercent() {
-		return nil, fmt.Errorf("forbidden to send certificate due epoch percentage")
-	}
-
-	a.log.Infof("certificate ready to be send to AggLayer: %s", certificate.Brief())
+	a.log.Infof("certificate ready to be send to AggLayer: %s start: %s , end: %s",
```
rachit77, line 302:
```suggestion
	a.log.Infof("certificate ready to be sent to AggLayer: %s start: %s , end: %s",
```

rachit77:
```suggestion
	a.log.Infof("certificate ready to be sent to AggLayer: %s start: %s , end: %s",
```

rachit77 (approved):
💯

PR #440, opened 2025-05-02
feat: add sanity check for initial L2 block on FEP

### LINKED ISSUE #436: feat: Sanity check skip block range in case of upgrade from PP to FEP
Sanity check that upgrading from PP to FEP there are no block gap:
- Setting the L2 block num when upgrading from PP to FEP is  a manual process, and while it should be reviewed heavily, there is no guarantee that this will be done right
- We will add a sanity check on the aggsender, that will trigger a fatal error if there is gaps: `last verified block while on PP +1 != first block that needs to be verified on FEP`
- We don't have any mechanism to recover from such an error

joanestebanr:
## Description

- A gap is not permitted (reverted PR #421)
- Add a sanity check to verify that on start of aggsender

Fixes #436


aggsender/types/types.go
```diff
@@ -29,6 +29,8 @@ const (
 // AggsenderFlow is an interface that defines the methods to manage the flow of the AggSender
 // based on the different prover types
 type AggsenderFlow interface {
+	// CheckInitialStatus checks the initial status for the flow it's ok
+	CheckInitialStatus(ctx context.Context) error
```
goran-ethernal, line 33: I would call it `CheckStartingBlock`:

```suggestion
	CheckStartingBlock(ctx context.Context) error
```

aggsender/aggsender.go
```diff
@@ -190,21 +190,30 @@ func (a *AggSender) checkInitialStatus(ctx context.Context) {
 	ticker := time.NewTicker(a.cfg.DelayBeetweenRetries.Duration)
 	defer ticker.Stop()
 	a.status.Status = types.StatusCheckingInitialStage
-	for {
+	for firstRun := true; ; firstRun = false {
```
goran-ethernal: I wouldn't complicate this loop, and would just move this check after the `checkInitialStatus` is finished, since this is checking the `db` and `agglayer` state, and after that, it is ok to call the check for start block on the flow.

joanestebanr: I think that you're right!, good suggestion!

aggsender/types/types.go
joanestebanr: I'll keep the name, because the idea is to check the initial Status of the flow, for aggchain is starting block, for PP is nothing... and who knowns if there are going to be another flow that check another thing... WDYT?

aggsender/aggsender.go
```diff
@@ -173,6 +173,9 @@ func (a *AggSender) Start(ctx context.Context) {
 
 	a.checkDBCompatibility(ctx)
 	a.checkInitialStatus(ctx)
+	if err := a.flow.CheckInitialStatus(ctx); err != nil {
+		a.log.Panicf("error checking flow Initial Status: %v, retrying in %s", err, a.cfg.DelayBeetweenRetries.String())
```
goran-ethernal: There is no retrying so, just remove it:

```suggestion
		a.log.Panicf("error checking flow Initial Status: %w, err)
```

aggsender/aggsender.go
```diff
@@ -181,10 +184,11 @@ func (a *AggSender) checkDBCompatibility(ctx context.Context) {
 		return
 	}
 	if err := a.compatibilityStoragedChecker.Check(ctx, nil); err != nil {
-		a.log.Fatalf("error checking compatibility data in DB, you can bypass this check using config file. Err: %w", err)
+		a.log.Panicf("error checking compatibility data in DB, you can bypass this check using config file. Err: %w", err)
 	}
 }
 
+// checkInitialStatus check local status vs agglayer status
```
goran-ethernal: Doubled comment, just remove one line.

aggsender/aggsender.go
```diff
@@ -202,6 +206,7 @@ func (a *AggSender) checkInitialStatus(ctx context.Context) {
 		}
 		select {
 		case <-ctx.Done():
+			log.Fatalf("checkInitialStatus: context Done!") //nolint:gocritic
```
goran-ethernal: Should we use `Panicf` here, like in the other cases?

Stefan-Ethernal: It is not obvious to me, why we would we even log this on fatal level? Isn't it enough just to return to the caller (I suspect this is the use case, where interrupt signal was caught, so process is going to be terminated anyways)...?

aggsender/flows/flow_pp.go
```diff
@@ -42,6 +42,12 @@ func NewPPFlow(log types.Logger,
 	}
 }
 
+// CheckInitialStatus checks that initial status it's ok.
```
Stefan-Ethernal:
```suggestion
// CheckInitialStatus checks that initial status is correct.
```

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -76,6 +80,26 @@ func NewAggchainProverFlow(log types.Logger,
 	}, nil
 }
 
+// CheckInitialStatus checks that initial status it's ok.
```
Stefan-Ethernal:
```suggestion
// CheckInitialStatus checks that initial status is correct.
```

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -51,16 +55,16 @@ func NewAggchainProverFlow(log types.Logger,
...
-		return nil, fmt.Errorf("aggchainProverFlow - error creating L2Etherman: %w", err)
+		return nil, fmt.Errorf("aggchainProverFlow - error creating EVMChainGERReader: %w", err)
 	}
 
 	startL2Block, err := getL2StartBlock(sovereignRollupAddr, l1Client)
 	if err != nil {
 		return nil, fmt.Errorf("aggchainProverFlow - error reading sovereign rollup: %w", err)
 	}
-
+	log.Infof("aggchainProverFlow -read from severeignRollup(L1) starting L2 block: %d", startL2Block)
```
Stefan-Ethernal:
```suggestion
	log.Infof("aggchainProverFlow - read from severeignRollup (L1) starting L2 block: %d", startL2Block)
```

Stefan-Ethernal (approved):
💯 LGTM, leaving a couple of nitpicks to consider

aggsender/aggsender.go
joanestebanr: It needs to interrupt the execution because is a infinite loop that only must return if it's ok. 
I change the other `Fatalf` for `Panic` because I need to be able to  unittest, anyway if you think that is better a Panic I would change it

goran-ethernal (approved):

goran-ethernal (approved):
LGTM 💯

PR #450, opened 2025-05-06
fix: Claimsponsor

rachit77:
## Description

Fix claimsponsor when transaction is already claimed

Aggkit CI Test: https://github.com/agglayer/aggkit/pull/488
Kurtosis PR: https://github.com/0xPolygon/kurtosis-cdk/pull/628
E2E PR: https://github.com/agglayer/e2e/pull/50


claimsponsor/claimsponsor.go
```diff
@@ -135,13 +137,32 @@ func (c *ClaimSponsor) claim(ctx context.Context) error {
...
 			}
-			return fmt.Errorf("error calling getClaim with globalIndex %s: %w", claim.GlobalIndex.String(), err)
+			return fmt.Errorf("getFirstPendingClaim failed: %w", err)
 		}
-		txID, err := c.sender.sendClaim(ctx, claim)
+
+		claim.TxID, err = c.sender.sendClaim(ctx, claim)
 		if err != nil {
+			// AlreadyClaimed(): execution reverted (0x646cf558)
+			if strings.Contains(err.Error(), "execution reverted (0x646cf558)") {
```
temaniarpit27: can we make it const?

temaniarpit27 (approved):
💯

claimsponsor/claimsponsor.go
Stefan-Ethernal: IMO, it would be better if we have computed the Keccak256 hash of the `AlreadyClaimed()` error and take the first 4 bytes and capture it in a variable, which then can be used in string comparison, wdyt?

```go
var alreadyClaimedRevertCode = crypto.Keccak256("AlreadyClaimed()")[:4]
```

claimsponsor/evmclaimsponsor.go
```diff
@@ -21,9 +22,10 @@ const (
 	LeafTypeAsset uint8 = 0
 	// LeafTypeMessage represents a bridge message
 	LeafTypeMessage uint8 = 1
+)
 
-	gasTooHighErrTemplate = "Claim tx estimated to consume more gas than the maximum allowed by the service. " +
-		"Estimated %d, maximum allowed: %d"
+var ErrGasEstimateTooHigh = errors.New(
+	"claim gas estimate exceeds maximum allowed by claimsponsor service",
```
Stefan-Ethernal:
```suggestion
	"claim gas estimate exceeds maximum gas allowed by claim sponsor service",
```

Stefan-Ethernal (approved):
:100: LGTM, leaving a couple of comments to consider

claimsponsor/claimsponsor.go
rachit77: as discussed extracted the revert code to a const

PR #452, opened 2025-05-07
fix: recognize no proofs built yet error from the aggkit prover

goran-ethernal:
## Description

This PR enables easy recognition of specific errors from GRPC communication so we can handle them properly.

In this example, we are recognizing the error returned from the `aggkit prover` when there are no span proofs built for a requested range of L2 blocks. This is a normal use case which should not be reported as an error on `aggkit` logs, but an info.

Fixes # (issue)


Stefan-Ethernal (approved):

vcastellm (approved):

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -170,6 +177,12 @@ func (a *AggchainProverFlow) verifyBuildParamsAndGenerateProof(
 	aggchainProof, rootFromWhichToProveClaims, err := a.GenerateAggchainProof(
 		ctx, lastProvenBlock, buildParams.ToBlock, buildParams.Claims)
 	if err != nil {
+		if errors.As(err, errNoProofBuiltYet) {
+			a.log.Infof("aggchainProverFlow - no proof built yet for lastProvenBlock: %d, maxEndBlock: %d",
+				lastProvenBlock, buildParams.ToBlock)
+			return nil, nil
+		}
+
 		return nil, fmt.Errorf("aggchainProverFlow - error generating aggchain proof: %w", err)
```
joanestebanr, line 186: maybe make sense to also add to this error the requested data: `lastProvenBlock: %d, maxEndBlock: %d` as the log.Infof

joanestebanr (approved):
This PR need to be cherry-picked to release/0.3?

aggsender/flows/flow_aggchain_prover.go
goran-ethernal: It is part of the wrapped error that gets returned by `GenerateAggchainProof` and wrapped here.

goran-ethernal:
> This PR need to be cherry-picked to release/0.3?

Yes. Eventually. It is not really a must have.

PR #454, opened 2025-05-09
fix: optimize check for pending certificates

goran-ethernal:
## Description

During profiling of `aggsender` run against the new `FEP` (`op succinct`) stack, we noticed that a significant portion of `CPU` time was used to read the full certificates from the `aggsender db` while checking for changes in the certificate status.

Since `aggsender` only needs to know couple of things when checking for changes in status of pending certificates (like `certificate_id`, `height`, `retry_count`, `status`), this PR enables fetching full certificate by status, or a light version of it from the `aggsender db`, so we do not load all the data. Most of the time was spent loading `aggchain proof` data on the certificate from `db`, since proofs are quite large and heavy, and they are not needed when checking for changes in the certificate status.

### `CertificateInfo` split

This PR splits the `CertificateInfo` struct that is used in `aggsender` into two structs:

```go
type CertificateHeader struct {
	Height                  uint64                          `meddler:"height"`
	RetryCount              int                             `meddler:"retry_count"`
	CertificateID           common.Hash                     `meddler:"certificate_id,hash"`
	PreviousLocalExitRoot   *common.Hash                    `meddler:"previous_local_exit_root,hash"`
	NewLocalExitRoot        common.Hash                     `meddler:"new_local_exit_root,hash"`
	FromBlock               uint64                          `meddler:"from_block"`
	ToBlock                 uint64                          `meddler:"to_block"`
	Status                  agglayertypes.CertificateStatus `meddler:"status"`
	CreatedAt               uint32                          `meddler:"created_at"`
	UpdatedAt               uint32                          `meddler:"updated_at"`
	FinalizedL1InfoTreeRoot *common.Hash                    `meddler:"finalized_l1_info_tree_root,hash"`
	L1InfoTreeLeafCount     uint32                          `meddler:"l1_info_tree_leaf_count"`
}

type Certificate struct {
	Header            *CertificateHeader
	SignedCertificate *string        `meddler:"signed_certificate"`
	AggchainProof     *AggchainProof `meddler:"aggchain_proof,aggchainproof"`
}
```

Since most of the time, `aggsender` only needs the data from `CertificateHeader`, this split is used to optimize data querying and loading from `aggsender db`, since `aggchain proof` can be quite large, and it is not optimal to load it every time, only in the cases when we need to resend it to the `agglayer`.

### DB changes

`AggSenderStorage` now contains functions to get the `CertificateHeader` instead of the full certificate that contains the `aggchain proof` as well. This way we are only loading simple data from the certificate and optimizing data loading.

Another function was added, called: `GetCertificateAggchainProof(height uint64, certificateID common.Hash) (*types.AggchainProof, error)`, that is called to load `aggchain proof` from the `db` in the cases when we actually need to load it (when resending the `InError` certificate).

Fixes # (issue)

### DIFF
--- a/aggsender/db/aggsender_db_storage.go
+++ b/aggsender/db/aggsender_db_storage.go
@@ -97,7 +97,7 @@ func NewAggSenderSQLStorage(logger *log.Logger, cfg AggSenderSQLStorageConfig) (
 // GetCertificateHeadersByStatus returns a list of certificate headers by their status
 func (a *AggSenderSQLStorage) GetCertificateHeadersByStatus(
 	statuses []agglayertypes.CertificateStatus) ([]*types.CertificateHeader, error) {
-	query := "SELECT * FROM certificate_info"
+	query := selectQueryCertificateHeader
 
 	args := make([]any, len(statuses))
 
@@ -142,7 +142,7 @@ func (a *AggSenderSQLStorage) GetCertificateByHeight(height uint64) (*types.Cert
 func (a *AggSenderSQLStorage) GetCertificateHeaderByHeight(height uint64) (*types.CertificateHeader, error) {
 	var certificateHeader types.CertificateHeader
 	if err := meddler.QueryRow(a.db, &certificateHeader,
-		"SELECT * FROM certificate_info WHERE height = $1;", height); err != nil {
+		fmt.Sprintf("%s WHERE height = $1;", selectQueryCertificateHeader), height); err != nil {
 		return nil, getSelectQueryError(height, err)
 	}
 	return &certificateHeader, nil
@@ -175,7 +175,7 @@ func (a *AggSenderSQLStorage) GetLastSentCertificate() (*types.Certificate, erro
 func (a *AggSenderSQLStorage) GetLastSentCertificateHeader() (*types.CertificateHeader, error) {
 	var certificateHeader types.CertificateHeader
 	if err := meddler.QueryRow(a.db, &certificateHeader,
-		"SELECT * FROM certificate_info ORDER BY height DESC LIMIT 1;"); err != nil {
+		fmt.Sprintf("%s ORDER BY height DESC LIMIT 1;", selectQueryCertificateHeader)); err != nil {
 		return nil, getSelectQueryError(0, err)
 	}
 	return &certificateHeader, nil
--- a/aggsender/db/aggsender_db_storage.go
+++ b/aggsender/db/aggsender_db_storage.go
@@ -45,9 +45,9 @@ type AggSenderStorage interface {
 	// DeleteCertificate deletes a certificate from the storage
 	DeleteCertificate(ctx context.Context, certificateID common.Hash) error
 	// GetCertificatesByStatus returns a list of certificates by their status
-	GetCertificatesByStatus(status []agglayertypes.CertificateStatus) ([]*types.CertificateInfo, error)
-	// UpdateCertificate updates certificate in db
-	UpdateCertificate(ctx context.Context, certificate types.CertificateInfo) error
+	GetCertificatesByStatus(status []agglayertypes.CertificateStatus, lightQuery bool) ([]*types.CertificateInfo, error)
+	// UpdateCertificateStatus updates certificate status in db
+	UpdateCertificateStatus(ctx context.Context, certificate types.CertificateInfo) error
 }
 
 var _ AggSenderStorage = (*AggSenderSQLStorage)(nil)
@@ -84,10 +84,16 @@ func NewAggSenderSQLStorage(logger *log.Logger, cfg AggSenderSQLStorageConfig) (
 	}, nil
 }
 
+// GetCertificatesByStatus returns a list of certificates by their status
 func (a *AggSenderSQLStorage) GetCertificatesByStatus(
-	statuses []agglayertypes.CertificateStatus) ([]*types.CertificateInfo, error) {
+	statuses []agglayertypes.CertificateStatus, lightQuery bool) ([]*types.CertificateInfo, error) {
 	query := "SELECT * FROM certificate_info"
-	args := make([]interface{}, len(statuses))
+	if lightQuery {
+		query = "SELECT height, retry_count, certificate_id, status, created_at, updated_at, " +
+			"from_block, to_block FROM certificate_info"
+	}
+
+	args := make([]any, len(statuses))
 
 	if len(statuses) > 0 {
 		placeholders := make([]string, len(statuses))
@@ -233,8 +239,8 @@ func deleteCertificate(tx db.Querier, certificateID common.Hash) error {
 	return nil
 }
 
-// UpdateCertificate updates a certificate
-func (a *AggSenderSQLStorage) UpdateCertificate(ctx context.Context, certificate types.CertificateInfo) error {
+// UpdateCertificateStatus updates a certificate status in the storage
+func (a *AggSenderSQLStorage) UpdateCertificateStatus(ctx context.Context, certificate types.CertificateInfo) error {
 	tx, err := db.NewTx(ctx, a.db)
 	if err != nil {
 		return err
--- a/aggsender/db/aggsender_db_storage.go
+++ b/aggsender/db/aggsender_db_storage.go
@@ -56,8 +56,10 @@ type AggSenderStorage interface {
 	GetLastSentCertificateHeader() (*types.CertificateHeader, error)
 	// GetCertificateHeaderByHeight returns a certificate header by its height
 	GetCertificateHeaderByHeight(height uint64) (*types.CertificateHeader, error)
-	// IsLastSentCertificateInError checks if the last sent certificate is in error
-	IsLastSentCertificateInError() (bool, error)
+	// GetLastSentCertificateHeaderWithProofIfInError returns the last certificate header sent to the aggLayer
+	// and the aggchain proof if the certificate is in error
+	GetLastSentCertificateHeaderWithProofIfInError(
+		ctx context.Context) (*types.CertificateHeader, *types.AggchainProof, error)
 }
 
 var _ AggSenderStorage = (*AggSenderSQLStorage)(nil)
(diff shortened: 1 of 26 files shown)


aggsender/db/aggsender_db_storage_test.go
```diff
@@ -170,32 +197,47 @@ func Test_Storage(t *testing.T) {
...
+		require.NoError(t, err)
+		require.Len(t, certificatesFromDB, 1)
+		require.Equal(t, certificates[2].Height, certificatesFromDB[0].Height)
+		require.Equal(t, certificates[2].RetryCount, certificatesFromDB[0].RetryCount)
+		require.Equal(t, certificates[2].CertificateID, certificatesFromDB[0].CertificateID)
+		require.Equal(t, certificates[2].Status, certificatesFromDB[0].Status)
+		require.Equal(t, certificates[2].CreatedAt, certificatesFromDB[0].CreatedAt)
+		require.Equal(t, certificates[2].UpdatedAt, certificatesFromDB[0].UpdatedAt)
+		require.Equal(t, certificates[2].FromBlock, certificatesFromDB[0].FromBlock)
+		require.Equal(t, certificates[2].ToBlock, certificatesFromDB[0].ToBlock)
```
Stefan-Ethernal: Hm, why this wouldn't work? It is tedious to write all those assertions...
```suggestion
		require.Equal(t, certificates[2], certificatesFromDB[0])
```

Stefan-Ethernal (approved):
:100: LGTM

aggsender/db/aggsender_db_storage_test.go
goran-ethernal: Its because they should be the same except for the `AggchainProof` field, which should be nil in the case of the `certficatesFromDB[0]`, because we used the light query option, in which we do not load the proof from `db`.

aggsender/db/aggsender_db_storage.go
```diff
@@ -233,8 +239,8 @@ func deleteCertificate(tx db.Querier, certificateID common.Hash) error {
 	return nil
 }
 
-// UpdateCertificate updates a certificate
-func (a *AggSenderSQLStorage) UpdateCertificate(ctx context.Context, certificate types.CertificateInfo) error {
+// UpdateCertificateStatus updates a certificate status in the storage
+func (a *AggSenderSQLStorage) UpdateCertificateStatus(ctx context.Context, certificate types.CertificateInfo) error {
```
joanestebanr: I agree with the naming, but I'll take advantatge to refactor also the parameters. With this name you expect to pass the new status only instead of a full certificate

aggsender/statuschecker/cert_status_checker.go
```diff
@@ -99,7 +99,7 @@ func (c *certStatusChecker) CheckInitialStatus(
 // It returns:
 // bool -> if there are pending certificates
 func (c *certStatusChecker) CheckPendingCertificatesStatus(ctx context.Context) types.CertStatus {
-	pendingCertificates, err := c.storage.GetCertificatesByStatus(agglayertypes.NonSettledStatuses)
+	pendingCertificates, err := c.storage.GetCertificatesByStatus(agglayertypes.NonSettledStatuses, true)
```
joanestebanr: why is required there the full certitifacate?

aggsender/db/aggsender_db_storage.go
```diff
@@ -45,9 +45,9 @@ type AggSenderStorage interface {
 	// DeleteCertificate deletes a certificate from the storage
 	DeleteCertificate(ctx context.Context, certificateID common.Hash) error
 	// GetCertificatesByStatus returns a list of certificates by their status
-	GetCertificatesByStatus(status []agglayertypes.CertificateStatus) ([]*types.CertificateInfo, error)
-	// UpdateCertificate updates certificate in db
-	UpdateCertificate(ctx context.Context, certificate types.CertificateInfo) error
+	GetCertificatesByStatus(status []agglayertypes.CertificateStatus, lightQuery bool) ([]*types.CertificateInfo, error)
```
joanestebanr: It's a bit confusing the bool parameter and the expected result, because it's difficult to  foresee which fields are going to have valid data. Also, after this point the same, we don't known if the value is valid or not and we can't prevent to use this not reported fields. Maybe it will be better to split the struct `CertificateInfo` into `header`+ `data` as Agglayer do (`Certificate` and `CertificateHeader`). WDTY?

joanestebanr (commented):
I think that it's better to split the Certificate struct into header and data to reduce bad use in a long term

aggsender/db/aggsender_db_storage.go
goran-ethernal: Makes sense. I will implement this suggestion.

aggsender/statuschecker/cert_status_checker.go
goran-ethernal: Its not required. This means it will execute a light query. But this will change with `Certificate` - `CertificateHeader`.

aggsender/db/aggsender_db_storage.go
goran-ethernal: Looks like meddler is not supporting embedded types.

If we go with your suggestion, we need to choose between two things:
- if we want to use clean, simple code to query and save data, we need to use meddler. But to use meddler, we need to have both `CertificateInfo`, and this new `Certificate {Header, Data}` struct. This means, when we want to add or remove a new field, we need to update both `CertificateInfo` and `Certificate` structs (which can be difficult to maintain, since someone can forget to add it or remove it from one of the structs).
- if we want to use just new `Certificate {Header, Data}` struct, and remove `CertificateInfo`, which is easy to maintain, and less error prone, we can not use meddler, and need to manually load every field from db, which makes the `aggsender_db_storage.go` much more complex.

aggsender/db/aggsender_db_storage.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/454/commits/e650a3a65be39055eb2af897a3f67cf9a77aa7f3

aggsender/db/aggsender_db_storage.go
goran-ethernal: @joanestebanr This is going to be a big change. I would like to do it through a separate PR.

joanestebanr: I think that the change is make sense with this optimization, WDYT?

goran-ethernal: Done: https://github.com/agglayer/aggkit/pull/454/commits/0fc52665bc9d3cf2ebd64e550d5a01338c873655

aggsender/statuschecker/cert_status_checker.go
goran-ethernal: Done: https://github.com/agglayer/aggkit/pull/454/commits/0fc52665bc9d3cf2ebd64e550d5a01338c873655

aggsender/db/aggsender_db_storage.go
```diff
@@ -84,10 +94,12 @@ func NewAggSenderSQLStorage(logger *log.Logger, cfg AggSenderSQLStorageConfig) (
 	}, nil
 }
 
-func (a *AggSenderSQLStorage) GetCertificatesByStatus(
-	statuses []agglayertypes.CertificateStatus) ([]*types.CertificateInfo, error) {
+// GetCertificateHeadersByStatus returns a list of certificate headers by their status
+func (a *AggSenderSQLStorage) GetCertificateHeadersByStatus(
+	statuses []agglayertypes.CertificateStatus) ([]*types.CertificateHeader, error) {
 	query := "SELECT * FROM certificate_info"
```
joanestebanr: to improve the query make sense to avoid retrieve the unneded fields like `aggchain_proof` so make sense to do a `SELECT` with a list of fields

aggsender/db/aggsender_db_storage.go
```diff
@@ -113,8 +125,27 @@ func (a *AggSenderSQLStorage) GetCertificatesByStatus(
...
+	}
+
+	return certInfo.ToCertificate(), nil
+}
+
+// GetCertificateHeaderByHeight returns a certificate by its height
+func (a *AggSenderSQLStorage) GetCertificateHeaderByHeight(height uint64) (*types.CertificateHeader, error) {
+	var certificateHeader types.CertificateHeader
+	if err := meddler.QueryRow(a.db, &certificateHeader,
+		"SELECT * FROM certificate_info WHERE height = $1;", height); err != nil {
```
joanestebanr: same here, select only required fields

aggsender/types/types.go
```diff
@@ -79,25 +81,22 @@ func (s *SP1StarkProof) String() string {
 	)
 }
 
-type CertificateInfo struct {
```
joanestebanr, line 82: `CertificateInfo` may be make sense to be a private type for db (It looks like you use it to be able to use meddler) and also renaming to avoid confunsion to something like  `certificateInfoRow`

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -99,35 +99,40 @@ func (a *AggchainProverFlow) sanityCheckNoBlockGaps(lastSentCertificate *types.C
 // if the last sent certificate is in error, we need to resend the exact same certificate
 // also, it calls the aggchain prover to get the aggchain proof
 func (a *AggchainProverFlow) GetCertificateBuildParams(ctx context.Context) (*types.CertificateBuildParams, error) {
-	lastSentCertificateInfo, err := a.storage.GetLastSentCertificate()
+	isLastSentCertificateInError, err := a.storage.IsLastSentCertificateInError()
```
joanestebanr: Here can be a potential  db data inconsistency. 
- maybe make sense to  instead or returning a bool, returns the header
- having a second function to fill `SignedCertificate` and `AggchainProof` as you suggested initially

aggsender/db/aggsender_db_storage.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/454/commits/7994506093b374804bafed5f91b1fbcf856a86cf

aggsender/db/aggsender_db_storage.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/454/commits/7994506093b374804bafed5f91b1fbcf856a86cf

aggsender/types/types.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/454/commits/2c88da8bd0b2c14b6244401efda589149ad2dbf2

aggsender/flows/flow_aggchain_prover.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/454/commits/c96552b01eb80e968e52faa27bc3040f4bae6742

joanestebanr (approved):

Issue #325, opened 2025-03-22
Add support of aggkit-bridge-service in kurtosis while running aggkit


temaniarpit27:
Add option to use new `aggkit-bridge-service` instead of `zkevm-bridge-service` in kurtosis

PR #467, opened 2025-05-13
feat: Add support of aggkit-bridge-service in kurtosis while running aggkit

temaniarpit27:
https://github.com/agglayer/aggkit/issues/325


.github/test_e2e_single_chain_fork12_op_succinct_args.json
```diff
@@ -1,54 +1,50 @@
...
+    "deploy_cdk_bridge_infra": false
   },
   "args": {
     "verbosity": "debug",
     "aggkit_image": "aggkit:local",
-    "agglayer_image": "ghcr.io/agglayer/agglayer:0.3.0-rc.16",
-    "aggkit_prover_image": "ghcr.io/agglayer/aggkit-prover:0.1.0-rc.20",
-    "op_succinct_contract_deployer_image": "atanmarko/op-succinct-contract-deployer:v1.2.11-agglayer",
-    "op_succinct_server_image": "ghcr.io/agglayer/op-succinct/succinct-proposer:v1.2.12-agglayer",
-    "op_succinct_proposer_image": "ghcr.io/agglayer/op-succinct/op-proposer:v1.2.12-agglayer",
```
Stefan-Ethernal: Double check if this has been reverted by accident (IMO it is good to have pinned docker images we are used, because that way we are not dependent upon changes being merged to the Kurtosis, that could crash the CI suddenly).

WDYT?

Stefan-Ethernal (approved):

.github/test_e2e_single_chain_fork12_op_succinct_args.json
temaniarpit27: i removed it because i wanted these values to be the default ones. as we are not modifying any code, i think we should always be in sync with latest images. this way if there is some issue in our testcases, it should fail. if we are way behind the actual image, later on we face issue of fixing things and not able to find the root cause

PR #471, opened 2025-05-15
feat: support `TLS` in `gRPC`

goran-ethernal:
## Description

This PR supports configuring `gRPC` clients to expect `TLS` handshakes when connecting to a `gRPC` server.

We need this in real systems, which will use TLS.

### Config change

This PR adds two new config parameters to `Aggsender` config:

- `UseAgglayerTLS` - it says to the `aggsender grpc` communication client to do TLS handshakes when opening connections.
- `UseAggkitProverTLS` -  it says to the `aggsender grpc` communication client to do TLS handshakes when opening connections.

### Kurtosis CDK
Kurtosis doesn't have TLS, everything is in local, so these parameters should be false (they are false by default).

Fixes # (issue)

### DIFF
--- a/common/grpc_client.go
+++ b/common/grpc_client.go
@@ -1,11 +1,13 @@
 package common
 
 import (
+	"crypto/tls"
 	"fmt"
 	"strings"
 
 	"google.golang.org/genproto/googleapis/rpc/errdetails"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/status"
 )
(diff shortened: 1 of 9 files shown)


common/grpc_client.go
```diff
@@ -17,10 +19,16 @@ type Client struct {
...
 
 // NewClient initializes and returns a new gRPC client
-func NewClient(serverAddr string) (*Client, error) {
-	// TODO - Check if we need to use this
+func NewClient(serverAddr string, useTLS bool) (*Client, error) {
 	var opts []grpc.DialOption
-	opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
+
+	if useTLS {
+		creds := credentials.NewTLS(&tls.Config{InsecureSkipVerify: false, MinVersion: tls.VersionTLS12})
```
vcastellm, line 25: Is out intention to use self-signed certificates only?

goran-ethernal: I do not know. We need to discuss this. I will keep this on hold for now.

Stefan-Ethernal (approved):
:100: :fire:

config/default.go
```diff
@@ -238,6 +238,8 @@ GenerateAggchainProofTimeout="{{GenerateAggchainProofTimeout}}"
 SovereignRollupAddr = "{{L1Config.polygonZkEVMAddress}}"
 RequireStorageContentCompatibility = {{RequireStorageContentCompatibility}}
 RequireNoFEPBlockGap = true
+UseAgglayerTLS = false
```
joanestebanr, line 241: You must be able to provide the certificate to use

joanestebanr (commented):
I think that it's more convinient to be able to set the certificate to usage in each of the clients

config/default.go
goran-ethernal: Yeah, we need to have a discussion about this. Lets keep this on hold for now.

goran-ethernal: Ok, since we are acting only as a client, this might be enough of a solution for now. WDYT?

common/grpc_client.go
goran-ethernal: Ok, since we are acting only as a client, this might be enough of a solution for now, and we do not need to configure a certficate on the client side. WDYT?

joanestebanr (approved):

PR #472, opened 2025-05-15
feat: optimistic mode

### LINKED ISSUE #468: [FEATURE] Add support for Optimistic mode on the AggSender
1. Before requesting an OP FEP to aggkit prover, [check the smart contract](https://github.com/agglayer/agglayer-contracts/blob/v10.1.0-rc.1/contracts/v2/aggchains/AggchainFEP.sol#L93-L95) to see if the optimistic mode is enabled
2. If enabled, call [this RPC](https://github.com/agglayer/provers/pull/201/files#diff-cebf19293f0c3734ed221337b87cdeaf3eb868a344a6d45b9ad477d356e0311a) instead of following the normal OP FEP path

joanestebanr:
## Description
This PR add optimistic mode to aggkit. It checks if optimistic mode is active, if so, then instead of generating the FEP it call another end-point of `aggkit-prover` to generate a optimistic proof.

## Config: 
- Add new struct AggSender.OptimisticModeConfig:
   - `SovereignRollupAddr`: address of rollup, by default `AggSender.SovereignRollupAddr`
   - `TrustedSequencerKey`: key to sign optimistic certificates. Must match trustedSequencer and usually it's the same used for agglayer. So by default is `AggSender.AggsenderPrivateKey`
   - `OpNodeURL`: url to op-node
   - `RequireKeyMatchTrustedSequencer`: require that optmistic key match trustedKey (by default `true`)
  
- Define a new **defaut var** `AggsenderPrivateKey`  to allow override this key with  a KMS. It's still composed with `SequencerPrivateKeyPath` and `SequencerPrivateKeyPassword` so previous config file works in the same way.
- Disabled aggSender maxsize checked:  `AggSender.maxCertSize = 0`
- 

- Add new fields to DB (migration 0003): 
   - **extra_data**:  not structured data related to the cert for debug. e.g: optimistic store publicValues there
   - add new type `optimistic` to field **cert_type**

## Internal changes: 
- AggchainProverFlow: 
    -  configuration params have been grouped in a new struct `AggchainProverFlowConfig`
- Client for aggkit-prover have been moved from folder `aggsender/grpc` to `aggsender/aggchainproofclient`
- Moved variable `GenerateAggchainProofTimeout` from **MandatoryVars** (so it must be defined by the config-file) to **DefaultVars** that it defined a default variable so don't need to be defined on config-file
- Created a script `test/scipts/optimistic.sh` to allow check / modify contract status in kurtosis-cdk environment
- `baseFlow` object instead of a embedded object in the rest of flows is explicity call and use throught a interface to be able to mock in unittest
- Local debug scripts
      -  duplicated script to set a fep_optimistic debug: `local/local_config_fep_optimistic`and also created a specific config file for this case: `test/config/fep_optimistic.toml.template`
- Rendering confgfile allow to reference complex objects like the key configuration (it was required to allow `AggSender.OptimisticModeConfig.SovereignRollupAddr` = ` {{AggSender.SovereignRollupAddr}}`
Fixes #468

## Future tasks: 
- integrate e2e for optimistic


aggsender/flows/factory_test.go
```diff
@@ -79,7 +101,17 @@ func TestNewFlow(t *testing.T) {
...
 			mockL2BridgeSyncer.EXPECT().OriginNetwork().Return(1)
 			mockLogger := log.WithFields("test", "NewFlow")
+			funcNewEVMChainGERReader = func(_ common.Address, _ aggoracletypes.EthClienter) (*chaingerreader.EVMChainGERReader, error) {
+				return &chaingerreader.EVMChainGERReader{}, nil
+			}
+			funcGetL2StartBlock = func(_ common.Address, _ types.EthClient) (uint64, error) {
+				return 100, nil
+			}
 
+			mockL1Client.EXPECT().CallContract(mock.Anything, mock.Anything, mock.Anything).Return([]byte{1, 2, 3}, nil).Maybe()
```
goran-ethernal, line 110: Maybe you could do the same for `NewOptimisticModeQuerierFromContract` that you did for `var funcGetL2StartBlock = getL2StartBlock` and not have these mocks?

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -51,31 +56,85 @@ func getL2StartBlock(sovereignRollupAddr common.Address, l1Client types.EthClien
 
 var funcNewEVMChainGERReader = chaingerreader.NewEVMChainGERReader
 
+type AggchainProverFlowConfig struct {
```
goran-ethernal, line 62: Add comments on the struct and constructor functions.

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -26,11 +25,17 @@ var errNoProofBuiltYet = &aggkitcommon.GRPCError{
...
-	aggchainProofClient  grpc.AggchainProofClientInterface
-	gerQuerier           types.GERQuerier
-	requireNoFEPBlockGap bool
+	log     types.Logger
+	storage db.AggSenderStorage
+
+	aggchainProofClient   types.AggchainProofClientInterface
+	gerQuerier            types.GERQuerier
+	requireNoFEPBlockGap  bool
+	signer                signertypes.Signer
```
goran-ethernal: Should we rename this to `certificateSigner` perhaps. So that we know exactly what each signer do.

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -51,31 +56,85 @@ func getL2StartBlock(sovereignRollupAddr common.Address, l1Client types.EthClien
...
 		},
+		signer:                signer,
+		optimisticModeQuerier: optimisticModeQuerier,
+		optimisticSigner:      optimisticSigner,
+	}
+}
+
+// NewAggchainProverFlowInjectedBaseFlow returns a new instance of the AggchainProverFlow injecting baseFlow instead of
+// creating it
+func NewAggchainProverFlowInjectedBaseFlow(
```
goran-ethernal: If this is only used in tests, I would suggests to move this function to the test file, and make it a `test.Helper()`.

aggsender/flows/factory.go
```diff
@@ -62,28 +66,39 @@ func NewFlow(
...
+		if err != nil {
+			return nil, fmt.Errorf("aggchainProverFlow - error creating optimistic mode querier: %w", err)
+		}
 
 		return NewAggchainProverFlow(
 			logger,
-			cfg.MaxCertSize,
-			startL2Block,
+			AggchainProverFlowConfig{
+				baseFlowConfig: BaseFlowConfig{
```
goran-ethernal: If it has only two config params, I am not sure if it is worth it to have the `BaseFlowConfig` 😕  . It does make sense to have a config struct for the `AggchainProverFlow`, and then use the params from that to pass it to the base flow.

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -97,11 +156,11 @@ func (a *AggchainProverFlow) sanityCheckNoBlockGaps(lastSentCertificate *types.C
 		lastSentCertficateStr = fmt.Sprintf("cert from:%d, to:%d", lastSentCertificate.FromBlock, lastSentCertificate.ToBlock)
 	}
 	msg := fmt.Sprintf("aggchainProverFlow - sanityCheckNoBlockGaps - last sent certificate: %s, startL2Block:%d",
-		lastSentCertficateStr, a.startL2Block)
+		lastSentCertficateStr, a.baseFlow.StartL2Block())
```
goran-ethernal, line 131: I am not sure why `baseFlow` needs this `StartL2Block`. It is only needed for the sanity check on the `aggchain prover flow`.

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -171,8 +250,14 @@ func (a *AggchainProverFlow) GetCertificateBuildParams(ctx context.Context) (*ty
 
 		return buildParams, nil
 	}
+	// This line is just for emit a warning
```
```suggestion
	// This line is just for emitting a warning
```

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -300,14 +385,14 @@ func adjustBlockRange(buildParams *types.CertificateBuildParams,
 func (a *AggchainProverFlow) GenerateAggchainProof(
 	ctx context.Context,
 	lastProvenBlock, toBlock uint64,
-	claims []bridgesync.Claim,
+	certBuildParams *types.CertificateBuildParams,
 ) (*types.AggchainProof, *treetypes.Root, error) {
-	proof, leaf, root, err := a.l1InfoTreeDataQuerier.GetFinalizedL1InfoTreeData(ctx)
+	proof, leaf, root, err := a.baseFlow.L1InfoTreeDataQuerier().GetFinalizedL1InfoTreeData(ctx)
```
goran-ethernal: Hmmm, I am not sure about having this `baseFlow.GetInterface().Function()`. This makes mocking in the UTs even harder.

How about just having `GetFinalizedL1InfoTreeData` function on the `baseFlow` and just returning:
```go
func (b *baseFlow) GetFinalizedL1InfoTreeData() (*data, error) {
   return b.l1InfoTreeQuerier.GetFinalizedL1InfoTreeData()
 }
```
That way, in the test, you won't have to create two mocks, but just one. WDYT? Same goes for the `L2BridgeQuerier()`.

Then, after we merge this, we can separate the `baseFlow` into smaller interfaces that make more sense, and move away from this `base` class, to call it like that.

And once this gets merged, we can discuss

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -323,42 +408,47 @@ func (a *AggchainProverFlow) GenerateAggchainProof(
...
+		ImportedBridgeExitsWithBlockNumber: importedBridgeExits,
+	}
+	// It decide if must generate optimistic proof using CertType
+	optimisticMode := certBuildParams.CertificateType == types.CertificateTypeOptimistic
+	a.log.Infof("aggchainProverFlow - requesting proof lastProvenBlock: %d, maxEndBlock: %d, optimisticMode: %t",
+		lastProvenBlock, toBlock, optimisticMode)
+	if !optimisticMode {
+		aggchainProof, err = a.aggchainProofClient.GenerateAggchainProof(request)
+	} else {
+		newLER, err := a.baseFlow.GetNewLocalExitRoot(ctx, certBuildParams)
```
goran-ethernal: This else looks like it belongs to some separate interface or a struct, called `optimistic_proof_query`, something like that. WDYT? It will probably make it easier to unit test the `flow_aggchain_prover`.

aggsender/flows/flow_aggchain_prover.go
joanestebanr: This variable was originally there, I see that baseFlow use it. We can talk about that and remove from baseFlow, for me it's also really artificial because this value is set by AggchainFlow

aggsender/flows/flow_aggchain_prover.go
joanestebanr: Yes, I dislike it, but the idea was to reduce the impact introducing the interface. Another option is that AggchainFlow has his own instance of whatever use (like  `l1InfoTreeQuerier`)  that for me, make more sense

aggsender/flows/flow_aggchain_prover.go
goran-ethernal: Ahh, yes, sorry, I was only looking of references to the config field, not the function. I will resolve the comment.

aggsender/flows/flow_aggchain_prover.go
joanestebanr: I don't get your idea... what do you mean: move the optimisticMode 'else' to another function?

aggsender/flows/flow_aggchain_prover.go
goran-ethernal: That would be fine as well. `aggchainProverFlow` having a field of its own, called `l1infotreeQuerier` and `l2bridgeQuerier`, and just use them.

aggsender/flows/flow_aggchain_prover.go
goran-ethernal: Yes, to move this logic in the else part to another function or even another file in your `optimistic` package, called `optimistic_proof_query` (this would make unit testing easier for `aggchainProverFlow`).

aggsender/flows/flow_aggchain_prover.go
joanestebanr: do you think that `certificateSigner` could work?

aggsender/flows/flow_aggchain_prover.go
joanestebanr: The idea would be deprecate `NewAggchainProverFlow` and start using this `NewAggchainProverFlowInjectedBaseFlow` because is more close to depency injections principle

aggsender/flows/flow_aggchain_prover.go
goran-ethernal: Yes, that makes sense to me, since it is signing the certificate per se.

aggsender/flows/factory.go
joanestebanr: Yes, I agree, that now it's a bit artificial, but when inject the baseFlow into the object it's going to own `baseFlowConfig` the idea of be an struct is that we can add parameters without the painful of changing an extra param everywhere (that the reason of the Default constructor)

aggsender/flows/flow_aggchain_prover.go
goran-ethernal: I do agree, but in your code, this constructor is only used in UTs. We agreed, that we will slice this BaseFlow to smaller interfaces that make more sense, and discard it all together. But, until then, if it is not used in real logic, I would suggest moving it to UT file.

aggsender/flows/factory_test.go
joanestebanr: We have a huddle and agree a solution

aggsender/flows/flow_aggchain_prover.go
joanestebanr: I'm going to change the constructor for use this one

aggsender/flows/factory.go
joanestebanr: After chainging the constructor to the one that inject the baseFlow se can remove the field AggchainProverFlowConfig. baseFlowConfig because will be no longer in use

aggsender/flows/flow_aggchain_prover.go
joanestebanr: We agree to move this 'else' part to another function

aggsender/flows/factory_test.go
joanestebanr: We have agree that it's not required

aggsender/flows/flow_aggchain_prover.go
joanestebanr: Done 053a3865ca48edba8d1664c49417c83ca246c89e

aggsender/flows/flow_aggchain_prover.go
joanestebanr: We agree to remove the old constructor in favour of the new one with injection of baseFlow

joanestebanr: done c1004609cc217644782dd4d4302a6c75b05198b1

aggsender/flows/flow_aggchain_prover.go
joanestebanr: Done c1004609cc217644782dd4d4302a6c75b05198b1

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -51,31 +58,55 @@ func getL2StartBlock(sovereignRollupAddr common.Address, l1Client types.EthClien
...
 var funcNewEVMChainGERReader = chaingerreader.NewEVMChainGERReader
 
-// NewAggchainProverFlow returns a new instance of the AggchainProverFlow
-func NewAggchainProverFlow(log types.Logger,
-	maxCertSize uint,
-	startL2Block uint64,
-	aggkitProverClient grpc.AggchainProofClientInterface,
+// AggchainProverFlowConfig holds the configuration for the AggchainProverFlow
+type AggchainProverFlowConfig struct {
+	requireNoFEPBlockGap bool
```
goran-ethernal, line 63: Do we need this now, since its one parameter only? If we need more parameters in the future, we can add a config struct then.

aggsender/optimistic/optimistic.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/agglayer/aggkit/aggsender/types"
+	"github.com/agglayer/aggkit/log"
+)
+
+func NewOptimistic(ctx context.Context,
```
goran-ethernal, line 12: Add a comment to the function.

aggsender/optimistic/optimistic_aggegation_proof_public_values_query.go
```diff
@@ -0,0 +1,77 @@
...
+	"fmt"
+
+	"github.com/0xPolygon/cdk-contracts-tooling/contracts/pp/l2-sovereign-chain/aggchainfep"
+	optimistichash "github.com/agglayer/aggkit/aggsender/optimistic/optimistichash"
+	"github.com/agglayer/aggkit/opnode"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// The real object that implements OpNodeClienter is opnode.OpNodeClient
+var _ OpNodeClienter = (*opnode.OpNodeClient)(nil)
```
goran-ethernal:
```suggestion
var (
      _ OpNodeClienter = (*opnode.OpNodeClient)(nil)
      _ FEPContractQuerier = (*aggchainfep.Aggchainfep)(nil)
      _ OptimisticAggregationProofPublicValuesQuerier = (*OptimisticAggregationProofPublicValuesQuery)(nil)
)
      
```

aggsender/optimistic/optimistic_aggegation_proof_public_values_query.go
```diff
@@ -0,0 +1,77 @@
...
+
+var _ FEPContractQuerier = (*aggchainfep.Aggchainfep)(nil)
+
+var _ OptimisticAggregationProofPublicValuesQuerier = (*OptimisticAggregationProofPublicValuesQuery)(nil)
+
+type OptimisticAggregationProofPublicValuesQuery struct {
```
goran-ethernal, line 20: Add comments to struct an functions that are missing it.

aggsender/optimistic/optimistic_aggegation_proof_public_values_query.go
```diff
@@ -0,0 +1,77 @@
...
+) *OptimisticAggregationProofPublicValuesQuery {
+	return &OptimisticAggregationProofPublicValuesQuery{
+		aggchainFEPContract: aggchainFEPContract,
+		aggchainFEPAddr:     aggchainFEPAddr,
+		opNodeClient:        opNodeClient,
+		proverAddress:       proverAddress,
+	}
+}
+
+// The parametameters are contained in the AggchainProofRequest struct
```
goran-ethernal: I think the comment is missing that describes what this function does.

aggsender/optimistic/optimistic_mode_query.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/0xPolygon/cdk-contracts-tooling/contracts/pp/l2-sovereign-chain/aggchainfep"
+	"github.com/agglayer/aggkit/aggoracle/types"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+type OptimisticModeQuerierFromContract struct {
+	aggchainFEPContract FEPContractQuerier
+	aggchainFEPAddr     common.Address
+	// op-node client
```
goran-ethernal: Remove this comment?

aggsender/optimistic/optimistic_sign.go
```diff
@@ -0,0 +1,118 @@
...
+type OptimisticSignatureCalculator interface {
+	Sign(ctx context.Context,
+		aggchainReq types.AggchainProofRequest,
+		newLocalExitRoot common.Hash,
+		certBuildParams *types.CertificateBuildParams,
+	) (common.Hash, error)
+}
+
+type OptimisticSignatureCalculatorImpl struct {
+	QueryAggregationProofPublicValues OptimisticAggregationProofPublicValuesQuerier
```
goran-ethernal: It seems that you don't need to export the fields on this struct. They can be private.

aggsender/flows/flow_aggchain_prover.go
joanestebanr: done 2a67f21c04acb6fab2d4625b165b0dd517b3a8fb

aggsender/flows/factory.go
joanestebanr: done c1004609cc217644782dd4d4302a6c75b05198b1

aggsender/optimistic/optimistic_sign.go
```diff
@@ -0,0 +1,118 @@
...
+		Logger:                            logger,
+	}, nil
+}
+
+// Sign calculate hash and sign it.
+// It returns the signed hash, extra data for logging, and an error if any.
+func (o *OptimisticSignatureCalculatorImpl) Sign(ctx context.Context,
+	aggchainReq types.AggchainProofRequest,
+	newLocalExitRoot common.Hash,
+	certBuildParams *types.CertificateBuildParams,
```
goran-ethernal: Looks like you only need `Claims` from the certificate. Maybe you can pass just them, instead of build parameters object. WDYT?

aggsender/optimistic/optimistic_sign.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return nil, "", err
+	}
+	o.Logger.Infof("OptimisticSignatureCalculatorImpl.Sign agg:%s", aggregationProofPublicValues.String())
+	aggregationProofPublicValuesHash, err := aggregationProofPublicValues.Hash()
+	if err != nil {
+		return nil, "", fmt.Errorf("aggregationProofPublicValues.Hash: error hashing aggregationProofPublicValues: %w", err)
+	}
+	importedBridgesHash := optimistichash.CalculateCommitImportedBrdigeExitsHashFromClaims(certBuildParams.Claims)
+	o.Logger.Infof("OptimisticSignatureCalculatorImpl.Sign aggHash:%s", aggregationProofPublicValuesHash.Hex())
```
goran-ethernal: Could we maybe merge these three log lines into one, and put it at the end when we sign the hash, or if signing the hash fails?

aggsender/prover/proof_generation_tool.go
```diff
@@ -82,19 +87,29 @@ func NewAggchainProofGenerationTool(
 	}
 
 	l1InfoTreeQuerier := query.NewL1InfoTreeDataQuerier(l1Client, l1InfoTreeSyncer)
+	// TODO: the signer it's required?
```
goran-ethernal: You can remove this comment. The signer is not required, since this tool is only used for regular requests to generate `aggchain proof`. We can add support for optimistic proofs later if its needed, but for now, you can remove the TODO.

joanestebanr: I mean the certificateSigner, if you think that is not required I will deleted the TODO comment

goran-ethernal: Yes. Its not required.

aggsender/optimistic/optimistic_sign.go
joanestebanr: Done bac43c0756c4f63249d23127af6f70c5d359dfe3

aggsender/flows/flow_aggchain_prover.go
joanestebanr: I prefer in that way, is going to be more simple to add new parameters (that for sure we are going to need soon)

aggsender/db/migrations/0003.sql
```diff
@@ -1,11 +1,15 @@
 -- +migrate Down
 ALTER TABLE certificate_info DROP COLUMN cert_type;
 ALTER TABLE certificate_info DROP COLUMN cert_source;
+ALTER TABLE certificate_info DROP COLUMN extra_data;
```
temaniarpit27, line 4: i hope that this migration hasn't gone in v0.3

aggsender/optimistic/optimistic_sign.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/agglayer/aggkit/bridgesync"
+	"github.com/agglayer/aggkit/log"
+	"github.com/agglayer/aggkit/opnode"
+	"github.com/agglayer/go_signer/signer"
+	signertypes "github.com/agglayer/go_signer/signer/types"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// OptimisticSignatureCalculator is an interface that defines the method for signing optimistic aggregation proofs.
+// type OptimisticSignatureCalculator interface {
```
temaniarpit27: commented code?

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -323,67 +380,85 @@ func (a *AggchainProverFlow) GenerateAggchainProof(
...
-			root.Hash,
-			*leaf,
-			root.Hash,
-			proof,
-			injectedGERsProofs,
-			importedBridgeExits,
+		err := fmt.Errorf("aggchainProverFlow - error fetching aggchain proof (optimisticMode: %t) for lastProvenBlock: %d, "+
+			"maxEndBlock: %d. Err: %w. Message sent: %s", optimisticMode, lastProvenBlock, toBlock, err, request.String(),
 		)
+		a.log.Error(err.Error())
```
goran-ethernal, line 409: Do we need this log now? You are using it in the above lines to log it anyway.

goran-ethernal (approved):

aggsender/flows/flow_aggchain_prover.go
joanestebanr: Do you mean that is going to be logged eventually because we are returning as error? I think that is good to have the logs there because we can use the backtrace

goran-ethernal: Ah, you are shadowing the err object, to create a new one to log it. Which is fine, I guess.

We need to review how we are passing and handling errors in the aggsender in future versions, because we have a lot of cases where we have something like this:

FunctionC() - catches an error, logs it, wraps it to a new error, and returns it.
FunctionB() - gets the wrapped error from FunctionC, logs it, wraps it and returns it.
FunctionA() - gets the wrapped error from FunctionB, logs it. And now, we have errors that are logged and wrapped multiple times.

But, this is something for the `0.6.0` version.

aggsender/types/interfaces.go
```diff
@@ -27,6 +27,21 @@ type AggsenderFlow interface {
 		buildParams *CertificateBuildParams) (*agglayertypes.Certificate, error)
 }
 
+type AggsenderFlowBaser interface {
```
temaniarpit27, line 30: baser? or base?

joanestebanr: Baser... because it's an interface 😅

aggsender/optimistic/optimistic_sign.go
joanestebanr: oops!

aggsender/db/migrations/0003.sql
joanestebanr: No, this migration have been introduce in develop a few days ago, for that reason I take advatange of that instead of creating `0004.sql`

temaniarpit27 (approved):
LGTM

aggsender/flows/flow_base_test.go
```diff
@@ -236,3 +239,135 @@ func Test_baseFlow_getNewLocalExitRoot(t *testing.T) {
...
+	t.Parallel()
+	mockStorage := mocks.NewAggSenderStorage(t)
+	sut := NewBaseFlow(
+		log.WithFields("test", t.Name()),
+		nil,
+		mockStorage,
+		nil,
+		NewBaseFlowConfigDefault())
+
+	t.Run("we can't generate new cert because a peding cert (not inError not Settle) we don't known if is going to settle or not ",
```
rachit77:
```suggestion
	t.Run("cannot generate new certificate because a pending certificate (neither InError nor Settled) has an unknown settlement status",
```

how about this?

joanestebanr: Done 460cbb04e34ec8475ad9115659468620df9d6c41

joanestebanr: Done 460cbb04e34ec8475ad9115659468620df9d6c41

Issue #469, opened 2025-05-15
add a filter on from_address field - bridge and claims


temaniarpit27:
add a filter on from_address field. Since the test frontend app will directly fetch from aggKit, bridge service.

PR #473, opened 2025-05-16
feat: from_address filter

temaniarpit27:
#469


bridgesync/processor.go
```diff
@@ -618,22 +610,27 @@ func (p *processor) buildBridgesFilterClause(depositCount *uint64, networkIDs []
 		clauses = append(clauses, buildNetworkIDsFilter(networkIDs, "destination_network"))
 	}
 
+	if fromAddress != "" && common.IsHexAddress(fromAddress) {
+		clauses = append(clauses, fmt.Sprintf("UPPER(from_address) LIKE '%s'", fromAddress))
```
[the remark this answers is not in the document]
temaniarpit27: it works this way as well

.mockery.yaml
```diff
@@ -73,7 +73,8 @@ packages:
     config:
       dir: "{{ .InterfaceDir }}/mocks"
       all: true
-      exclude: "^UintParam$"
+      exclude:
+        - "^UintParam$"
```
Stefan-Ethernal, line 77: Seems like this approach doesn't work, as the mock for UintParam is generated https://github.com/agglayer/aggkit/blob/0acf52ae18b23ea073ac0fc6309c389c69432f2e/bridgeservice/mocks/mock_uint_param.go#L1

bridgesync/processor.go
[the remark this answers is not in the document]
Stefan-Ethernal: IMO the Mr. Copilot has right here:
we should either be sure that we are comparing the same case.

So for example, to be safe, I would say, let's convert everything to upper case

```suggestion
		clauses = append(clauses, fmt.Sprintf("UPPER(from_address) LIKE UPPER('%s')", fromAddress))
```

Also, consider using equality check (it should be slightly more performant), e.g.

```go
if common.IsHexAddress(fromAddress) {
    normalizedAddress := strings.ToLower(fromAddress)
    clauses = append(clauses, fmt.Sprintf("LOWER(from_address) = '%s'", normalizedAddress))
}
```

bridgesync/processor.go
```diff
@@ -910,13 +916,13 @@
...
 // GetTotalNumberOfRecords returns the total number of records in the given table
-func (p *processor) GetTotalNumberOfRecords(tableName string) (int, error) {
+func (p *processor) GetTotalNumberOfRecords(tableName, whereClause string) (int, error) {
 	if !tableNameRegex.MatchString(tableName) {
 		return 0, fmt.Errorf("invalid table name '%s' provided", tableName)
 	}
 
 	count := 0
-	err := p.db.QueryRow(fmt.Sprintf(`SELECT COUNT(*) AS count FROM %s;`, tableName)).Scan(&count)
+	err := p.db.QueryRow(fmt.Sprintf(`SELECT COUNT(*) AS count FROM %s %s;`, tableName, whereClause)).Scan(&count)
```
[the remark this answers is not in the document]
Stefan-Ethernal: Maybe we should check if `whereClause` is non-empty string?

Otherwise we would end up with extra space on the end of the query:
```sql
SELECT COUNT(*) AS count FROM foo ;
```

Stefan-Ethernal (approved):
:100: generally LGTM

.mockery.yaml
temaniarpit27: but if we dont do this, we get an error

temaniarpit27:
```
aggkit/test|arpit/469⚡ ⇒  make generate-mocks
mockery --disable-version-string --with-expecter --exported
16 May 25 18:30 IST INF Starting mockery dry-run=false version=v2.53.3
16 May 25 18:30 IST INF Using config: /Users/arpit.temani/Workspace/Polygon/aggkit/.mockery.yaml dry-run=false version=v2.53.3
failed to discover recursive packages: failed to get package config: 1 error(s) decoding:

* 'exclude': source data must be an array or slice, got string
make: *** [generate-mocks] Error 1
```

temaniarpit27: @Stefan-Ethernal did this work for you earlier?

Stefan-Ethernal: TBH, I cannot remember anymore, but probably not (seems like I have manually removed the mock and thought that this configuration was correct).

I'm getting the same error as you do. I'll let you know if I find the correct solution.

temaniarpit27: ok merging this for now then.
we can then raise a PR if needed

PR #483, opened 2025-05-16
feat: be able to disable sanity check of block range gap PP to FEP

### LINKED ISSUE #482: feat: be able to disable sanity check of block range gap  PP to FEP
Add a config flag to disable the sanity check #436 and allow to use a network with that gap

joanestebanr:
## Description

Add flag `RequireNoFEPBlockGap` to say if it's mandatory this sanity check

## Default value
```
[AggSender]
(...)
RequireNoFEPBlockGap = true
```

Fixes #482

### DIFF
--- a/aggsender/config.go
+++ b/aggsender/config.go
@@ -65,6 +65,9 @@ type Config struct {
 	// RequireStorageContentCompatibility is true it's mandatory that data stored in the database
 	// is compatible with the running environment
 	RequireStorageContentCompatibility bool `mapstructure:"RequireStorageContentCompatibility"`
+	// RequireNoFEPBlockGap is true if the AggSender should not accept a gap between
+	// lastBlock from lastCertificate and first block of FEP
+	RequireNoFEPBlockGap bool `mapstructure:"RequireNoFEPBlockGap"`
 }
 
 func (c Config) CheckCertConfigBriefString() string {
--- a/aggsender/flows/flow_aggchain_prover.go
+++ b/aggsender/flows/flow_aggchain_prover.go
@@ -21,8 +21,9 @@ import (
 type AggchainProverFlow struct {
 	*baseFlow
 
-	aggchainProofClient grpc.AggchainProofClientInterface
-	gerReader           types.ChainGERReader
+	aggchainProofClient  grpc.AggchainProofClientInterface
+	gerReader            types.ChainGERReader
+	requireNoFEPBlockGap bool
 }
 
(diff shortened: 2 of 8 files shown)


aggsender/config.go
```diff
@@ -65,6 +65,9 @@ type Config struct {
 	// RequireStorageContentCompatibility is true it's mandatory that data stored in the database
 	// is compatible with the running environment
 	RequireStorageContentCompatibility bool `mapstructure:"RequireStorageContentCompatibility"`
+	// RequireNoFEPBlockGap is true if the AggSender should not accept a gap between
+	// lastBlock from lastCertificate and first block of FEP
+	RequireNoFEPBlockGap bool `mapstructure:"RequireNoFEPBlockGap"`
```
Stefan-Ethernal, line 70: Do we want to keep it `false` by default?

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -87,7 +90,15 @@ func (a *AggchainProverFlow) CheckInitialStatus(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("aggchainProverFlow - error getting last sent certificate: %w", err)
 	}
-	return a.sanityCheckNoBlockGaps(lastSentCertificate)
+	err = a.sanityCheckNoBlockGaps(lastSentCertificate)
```
Stefan-Ethernal: Do we even care performing a sanity check if it the `requireNoFEPBlockGap` config is set to `false`?
It seems to me that the answer is not straight-forward.

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -87,7 +88,15 @@ func (a *AggchainProverFlow) CheckInitialStatus(ctx context.Context) error {
...
+	if a.requireNoFEPBlockGap {
```
goran-ethernal: This is more cleaner to me, but its just a suggestion. I would change the log a bit though:

```suggestion
	if err := a.sanityCheckNoBlockGaps(lastSentCertificate); err != nil {
	     if a.requireNoFEPBlockGAP {
	           return err
	     }
	     
	     a.log.Warnf("aggchainProverFlow - ignoring block gaps due to RequireNoFEPBlockGap config. Err: %w", err}
```

Stefan-Ethernal (approved):

goran-ethernal (approved):

PR #489, opened 2025-05-19
fix: choose right initial block if there are a gap

### LINKED ISSUE #482: feat: be able to disable sanity check of block range gap  PP to FEP

joanestebanr:
## Description

In some case of upgrade from PP to FEP with a gap  the system choose the wrong starting block for requesting a certificate. 
This is a intermediate fix: it already have some pending fixes: 
- The range of blocks could no respect the `startL2Block` (#502)
- There are no check that after a upgrade from PP to FEP it's forbidden that the last cert is InError (#499)
- Missing some sanity checks in the block gap to avoid missing relevant information (#498)

Fixes #482

### DIFF
--- a/aggsender/flows/flow_aggchain_prover.go
+++ b/aggsender/flows/flow_aggchain_prover.go
@@ -412,6 +412,12 @@ func (a *AggchainProverFlow) getLastProvenBlock(fromBlock uint64, lastCertificat
 			lastCertificate.ToBlock, a.startL2Block)
 		return a.startL2Block
 	}
+	if fromBlock+1 < a.startL2Block {
+		// if the fromBlock is less than the starting L2 block, we need to start from the starting L2 block
+		a.log.Infof("aggchainProverFlow - getLastProvenBlock. FromBlock: %d < startL2Block: %d",
+			fromBlock, a.startL2Block)
+		return a.startL2Block
+	}
 
 	return fromBlock - 1
 }
--- a/aggsender/flows/flow_aggchain_prover.go
+++ b/aggsender/flows/flow_aggchain_prover.go
@@ -378,10 +378,11 @@ func (a *AggchainProverFlow) GenerateAggchainProof(
 	return aggchainProof, root, nil
 }
 
-func (a *AggchainProverFlow) getLastProvenBlock(fromBlock uint64) uint64 {
-	if fromBlock == 0 {
+func (a *AggchainProverFlow) getLastProvenBlock(fromBlock uint64, lastCertificate *types.CertificateInfo) uint64 {
+	if fromBlock == 0 || (lastCertificate != nil && lastCertificate.ToBlock < a.startL2Block) {
 		// if this is the first certificate, we need to start from the starting L2 block
 		// that we got from the sovereign rollup
+		// if the last certificate is settled on PP, the last proven block is the starting L2 block
 		return a.startL2Block
 	}
 
--- a/aggsender/flows/flow_aggchain_prover.go
+++ b/aggsender/flows/flow_aggchain_prover.go
@@ -100,17 +100,19 @@ func (a *AggchainProverFlow) sanityCheckNoBlockGaps(lastSentCertificate *types.C
 	if lastSentCertificate != nil {
 		lastSentCertficateStr = fmt.Sprintf("cert from:%d, to:%d", lastSentCertificate.FromBlock, lastSentCertificate.ToBlock)
 	}
-	a.log.Infof("aggchainProverFlow - sanityCheckNoBlockGaps - last sent certificate: %s, startL2Block:%d", lastSentCertficateStr, a.startL2Block)
+	msg := fmt.Sprintf("aggchainProverFlow - sanityCheckNoBlockGaps - last sent certificate: %s, startL2Block:%d", lastSentCertficateStr, a.startL2Block)
 	if lastSentCertificate != nil && lastSentCertificate.ToBlock+1 < a.startL2Block {
 		err := fmt.Errorf("gap of blocks detected: lastSentCertificate.ToBlock: %d, startL2Block: %d",
 			lastSentCertificate.ToBlock, a.startL2Block)
 		if a.requireNoFEPBlockGap {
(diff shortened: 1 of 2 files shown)


aggsender/flows/flow_aggchain_prover.go
```diff
@@ -96,15 +96,25 @@ func (a *AggchainProverFlow) CheckInitialStatus(ctx context.Context) error {
...
 		if a.requireNoFEPBlockGap {
+			a.log.Error("%s. Err: %s", msg+" fails!", err.Error())
 			return err
 		}
 		// The sanity check is disabled
-		a.log.Warnf("aggchainProverFlow - ignoring block gaps due to RequireNoFEPBlockGap. Err: %w", err)
+		a.log.Warnf("%s. Ignoring block gaps due to RequireNoFEPBlockGap. Err: %w", msg, err)
+		return nil
 	}
+	a.log.Infof("%s. pass", msg)
```
```suggestion
	a.log.Infof("%s. Passed check.", msg)
```

goran-ethernal (approved):

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -96,15 +96,25 @@ func (a *AggchainProverFlow) CheckInitialStatus(ctx context.Context) error {
 // sanityCheckNoBlockGaps checks that there are no gaps in the block range for next certificate
 // #436. Don't allow gaps updating from PP to FEP
 func (a *AggchainProverFlow) sanityCheckNoBlockGaps(lastSentCertificate *types.CertificateInfo) error {
+	lastSentCertficateStr := "nil"
```
```suggestion
	lastSentCertficateStr := types.NilStr
```

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -378,10 +398,24 @@ func (a *AggchainProverFlow) GenerateAggchainProof(
...
+			a.startL2Block)
+		return a.startL2Block
+	}
+	if lastCertificate != nil && lastCertificate.ToBlock < a.startL2Block {
+		// if the last certificate is settled on PP, the last proven block is the starting L2 block
+		a.log.Infof("aggchainProverFlow - getLastProvenBlock. Last certificate block: %d < startL2Block: %d",
+			lastCertificate.ToBlock, a.startL2Block)
+		return a.startL2Block
+	}
+	if fromBlock+1 < a.startL2Block {
```
goran-ethernal: `fromBlock` is basically `lastProvenBlock + 1` (or in other words, `lastSettledCert.ToBlock + 1`). 

So if we have this case:
- `lastSettledCertificate.ToBlock = 49`, `newCertificate.FromBlock `startL2Block = 50`, `fromBlock = 50`, meaning there is a gap of one block, and if gap check is disabled, then we need to return `startL2Block` as last proven block.

If we leave it like this, the `fromBlock + 1 (51) < a.startBlock (50)` check will be false, and we will return `fromBlock - 1 = 49`.

This `lastProvenBlock` change on the prover introduced a lot of confusion now on `aggsender`, since the new range is `lastProvenBlock + 1` - `max_end_block`, so in this case, we need to start asking for new range of `51 - some end block`.

joanestebanr: good catch!, fixed!

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -122,7 +132,12 @@ func (a *AggchainProverFlow) GetCertificateBuildParams(ctx context.Context) (*ty
 	if lastSentCertificateInfo != nil && lastSentCertificateInfo.Status == agglayertypes.InError {
 		// if the last certificate was in error, we need to resend it
 		a.log.Infof("resending the same InError certificate: %s", lastSentCertificateInfo.String())
-
+		lastProvenBlock := a.getLastProvenBlock(lastSentCertificateInfo.FromBlock, lastSentCertificateInfo)
+		if lastSentCertificateInfo.FromBlock != lastProvenBlock+1 {
+			a.log.Warnf("aggchainProverFlow - last sent certificate inError fromBlock: %d doesn't match lastProvenBlock: %d."+
```
```suggestion
			a.log.Warnf("aggchainProverFlow - last sent certificate is InError and its fromBlock: %d doesn't match lastProvenBlock: %d + 1. Check update process 😅", lastSentCertificateInfo.FromBlock, lastProvenBlock)
```

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -170,6 +185,12 @@ func (a *AggchainProverFlow) GetCertificateBuildParams(ctx context.Context) (*ty
 
 		return nil, err
 	}
+	lastProvenBlock := a.getLastProvenBlock(buildParams.FromBlock, lastSentCertificateInfo)
+	if buildParams.FromBlock != lastProvenBlock+1 {
```
goran-ethernal, line 188: I think we need to move this to the `getLastSentBlockAndRetryCount` function, because, here, you already got the bridges and claims for the given range, so to not ping the syncer database again, we might want to move it there.

joanestebanr: Yes totally agree, but in a separated PR

goran-ethernal: I know, but what can happen is that, you can take bridges and claims from block 50-100 for example, and the last proven block is 110.

Then, we need to get the bridges and claims again if this condition is true with the new fixed range. If we just fix the range, and not get the bridges and claims again, we might end up sending to the prover claims that are not in that range. Am I right?

joanestebanr: We agree that is an issue but that is something that is going to prevent a valid certificate in case that the block gap between PP and FEP have some relevant data , so right now it's a colateral effect that is good. So until we don't have the sanity check of block gap this is going to mitigate the damage is case of bad upgrade. WDYT?

aggsender/flows/flow_aggchain_prover.go
goran-ethernal: Fixed: https://github.com/agglayer/aggkit/pull/489/commits/ec617301f003f2e31e70e63cb73a52fa687af9bb

PR #495, opened 2025-05-19
feat: add `signature` to `aggchain data` in `certificate`

goran-ethernal:
## Description

This PR updates the `GenerateAggchainProofRequest` and `SubmitCertificate` APIs on `aggkit prover` and `agglayer` respectively.

The `AggchainProofFlow` on the `aggsender` now implements signing of the `FEP` certificate. The signer is the `trustedSequencer` address, configured in the `AggsenderPrivateKey` config parameter (similar to the old `PessimisticProof` flow).

The hash that needs to be signed differs from the one we are signing in the `PessimisticProof` flow, and is calculated by hashing this data from the certificate:
- new_local_exit_root
- claim_hash (the hash of all the global index + bridge exit hash from the imported bridge exits in certificate)
- Certificate's height
- aggchain_params

Fixes # (issue)

### DIFF
--- a/aggsender/types/types.go
+++ b/aggsender/types/types.go
@@ -100,13 +100,14 @@ type CertStatus struct {
 }
 
 type AggchainProof struct {
-	LastProvenBlock uint64            `json:"last_proven_block"`
-	EndBlock        uint64            `json:"end_block"`
-	CustomChainData []byte            `json:"custom_chain_data,omitempty"`
-	LocalExitRoot   common.Hash       `json:"local_exit_root"`
-	AggchainParams  common.Hash       `json:"aggchain_params"`
-	Context         map[string][]byte `json:"context,omitempty"`
-	SP1StarkProof   *SP1StarkProof    `json:"sp1_stark_proof,omitempty"`
+	LastProvenBlock uint64
+	EndBlock        uint64
+	CustomChainData []byte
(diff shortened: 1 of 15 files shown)


joanestebanr (approved):

aggsender/types/types.go
```diff
@@ -100,13 +100,14 @@ type CertStatus struct {
...
-	Context         map[string][]byte `json:"context,omitempty"`
-	SP1StarkProof   *SP1StarkProof    `json:"sp1_stark_proof,omitempty"`
+	LastProvenBlock uint64
+	EndBlock        uint64
+	CustomChainData []byte
+	LocalExitRoot   common.Hash
+	AggchainParams  common.Hash
+	Context         map[string][]byte
+	SP1StarkProof   *SP1StarkProof
+	Signature       []byte
```
Stefan-Ethernal, line 110: Does it make sense including the hex encoded string to `String` function?

Stefan-Ethernal (approved):
LGTM

aggsender/types/types.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/495/commits/3d0ce0c50ec179922cdf8c19988826da840e76be

PR #501, opened 2025-05-20
feat: introduce swagger docs for bridge service

### LINKED ISSUE #267: Swagger docs aggkit bridge service
re asses the docs

github having documentation

Stefan-Ethernal:
## Description

This PR introduces swagger docs and generation of it for bridge service API. It is hosted under `<aggkit_bridge_addr>/bridge/v1/swagger/` route.

Fixes #267


test/run-local-e2e.sh
```diff
@@ -16,15 +16,20 @@ log_error() {
 
 trap 'log_error "Script failed at line $LINENO"' ERR
 
-if [ "$#" -ne 4 ]; then
-    echo "Usage: $0 <test_type: single-l2-network-fork12-op-succinct | single-l2-network-fork12-pessimistic | multi-l2-networks> <path/to/kurtosis/cdk/repo> <path/to/e2e/repo> <run_tests: true | false>"
+if [ "$#" -lt 2 ]; then
+    echo "Usage: $0 <test_type: single-l2-network-fork12-op-succinct | single-l2-network-fork12-pessimistic | multi-l2-networks> <kurtosis_repo> [e2e_repo] [run_tests: true|false]"
```
temaniarpit27, line 20: @Stefan-Ethernal maybe we need similar changes in cdk repo

Makefile
```diff
@@ -76,6 +76,12 @@ test-unit:
 lint: ## Runs the linter
 	export "GOROOT=$$(go env GOROOT)" && $$(go env GOPATH)/bin/golangci-lint run --timeout 5m
 
+.PHONY: generate-swagger-docs
+generate-swagger-docs: ## Generates the swagger docs
+	@command -v swag >/dev/null 2>&1 || { echo >&2 "swag not installed. Please install swaggo/swag: https://github.com/swaggo/swag"; exit 1; }
```
goran-ethernal: I think we do these checks in separate functions, like:

```
.PHONY: check-go
check-go:
	@which go > /dev/null || (echo "Error: Go is not installed" && exit 1)

# Targets that require the checks
build: check-go

.PHONY: build ## Builds the binaries locally into ./target
build: build-aggkit build-tools
```

goran-ethernal (approved):
LGTM 💯

Makefile
Stefan-Ethernal: Fixed and cleaned up the Makefile (there were some unused targets) https://github.com/agglayer/aggkit/pull/501/commits/6df2bab98a7c3b79d66ba60f7095f7834a2f7e22

bridgeservice/types/types.go
```diff
@@ -1,39 +1,348 @@
...
+	SovereignToken
+)
+
+func (l TokenMappingType) String() string {
+	return [...]string{"WrappedToken", "SovereignToken"}[l]
+}
+
+// Proof represents a Merkle proof for a tree of a given height
+// @Description Merkle proof structure for a tree of a given height
+type Proof [tree.DefaultHeight]Hash
```
temaniarpit27, line 52: cant we directly use this treetypes.Proof? for swagger do we need to define it here?

Stefan-Ethernal: Yes, unfortunately we need to create this type, because the `treetypes.Proof` contains `common.Hash` from the `go-ethereum`, and it is unable to generate the docs when we have the non-trivial data types (such as `common.Hash`).

temaniarpit27 (approved):

Issue #479, opened 2025-05-16
Sync checker for new version of aggkit bridge service


temaniarpit27:

temaniarpit27:
solu 1 - compare the deposit count and bridge on contract and in aggkit bridge service. if both of them are same over a period of time and in sync then aggkit node is synced

PR #504, opened 2025-05-20
feat: add sync status

temaniarpit27:
#479


bridgeservice/types/types.go
```diff
@@ -37,3 +37,17 @@ type ClaimsResult struct {
...
 	Count  int                         `json:"count"`
 }
+
+// SyncStatus represents the synchronization status of the bridge service
+type SyncStatus struct {
+	L1 struct {
+		ContractDepositCount uint32 `json:"contract_deposit_count"`
+		BridgeDepositCount   uint32 `json:"bridge_deposit_count"`
+		IsSynced             bool   `json:"is_synced"`
+	} `json:"l1"`
```
Stefan-Ethernal: Why don't we extract the struct out of this (in order to prevent code repetitions), wdyt?

```go
type SyncStatus struct {
    L1Info *NetworkSyncInfo `json:"l1"`
    L2Info *NetworkSyncInfo `json:"l2"`
}

type NetworkSyncInfo struct {
    ContractDepositCount uint32 `json:"contract_deposit_count"`
    BridgeDepositCount   uint32 `json:"bridge_deposit_count"`
    IsSynced             bool   `json:"is_synced"`
}
```

Also we should add `example` and `@Description`, to keep the Swagger happier :slightly_smiling_face: 
For reference, take a look to #501 (`bridgeservice/types.go`).

bridgesync/bridgesync.go
```diff
@@ -407,3 +411,22 @@ func sanityCheckContract(logger *log.Logger, bridgeAddr common.Address,
...
+// GetContractDepositCount returns the last deposit count from the bridge contract
+func (s *BridgeSync) GetContractDepositCount(ctx context.Context) (uint32, error) {
+	if s.processor.isHalted() {
+		return 0, sync.ErrInconsistentState
+	}
+
+	bridge, err := polygonzkevmbridgev2.NewPolygonzkevmbridgev2(s.bridge, s.ethClient)
+	if err != nil {
+		return 0, fmt.Errorf("failed to create bridge contract instance: %w", err)
+	}
```
Stefan-Ethernal: I'd prefer to instantiate it only once and preserve it in a field instead, wdyt?

bridgesync/bridgesync.go
```diff
@@ -55,6 +55,8 @@ type BridgeSync struct {
 	originNetwork uint32
 	reorgDetector ReorgDetector
 	blockFinality etherman.BlockNumberFinality
+	ethClient     aggkittypes.EthClienter
+	bridge        common.Address
```
Stefan-Ethernal: We don't need this field, it if we capture the bridge SC binding instead?
```suggestion
```

bridgesync/bridgesync.go
```diff
@@ -55,6 +55,8 @@ type BridgeSync struct {
 	originNetwork uint32
 	reorgDetector ReorgDetector
 	blockFinality etherman.BlockNumberFinality
+	ethClient     aggkittypes.EthClienter
```
Stefan-Ethernal: In case we capture the bridge SC binding, we don't need this field as well?
```suggestion
        bridge *polygonzkevmbridgev2.Polygonzkevmbridgev2
```

Stefan-Ethernal (approved):
💯

PR #509, opened 2025-05-21
chore: add monitoring logs to bridge service

rachit77:
Issue #435


temaniarpit27 (approved):
💯

temaniarpit27:
@rachit77 also pls check about monitoring. do we need any metrics to be collected when the bridge service is running?
we have metrics which shows how much an api has been hit but imo that doesn't provide us much info. see if we need to change that or add anything else as well

bridgesync/bridgesync.go
```diff
@@ -245,13 +245,15 @@ func (s *BridgeSync) GetClaimsPaged(
...
 	if s.processor.isHalted() {
+		s.processor.log.Error("processor is halted, cannot get claims")
 		return nil, 0, sync.ErrInconsistentState
 	}
 	return s.processor.GetClaimsPaged(ctx, page, pageSize, networkIDs, fromAddress)
 }
 
 // Start starts the synchronization process
 func (s *BridgeSync) Start(ctx context.Context) {
+	s.processor.log.Info("starting bridge synchronization")
```
Stefan-Ethernal: wdyt?
```suggestion
	s.processor.log.Info("starting bridge synchronizer")
```

claimsponsor/evmclaimsponsor.go
```diff
@@ -64,8 +64,10 @@ func NewEVMClaimSponsor(
 	waitTxToBeMinedPeriod time.Duration,
 	waitOnEmptyQueue time.Duration,
 ) (*ClaimSponsor, error) {
+	logger.Info("initializing EVM claim sponsor")
```
Stefan-Ethernal: Consistency sake
```suggestion
	logger.Info("initializing evm claim sponsor")
```

bridgesync/processor.go
```diff
@@ -872,41 +881,48 @@ func (p *processor) ProcessBlock(ctx context.Context, block sync.Block) error {
...
 			_, err := tx.Exec(deleteLegacyTokenSQL, event.RemoveLegacyToken.LegacyTokenAddress.Hex())
 			if err != nil {
+				p.log.Errorf("failed to remove legacy token at block %d: %v", block.Num, err)
 				return err
 			}
 		}
 	}
 
 	if err := tx.Commit(); err != nil {
+		p.log.Errorf("failed to commit transaction for block %d: %v", block.Num, err)
```
Stefan-Ethernal: wdyt?

```suggestion
		p.log.Errorf("failed to commit db transaction (block number %d): %v", block.Num, err)
```

bridgesync/processor.go
```diff
@@ -804,29 +804,34 @@ func (p *processor) getLastProcessedBlockWithTx(tx db.Querier) (uint64, error) {
...
 			if errRllbck := tx.Rollback(); errRllbck != nil && !errors.Is(errRllbck, sql.ErrTxDone) {
-				log.Errorf("error while rolling back tx %v", errRllbck)
+				p.log.Errorf("error rolling back reorg transaction: %v", errRllbck)
 			}
 		}
 	}()
 
 	res, err := tx.Exec(`DELETE FROM block WHERE num >= $1;`, firstReorgedBlock)
 	if err != nil {
+		p.log.Errorf("failed to delete blocks during reorg: %v", err)
```
Stefan-Ethernal, line 662: Do we need such logs, because if I'm not mistaken, we are logging errors on the top level (OTOH it does make sense to make error messages more verbose)
https://github.com/agglayer/aggkit/blob/96c7cbaad068e0bc131a5d9eeb3f14183f6339ba/sync/evmdriver.go#L224

bridgesync/processor.go
```diff
@@ -837,28 +842,32 @@ func (p *processor) Reorg(ctx context.Context, firstReorgedBlock uint64) error {
...
 	if err != nil {
+		p.log.Errorf("failed to start transaction for block %d: %v", block.Num, err)
 		return err
 	}
 	shouldRollback := true
 	defer func() {
 		if shouldRollback {
 			if errRllbck := tx.Rollback(); errRllbck != nil && !errors.Is(errRllbck, sql.ErrTxDone) {
-				log.Errorf("error while rolling back tx %v", errRllbck)
+				p.log.Errorf("error rolling back transaction for block %d: %v", block.Num, errRllbck)
```
Stefan-Ethernal: wdyt?
```suggestion
				p.log.Errorf("error rolling back db transaction (block number %d): %v", block.Num, errRllbck)
```

bridgesync/downloader.go
```diff
@@ -55,25 +56,32 @@ const (
...
 	bridgeSovereignChain, err := bridgel2sovereignchain.NewBridgel2sovereignchain(bridgeAddr, client)
 	if err != nil {
+		logger.Errorf("failed to create BridgeL2SovereignChain SC binding: %v", err)
 		return nil, fmt.Errorf("failed to create BridgeL2SovereignChain SC binding (bridge addr: %s): %w",
 			bridgeAddr, err)
 	}
 
 	gasTokenAddress, err := bridgeContractV2.GasTokenAddress(nil)
 	if err != nil {
+		logger.Errorf("failed to get gas token address: %v", err)
```
Stefan-Ethernal: Do we need this log, IMO the error message we are propagating is verbose enough. WDYT?

bridgesync/downloader.go
```diff
@@ -55,25 +56,32 @@ const (
...
 		return nil, fmt.Errorf("error parsing gas token address: %w", err)
```
Stefan-Ethernal:
```suggestion
		return nil, fmt.Errorf("failed to get gas token address: %w", err)
```

bridgeservice/bridge.go
```diff
@@ -401,12 +420,14 @@ func (b *BridgeService) GetLegacyTokenMigrationsHandler(c *gin.Context) {
 
 	networkID, err := parseUintQuery(c, networkIDParam, true, uint32(0))
 	if err != nil {
+		b.logger.Warnf("invalid network ID parameter: %v", err)
 		c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
 		return
 	}
 
 	ctx, cancel, pageNumber, pageSize, setupErr := b.setupRequest(c, "get_legacy_token_migrations")
 	if setupErr != nil {
```
Stefan-Ethernal: Unrelated to this PR, but maybe it would make sense renaming the `setupErr` to `err` (implies to the other functions as well), wdyt?

Stefan-Ethernal (approved):
:100: generally LGTM, leaving a couple of comments to consider.

bridgesync/processor.go
rachit77: I think it’s useful to keep some low-level logs in error cases like this.
DB tx failures are rare, so this shouldn't add much noise or Datadog volume, but can really help with debugging.

wdyt?

bridgesync/downloader.go
rachit77: yeah that makes sense this logs seems a bit redundant 
i will remove it

bridgeservice/bridge.go
rachit77: either way it looks good,
i will rename `setupErr` to `err`

PR #515, opened 2025-05-22
feat: provide gin mode based on environment variable

### LINKED ISSUE #470: check the build mode for gin library - release or debug

Stefan-Ethernal:
## Description

The gin library has multiple modes, since it is recommended to use the release mode in production, this PR allows to provide the gin mode, by specifying the `GIN_MODE` env variable (if the variable is missing or invalid, it fallbacks to release mode).

Fixes #470

### DIFF
--- a/bridgeservice/bridge.go
+++ b/bridgeservice/bridge.go
@@ -9,6 +9,7 @@ import (
 	"math"
 	"math/big"
 	"net/http"
+	"os"
 	"time"
 
 	"github.com/agglayer/aggkit/bridgeservice/types"
@@ -84,6 +85,14 @@ func New(
 ) *BridgeService {
 	meter := otel.Meter(meterName)
 	cfg.Logger.Infof("starting bridge service (network id=%d)", cfg.NetworkID)
+
+	ginMode := os.Getenv("GIN_MODE")
+	switch ginMode {
+	case gin.DebugMode, gin.ReleaseMode, gin.TestMode:
+		gin.SetMode(ginMode)
+	default:
+		gin.SetMode(gin.ReleaseMode) // fallback to release mode
+	}
 
 	b := &BridgeService{
 		logger:       cfg.Logger,
--- a/bridgeservice/bridge.go
+++ b/bridgeservice/bridge.go
@@ -94,6 +94,8 @@ func New(
 	case gin.DebugMode, gin.ReleaseMode, gin.TestMode:
 		gin.SetMode(ginMode)
 	default:
+		cfg.Logger.Infof("invalid or missing GIN_MODE value ('%s') provided, defaulting to '%s' mode",
+			ginMode, gin.ReleaseMode)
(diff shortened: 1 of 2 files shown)


bridgeservice/bridge.go
```diff
@@ -85,6 +86,14 @@ func New(
 	meter := otel.Meter(meterName)
 	cfg.Logger.Infof("starting bridge service (network id=%d)", cfg.NetworkID)
 
+	ginMode := os.Getenv("GIN_MODE")
```
temaniarpit27, line 92: @Stefan-Ethernal where are we setting this var?

Stefan-Ethernal: Currently nowhere, and it fallbacks to `release` mode. 

I'm unsure what is the best approach for setting it.

Thinking outloud, but should we create two targets in the `Makefile`:
- `build-debug` - builds a debug Docker image
- `build-release` - builds a release Docker image

temaniarpit27: yes but then this will change the workflow which is already set for others. building 2 docker files and all will be hassle which i dont think its worth time and effort. `release` mode can be the default mode until someone has usecase to modify this
maybe we can set this var to `release` mode in makefile and put a comment that if someone wants to build with a diff mode, modify this
maybe we can see if there are other variables like this where we want to set different for debug and release, then i think we can do a overall revamp of builds and then for kurtosis in any upcoming release
wdyt?

temaniarpit27 (approved):

bridgeservice/bridge.go
Stefan-Ethernal: Fixed https://github.com/agglayer/aggkit/pull/515/commits/adc650512a35affbbd91161dd851c09a0e405804

bridgeservice/bridge.go
```diff
@@ -85,6 +86,17 @@ func New(
...
 
+	// The GIN_MODE environment variable controls the mode of the Gin framework.
+	// Valid values are "debug", "release", and "test". If an invalid value is provided,
+	// the mode defaults to "release" for safety and performance.
+	ginMode := os.Getenv("GIN_MODE")
+	switch ginMode {
+	case gin.DebugMode, gin.ReleaseMode, gin.TestMode:
+		gin.SetMode(ginMode)
+	default:
+		gin.SetMode(gin.ReleaseMode) // fallback to release mode
```
goran-ethernal, line 99: Should we log this situation at least?

goran-ethernal (approved):
LGTM

bridgeservice/bridge.go
Stefan-Ethernal: Makes sense... I'll add it.

Stefan-Ethernal: Fixed https://github.com/agglayer/aggkit/pull/515/commits/f1e089f0647b96a903c9c4d3adf3381063652b0c

PR #516, opened 2025-05-22
feat: disable cert max size, store type of cert in db, still WIP

joanestebanr:
## Description

- Disable max size cert check
- Store on DB the type of cert (pp/fep)
- Check gaps

Fixes # (issue)


joanestebanr:
This PR had moved to https://github.com/agglayer/aggkit/pull/528

PR #526, opened 2025-05-26
feat: cherry pick #471 - support `TLS` in `gRPC`

goran-ethernal:
## Description

This PR cherry picks PR #471 to `develop`.

This PR supports configuring `gRPC` clients to expect `TLS` handshakes when connecting to a `gRPC` server.

We need this in real systems, which will use TLS.

### Config change

This PR adds two new config parameters to `Aggsender` config:

- `UseAgglayerTLS` - it says to the `aggsender grpc` communication client to do TLS handshakes when opening connections.
- `UseAggkitProverTLS` -  it says to the `aggsender grpc` communication client to do TLS handshakes when opening connections.

### Kurtosis CDK
Kurtosis doesn't have TLS, everything is in local, so these parameters should be false (they are false by default).

### DIFF
--- a/agglayer/grpc/agglayer_grpc_client.go
+++ b/agglayer/grpc/agglayer_grpc_client.go
@@ -28,11 +28,11 @@ type AgglayerGRPCClient struct {
 }
 
 // NewAggchainProofClient initializes a new AggchainProof instance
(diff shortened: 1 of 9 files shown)


vcastellm (approved):
LGTM

joanestebanr (approved):

Issue #499, opened 2025-05-20
feat: aggsender: store on DB the type of certificate (PP, FEP)


joanestebanr:
# Feature Request
- This fix can implies to store on DB the type of certificate (PP, FEP)

PR #528, opened 2025-05-26
feat: store in DB the type of certificate (PP/FEP)

joanestebanr:
## Description

Add a new field on DB:
-  `cert_type`:  PP or FEP certificate. 
- `cert_source`:  agglayer or local

The `cert_type` is stored on cert metadata, so it can be retrieved by aggsender in case that need to resync

Fixes: #499

### DIFF
--- a/aggsender/db/migrations/0003.sql
+++ b/aggsender/db/migrations/0003.sql
@@ -1,7 +1,11 @@
 -- +migrate Down
 ALTER TABLE certificate_info DROP COLUMN cert_type;
+ALTER TABLE certificate_info DROP COLUMN cert_source;
 ALTER TABLE certificate_info_history DROP COLUMN cert_type;
+ALTER TABLE certificate_info_history DROP COLUMN cert_source;
 
 -- +migrate Up
 ALTER TABLE certificate_info ADD COLUMN cert_type VARCHAR DEFAULT "";
+ALTER TABLE certificate_info ADD COLUMN cert_source VARCHAR DEFAULT "";
 ALTER TABLE certificate_info_history ADD COLUMN cert_type VARCHAR DEFAULT "";
+ALTER TABLE certificate_info_history ADD COLUMN cert_source VARCHAR DEFAULT "";
--- a/aggsender/db/migrations/0003.sql
+++ b/aggsender/db/migrations/0003.sql
@@ -1,0 +1,7 @@
+-- +migrate Down
+ALTER TABLE certificate_info DROP COLUMN cert_type;
+ALTER TABLE certificate_info_history DROP COLUMN cert_type;
+
+-- +migrate Up
+ALTER TABLE certificate_info ADD COLUMN cert_type VARCHAR DEFAULT "";
+ALTER TABLE certificate_info_history ADD COLUMN cert_type VARCHAR DEFAULT "";
--- a/aggsender/statuschecker/cert_status_checker.go
+++ b/aggsender/statuschecker/cert_status_checker.go
@@ -235,9 +235,11 @@ func newCertificateInfoFromAgglayerCertHeader(c *agglayertypes.CertificateHeader
 	}
 	now := uint32(time.Now().UTC().Unix())
 	meta := types.NewCertificateMetadataFromHash(c.Metadata)
-	var toBlock uint64
-	var createdAt uint32
-	var certType types.CertificateType
+	var (
+		toBlock   uint64
+		createdAt uint32
+		certType  types.CertificateType
+	)
 
 	switch meta.Version {
 	case types.CertificateMetadataV0:
--- a/aggsender/statuschecker/cert_status_checker.go
+++ b/aggsender/statuschecker/cert_status_checker.go
@@ -217,7 +217,10 @@ func (c *certStatusChecker) executeInitialStatusAction(ctx context.Context,
 // updateLocalStorageWithAggLayerCert updates the local storage with the certificate from the AggLayer
 func (c *certStatusChecker) updateLocalStorageWithAggLayerCert(ctx context.Context,
 	aggLayerCert *agglayertypes.CertificateHeader) (*types.Certificate, error) {
-	cert := newCertificateInfoFromAgglayerCertHeader(aggLayerCert)
+	cert, err := newCertificateInfoFromAgglayerCertHeader(aggLayerCert)
+	if err != nil {
+		return nil, fmt.Errorf("error creating certificate from AggLayer header: %w", err)
+	}
 	if cert == nil {
 		return nil, nil
 	}
@@ -226,9 +229,9 @@ func (c *certStatusChecker) updateLocalStorageWithAggLayerCert(ctx context.Conte
 	return cert, c.storage.SaveLastSentCertificate(ctx, *cert)
 }
 
-func newCertificateInfoFromAgglayerCertHeader(c *agglayertypes.CertificateHeader) *types.Certificate {
+func newCertificateInfoFromAgglayerCertHeader(c *agglayertypes.CertificateHeader) (*types.Certificate, error) {
 	if c == nil {
-		return nil
+		return nil, nil
 	}
 	now := uint32(time.Now().UTC().Unix())
 	meta := types.NewCertificateMetadataFromHash(c.Metadata)
@@ -250,8 +253,8 @@ func newCertificateInfoFromAgglayerCertHeader(c *agglayertypes.CertificateHeader
 		createdAt = meta.CreatedAt
 		certType = types.NewCertificateTypeFromInt(meta.CertType)
 	default:
-		log.Warnf("aggsender: unsupported certificate metadata version: %d", meta.Version)
-		return nil
+		return nil, fmt.Errorf("newCertificateInfoFromAgglayerCertHeader."+
+			" Unsupported certificate metadata version: %d", meta.Version)
 	}
 
 	res := &types.Certificate{
@@ -274,5 +277,5 @@ func newCertificateInfoFromAgglayerCertHeader(c *agglayertypes.CertificateHeader
 		res.Header.PreviousLocalExitRoot = c.PreviousLocalExitRoot
 	}
 
-	return res
+	return res, nil
 }
--- a/aggsender/statuschecker/cert_status_checker.go
+++ b/aggsender/statuschecker/cert_status_checker.go
@@ -232,34 +232,44 @@ func newCertificateInfoFromAgglayerCertHeader(c *agglayertypes.CertificateHeader
 	}
 	now := uint32(time.Now().UTC().Unix())
 	meta := types.NewCertificateMetadataFromHash(c.Metadata)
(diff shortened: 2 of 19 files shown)


aggsender/types/types.go
```diff
@@ -171,6 +182,20 @@ type Certificate struct {
 	AggchainProof     *AggchainProof `meddler:"aggchain_proof,aggchainproof"`
 }
 
+func (c *Certificate) GetCertType() CertificateType {
```
goran-ethernal: I think we need to expand this function to receive that `startL2Block` which denotes on which block the `FEP` network started, because, there is a case that happened on one of the upgrades, where they deleted `aggsender db` and `agglayer` had a `FEP` certificate. Since they do not return `aggchain proof` when we sync the state with `agglayer`, in current code, we would select `CerType == CertificateTypePP`, which is not correct.

joanestebanr: good point, I add to certificate metadata the type of certificate to mitigate this case

goran-ethernal: Yes, that will be useful for sure, but, for certificates that were added before V2 of metadata (on `v0.3.0` of `aggkit`), you will not have that in the metadata field. So that issue is still there in that case.

joanestebanr: You are right! The cert submited by `v0.3.0` we are not going to known the type. I have added `startL2Block` to do the guessing

aggsender/types/types.go
```diff
@@ -18,6 +17,56 @@ const (
 	AggchainProofMode    AggsenderMode = "AggchainProof"
 )
 
+type CertificateType string
```
goran-ethernal: To me it looks like we can just use one type, instead of having two. Maybe it is easiest to use the `CertificateType uint8`, and then have functions that return a string representation, and have a function that convert string to int. Having two types, its kinda difficult in my opinion to understand and know when to use.

We can have something like this:

```go
type CertificateType uint8

const (
	CertificateTypeUnknown CertificateType = 0
	CertificateTypePP      CertificateType = 1
	CertificateTypeFEP     CertificateType = 2
)

func (c CertificateType) String() string {
	switch c {
	case CertificateTypePP:
		return "PP"
	case CertificateTypeFEP:
		return "FEP"
	default:
		return "Unknown"
	}
}

func ParseCertificateType(s string) CertificateType {
	switch s {
	case "PP":
		return CertificateTypePP
	case "FEP":
		return CertificateTypeFEP
	default:
		return CertificateTypeUnknown
	}
}
```

Or have it backwards, and have the type `string` and have functions that parse it to `uint8`, and remove the other type.

But in my opinion using `int` (or `uint8`) enum is more clean, and just have `.String()` function on it.

aggsender/types/types.go
```diff
@@ -18,6 +17,56 @@ const (
...
+	default:
+		return CertificateTypeUnknown
+	}
+}
+
+type CertificateSource string
+
+const (
+	CertificateSourceAggLayer CertificateSource = "agglayer"
+	CertificateSourceLocal    CertificateSource = "Local"
```
goran-ethernal: To follow the same pattern:

```suggestion
	CertificateSourceLocal    CertificateSource = "local"
```

aggsender/types/types.go
```diff
@@ -138,7 +196,8 @@ func (c *CertificateHeader) ID() string {
 	if c == nil {
 		return NilStr
 	}
-	return fmt.Sprintf("%d/%s (retry %d)", c.Height, c.CertificateID.String(), c.RetryCount)
+	return fmt.Sprintf("%d/%s (retry: %d, type: %s)",
```
goran-ethernal, line 225: Should we log the source as well?

aggsender/types/types.go
```diff
@@ -171,6 +230,28 @@ type Certificate struct {
...
+func (c *Certificate) DetermineCertType(startL2Block uint64) CertificateType {
+	if c == nil {
+		return CertificateTypeUnknown
+	}
+	if c.Header.CertType == CertificateTypeUnknown {
+		if c.AggchainProof != nil {
+			return CertificateTypeFEP
+		}
+		// If the certificate is not set, we can determine the type based on the FromBlock
+		if startL2Block == 0 {
```
goran-ethernal, line 268: Hmm, thinking about this case. If we are in `PessimisticProofMode`, this `startL2Block` is 0. But, I agree, it is tricky to decide. Maybe someone mistakenly started `aggsender` in `PP` mode, but we had `FEP` mode before. 

My head is going to explode from these edge cases 🤯 .

aggsender/types/certificate_metadata.go
```diff
@@ -0,0 +1,104 @@
...
+	"log"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+const (
+	CertificateMetadataV0 = uint8(0) // Pre v1 metadata, only ToBlock is stored
+	CertificateMetadataV1 = uint8(1) // Post v1 metadata, FromBlock, Offset, CreatedAt are stored
+	CertificateMetadataV2 = uint8(2) // Same V1 + typeCert
```
goran-ethernal:
```suggestion
	CertificateMetadataV2 = uint8(2) // Same V1 + certType
```

aggsender/statuschecker/cert_status_checker.go
```diff
@@ -232,12 +232,26 @@ func newCertificateInfoFromAgglayerCertHeader(c *agglayertypes.CertificateHeader
 	}
 	now := uint32(time.Now().UTC().Unix())
 	meta := types.NewCertificateMetadataFromHash(c.Metadata)
-	toBlock := meta.FromBlock + uint64(meta.Offset)
-	createdAt := meta.CreatedAt
+	var toBlock uint64
```
goran-ethernal:
```suggestion
	var (
	   toBlock uint64
	   createdAt uint32
	   certType types.CertificateType
	)
```

aggsender/types/types.go
joanestebanr: I agree, but the problem is that if I want that database have the string not the number, and if I create a type `uint8` then DB store the number. If you known how fix that I would appreciate your help

aggsender/types/types.go
joanestebanr: I don't think so.. it's not so relevant the source of the information for be part of the ID()

aggsender/types/types.go
goran-ethernal: No easy solution for that case. Only writing a custom meddler. Which is of course, more complex than this. Let's leave it like this then.

goran-ethernal (approved):

aggsender/types/types.go
joanestebanr: Finally I implement some meddler function to be able to have a uint8 enum and store as string

aggsender/db/migrations/0003.sql
```diff
@@ -0,0 +1,11 @@
+-- +migrate Down
+ALTER TABLE certificate_info DROP COLUMN cert_type;
+ALTER TABLE certificate_info DROP COLUMN cert_source;
+ALTER TABLE certificate_info_history DROP COLUMN cert_type;
+ALTER TABLE certificate_info_history DROP COLUMN cert_source;
+
+-- +migrate Up
+ALTER TABLE certificate_info ADD COLUMN cert_type VARCHAR DEFAULT "";
+ALTER TABLE certificate_info ADD COLUMN cert_source VARCHAR DEFAULT "";
```
Stefan-Ethernal, line 9: Would this make sense for already existing certificates in the db?

```suggestion
ALTER TABLE certificate_info ADD COLUMN cert_source VARCHAR DEFAULT "local";
```

aggsender/db/migrations/0003.sql
```diff
@@ -0,0 +1,11 @@
...
+ALTER TABLE certificate_info_history ADD COLUMN cert_type VARCHAR DEFAULT "";
```
Stefan-Ethernal, line 10: Wdyt?
```suggestion
ALTER TABLE certificate_info_history ADD COLUMN cert_type VARCHAR DEFAULT "pp";
```

aggsender/db/migrations/0003.sql
```diff
@@ -0,0 +1,11 @@
+-- +migrate Down
+ALTER TABLE certificate_info DROP COLUMN cert_type;
+ALTER TABLE certificate_info DROP COLUMN cert_source;
+ALTER TABLE certificate_info_history DROP COLUMN cert_type;
+ALTER TABLE certificate_info_history DROP COLUMN cert_source;
+
+-- +migrate Up
+ALTER TABLE certificate_info ADD COLUMN cert_type VARCHAR DEFAULT "";
```
Stefan-Ethernal, line 8: Wdyt?
```suggestion
ALTER TABLE certificate_info ADD COLUMN cert_type VARCHAR DEFAULT "pp";
```

aggsender/db/migrations/0003.sql
```diff
@@ -0,0 +1,11 @@
...
+ALTER TABLE certificate_info ADD COLUMN cert_source VARCHAR DEFAULT "";
+ALTER TABLE certificate_info_history ADD COLUMN cert_type VARCHAR DEFAULT "";
+ALTER TABLE certificate_info_history ADD COLUMN cert_source VARCHAR DEFAULT "";
```
Stefan-Ethernal, line 11: Wdyt?
```suggestion
ALTER TABLE certificate_info_history ADD COLUMN cert_source VARCHAR DEFAULT "local";
```

Stefan-Ethernal (approved):
💯

aggsender/types/certificate_metadata.go
```diff
@@ -0,0 +1,104 @@
...
+		return &CertificateMetadata{
+			Version:   version,
+			FromBlock: binary.BigEndian.Uint64(b[1:9]),
+			Offset:    binary.BigEndian.Uint32(b[9:13]),
+			CreatedAt: binary.BigEndian.Uint32(b[13:17]),
+			CertType:  b[17],
+		}
+	} else {
+		// Unsupported version, return nil
+		log.Panicf("aggsender: unsupported certificate metadata version: %d", version)
```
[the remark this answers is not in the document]
joanestebanr: It returns an error

aggsender/db/migrations/0003.sql
joanestebanr: No, the previous certificate we don't known the source

aggsender/db/migrations/0003.sql
joanestebanr: The same, the data on DB we don't known the type of certificate, there are a function on code that try to guess it.. but don't cover the 100% of cases

aggsender/db/migrations/0003.sql
joanestebanr: The same, we don't known the type

aggsender/db/migrations/0003.sql
joanestebanr: The same we don't known the source

PR #544, opened 2025-05-29
fix: shadowed `err` in `db` txn commit

goran-ethernal:
## Description

This PR fixes error shadowing in processors of different syncers when they are trying to modify their database.

Since we use the `err` variable in the `defer` function to rollback `db` transaction if an error ocurrs, error shadowing is not allowed.

Fixes # (issue)


Stefan-Ethernal (approved):
:100: LGTM

Optionally consider to apply the same pattern, with a dedicated flag here as well
https://github.com/agglayer/aggkit/blob/f2ca1a57b0eac13f5c88d99594491aa209faa1da/aggsender/db/aggsender_db_storage.go#L264-L286

goran-ethernal:
> 💯 LGTM
> 
> Optionally consider to apply the same pattern, with a dedicated flag here as well
> 
> https://github.com/agglayer/aggkit/blob/f2ca1a57b0eac13f5c88d99594491aa209faa1da/aggsender/db/aggsender_db_storage.go#L264-L286

Done

bridgesync/processor.go
```diff
@@ -263,6 +265,9 @@ func (p *processor) Reorg(ctx context.Context, firstReorgedBlock uint64) error {
 	if err := tx.Commit(); err != nil {
```
temaniarpit27, line 265: can we just not redeclare the err here?
if we do `if err = tx.Commit(); err != nil {`, will it not work as expected?

temaniarpit27, line 265: if we not redeclare here, i think this should do the job

goran-ethernal: We are following the same pattern in other functions, using the `shouldRollback` flag. It would work as you described, but its more easier to use a bool flag for this, and then, not care about re-declaring errors, especially in complex and long functions.

temaniarpit27: but then you always need to make sure you are handling `shouldRollback` flag correctly which i think is more prone to error rather than making sure you dont redeclare `err`
anyway it was my opinion but we can definitely take a look at it later on

goran-ethernal: This was not my idea. This was actually a subject on one of the office hours. We decided to go with this, which is, in my opinion, much easier than to make sure you not shadow an err, because we had complex functions that changed the `db`, where code became much more hard to look at because of this. With this, you just set it to false after the commit, and that is it.

temaniarpit27 (approved):

PR #545, opened 2025-05-29
feat: save non-accepted certificates to `db`

### LINKED ISSUE #517: Consider saving certificate to `db` before sending it to `agglayer`
There were some cases where `agglayer` returned an error on certificate submission (when sending `submitCertificate` request).

It can be useful to save this certificate to the `certificate_info_history` table at least, so we can debug the issue more easily.

goran-ethernal:
## Description

This PR introduces saving of non-accepted certificates to a separate table in `aggsender db`: `nonaccepted_certificates`. 

This is done for easier debugging and root cause analysis when `agglayer` does not accept a certificate, but returns an error when the request for `submitCertificate` is sent.

These certificates need to be rebuild from scratch, and can not be saved in existing tables, because, they have are not resent for now. We will use them just as a way to do a root cause analysis if a given problem ocurrs.

Also, keep in mind, that they will be saved in `db` only if `KeepHistory` config parameter on `aggsender` is `true`.

Fixes #517

### DIFF
--- a/aggsender/db/aggsender_db_storage.go
+++ b/aggsender/db/aggsender_db_storage.go
@@ -386,9 +386,7 @@ func (a *AggSenderSQLStorage) SaveNonAcceptedCertificate(
 		}
 	}()
 
-	var raw []byte
-
-	raw, err = json.Marshal(certificate)
+	raw, err := json.Marshal(certificate)
 	if err != nil {
 		return fmt.Errorf("failed to marshal non-accepted certificate: %w", err)
 	}
--- a/aggsender/db/aggsender_db_storage.go
+++ b/aggsender/db/aggsender_db_storage.go
@@ -3,9 +3,11 @@ package db
 import (
 	"context"
 	"database/sql"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	agglayertypes "github.com/agglayer/aggkit/agglayer/types"
 	"github.com/agglayer/aggkit/aggsender/db/migrations"
@@ -60,6 +62,11 @@ type AggSenderStorage interface {
 	// and the aggchain proof if the certificate is in error
 	GetLastSentCertificateHeaderWithProofIfInError(
 		ctx context.Context) (*types.CertificateHeader, *types.AggchainProof, error)
+	// SaveNonAcceptedCertificate saves a non-accepted certificate in the storage
+	SaveNonAcceptedCertificate(
+		ctx context.Context, certificate *agglayertypes.Certificate, createdAt uint32) error
+	// GetNonAcceptedCertificates returns a list of non-accepted certificates
+	GetNonAcceptedCertificates() ([]*agglayertypes.Certificate, error)
 }
 
 var _ AggSenderStorage = (*AggSenderSQLStorage)(nil)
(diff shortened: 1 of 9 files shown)


aggsender/db/aggsender_db_storage.go
```diff
@@ -350,6 +364,78 @@ func (a *AggSenderSQLStorage) GetLastSentCertificateHeaderWithProofIfInError(
...
+// and to allow for debugging and analysis of why they were not accepted.
+func (a *AggSenderSQLStorage) SaveNonAcceptedCertificate(
+	ctx context.Context, certificate *agglayertypes.Certificate, createdAt uint32) error {
+	if !a.cfg.KeepCertificatesHistory {
+		return nil // non-accepted certificates are not saved in history
+	}
+
+	tx, err := db.NewTx(ctx, a.db)
+	if err != nil {
+		return fmt.Errorf("SaveNonAcceptedCertificate NewTx. Err: %w", err)
```
Stefan-Ethernal: wdyt?

```suggestion
		return fmt.Errorf("failed to create db transaction for non-accepted certificate persistence: %w", err)
```

aggsender/db/aggsender_db_storage.go
```diff
@@ -350,6 +364,78 @@ func (a *AggSenderSQLStorage) GetLastSentCertificateHeaderWithProofIfInError(
...
+		SignedCertificate: string(raw),
+		CreatedAt:         createdAt,
+	}
+
+	if err = meddler.Insert(tx, "nonaccepted_certificates", nonAcceptedCert); err != nil {
+		return fmt.Errorf("error inserting non-accepted certificate: %w", err)
+	}
+
+	if err = tx.Commit(); err != nil {
+		return fmt.Errorf("SaveNonAcceptedCertificate commit. Err: %w", err)
```
Stefan-Ethernal: wdyt?

```suggestion
		return fmt.Errorf("failed to commit db transaction for non-accepted certificate: %w", err)
```

aggsender/db/aggsender_db_storage.go
```diff
@@ -350,6 +364,78 @@ func (a *AggSenderSQLStorage) GetLastSentCertificateHeaderWithProofIfInError(
...
+	}
+
+	nonAcceptedCert := &nonAcceptedCertificate{
+		Height:            certificate.Height,
+		SignedCertificate: string(raw),
+		CreatedAt:         createdAt,
+	}
+
+	if err = meddler.Insert(tx, "nonaccepted_certificates", nonAcceptedCert); err != nil {
+		return fmt.Errorf("error inserting non-accepted certificate: %w", err)
```
Stefan-Ethernal:
```suggestion
		return fmt.Errorf("failed to insert non-accepted certificate: %w", err)
```

aggsender/db/aggsender_db_storage.go
```diff
@@ -350,6 +364,78 @@ func (a *AggSenderSQLStorage) GetLastSentCertificateHeaderWithProofIfInError(
...
+				a.logger.Errorf(errWhileRollbackFormat, errRllbck)
+			}
+		}
+	}()
+
+	var raw []byte
+
+	raw, err = json.Marshal(certificate)
+	if err != nil {
+		return fmt.Errorf("error marshalling non-accepted certificate: %w", err)
```
Stefan-Ethernal:
```suggestion
		return fmt.Errorf("failed to marshal non-accepted certificate: %w", err)
```

aggsender/db/migrations/0003.sql
```diff
@@ -3,9 +3,17 @@ ALTER TABLE certificate_info DROP COLUMN cert_type;
...
-ALTER TABLE certificate_info_history ADD COLUMN cert_source VARCHAR DEFAULT "";
\ No newline at end of file
+ALTER TABLE certificate_info_history ADD COLUMN cert_source VARCHAR DEFAULT "";
+
+CREATE TABLE nonaccepted_certificates (
+    height                     INTEGER NOT NULL,
+    created_at                 INTEGER NOT NULL,
+    signed_certificate         TEXT,
+    PRIMARY KEY (height, created_at)
+);
```
Stefan-Ethernal, line 19: Do we need a separate table for this, I mean could we persist it in the `certificate_info_history`, under some unique status (`rejected` or what not)?

aggsender/db/aggsender_db_storage.go
```diff
@@ -198,25 +205,30 @@ func (a *AggSenderSQLStorage) SaveLastSentCertificate(ctx context.Context, certi
 		}
 	}()
 
-	certificateInfo, err := convertCertificateToCertificateInfo(&certificate)
+	var (
+		certInfo *certificateInfo
+		certInDB *certificateInfo
+	)
```
Stefan-Ethernal: It should be safe to revert to the old logic, because we are relying on a separate flag (`shouldRollback`) to determine if we should rollback or not. WDYT?

Although I agree that this naming is clearer than it was before.

aggsender/db/migrations/0003.sql
goran-ethernal: No, that is the point, we can not put it in the `certificate_info_history` table. That table is reserved for accepted certificates and their retries. When a certificate gets accepted, but can not be settled, so we retry to send it. Non-accepted certificates, are those for which agglayer returned an error immediately when we tried to send it (probably because of some data invalidity, or what not, so we need to rebuild them from scratch).

aggsender/db/aggsender_db_storage.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/545/commits/ee2bed05948f11054bb3eb416cfba7989ebc4326

aggsender/db/aggsender_db_storage.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/545/commits/ee2bed05948f11054bb3eb416cfba7989ebc4326

aggsender/db/aggsender_db_storage.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/545/commits/ee2bed05948f11054bb3eb416cfba7989ebc4326

aggsender/db/aggsender_db_storage.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/545/commits/ee2bed05948f11054bb3eb416cfba7989ebc4326

aggsender/db/aggsender_db_storage.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/545/commits/ee2bed05948f11054bb3eb416cfba7989ebc4326

aggsender/db/migrations/0003.sql
Stefan-Ethernal: I understand, but OTOH you could differentiate such certificates by setting some exclusive `status` value and you would not resend them to the agglayer.

The thing is that I find it better to store certificates this way, having separate columns for each field.

If you think it would unnecessarily complicate the logic, we can leave it this way.

aggsender/db/aggsender_db_storage.go
```diff
@@ -350,6 +359,78 @@ func (a *AggSenderSQLStorage) GetLastSentCertificateHeaderWithProofIfInError(
...
+		if shouldRollback {
+			if errRllbck := tx.Rollback(); errRllbck != nil {
+				a.logger.Errorf(errWhileRollbackFormat, errRllbck)
+			}
+		}
+	}()
+
+	var raw []byte
+
+	raw, err = json.Marshal(certificate)
```
Stefan-Ethernal:
```suggestion
	raw, err := json.Marshal(certificate)
```

Stefan-Ethernal (approved):
:100:

aggsender/db/aggsender_db_storage.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/545/commits/60498a6a25e56070d2b3aa0e2cb011a138ce371d

aggsender/db/migrations/0003.sql
```diff
@@ -3,9 +3,17 @@ ALTER TABLE certificate_info DROP COLUMN cert_type;
 ALTER TABLE certificate_info DROP COLUMN cert_source;
```
temaniarpit27, line 3: since v0.3 is released, shouldn't we do this in a new migration file?

goran-ethernal: This migration is not a part of v0.3.0. This was added after it.

aggsender/db/migrations/0003.sql
```diff
@@ -3,9 +3,17 @@ ALTER TABLE certificate_info DROP COLUMN cert_type;
...
 ALTER TABLE certificate_info_history DROP COLUMN cert_type;
 ALTER TABLE certificate_info_history DROP COLUMN cert_source;
+DROP TABLE IF EXISTS nonaccepted_certificates;
```
rachit77, line 6: how about adding some unit test for db migrations?

rachit77 (approved):
💯

aggsender/db/migrations/0003.sql
goran-ethernal: Added UTs for each migration file: https://github.com/agglayer/aggkit/pull/545/commits/fc5bea142136afec293856c663545662f5b1c14a

PR #547, opened 2025-05-30
fix: grpc error comparison

goran-ethernal:
## Description

This PR fixes the `grpc` error comparison, to compare code and message as well.

Fixes # (issue)

### DIFF
--- a/common/grpc_client.go
+++ b/common/grpc_client.go
@@ -63,6 +63,21 @@ func (e GRPCError) Error() string {
 	return fmt.Sprintf("Code: %s, Message: %s, Details: %s", e.Code.String(), e.Message, joinDetails(e.Details))
 }
 
+// Is is an implementation of the error.Is interface for GRPCError.
+// It checks if the provided error matches the GRPCError's code and message.
+func (e GRPCError) Is(err error) bool {
+	if e.Code == codes.Unknown {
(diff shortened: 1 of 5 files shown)


temaniarpit27 (approved):

common/grpc_client.go
```diff
@@ -63,6 +63,21 @@ func (e GRPCError) Error() string {
...
+// Is is an implementation of the error.Is interface for GRPCError.
+// It checks if the provided error matches the GRPCError's code and message.
+func (e GRPCError) Is(err error) bool {
+	if e.Code == codes.Unknown {
+		return false // Do not match Unknown errors
+	}
+
+	if grpcErr, ok := err.(*GRPCError); ok {
+		return e.Code == grpcErr.Code && strings.Contains(
+			strings.ToLower(e.Message), strings.ToLower(grpcErr.Message))
```
Stefan-Ethernal, line 75: Should we insist that the error messages are equal?

Stefan-Ethernal (approved):

common/grpc_client.go
goran-ethernal: Good question. But since this is our own implementation, when we unpack a `grpc` error, I think this is good enough to protect us from possible changes of the message, or not knowing the entire message of the error that the prover or agglayer returns.

PR #551, opened 2025-05-30
feat: disable claim sponsor

temaniarpit27:
#550


bridgeservice/bridge.go
```diff
@@ -147,8 +143,8 @@ func (b *BridgeService) registerRoutes() {
 		bridgeGroup.GET("/l1-info-tree-index", b.L1InfoTreeIndexForBridgeHandler)
 		bridgeGroup.GET("/injected-l1-info-leaf", b.InjectedL1InfoLeafHandler)
 		bridgeGroup.GET("/claim-proof", b.ClaimProofHandler)
-		bridgeGroup.POST("/sponsor-claim", b.SponsorClaimHandler)
-		bridgeGroup.GET("/sponsored-claim-status", b.GetSponsoredClaimStatusHandler)
+		// bridgeGroup.POST("/sponsor-claim", b.SponsorClaimHandler)
+		// bridgeGroup.GET("/sponsored-claim-status", b.GetSponsoredClaimStatusHandler)
```
Stefan-Ethernal, line 147: Probably it should be enough just updating the `default.go` and disabling the claim sponsor, we don't need to remove the implemenation. WDYT?

temaniarpit27: if someone messes up config, i dont want to take that chance
it is false by default though

PR #570, opened 2025-06-02
refactor: declutter etherman logic

### LINKED ISSUE #574: Declutter etherman logic and aggkit configuration
The etherman component has code that is deprecated, as it was used in the past by the zkevm-node, cdk validium node and cdk, however it is not used by the aggkit, so it is safe to remove it.

Also remove the obsolete configuration parameters.

Stefan-Ethernal:
## Description
This PR refactors Etherman implementation by removing deprecated configuration parameters, centralizing Etherman client initialization, and replacing direct `etherman` references with `aggkittypes` equivalents.

### Deprecated configuration params
- `[Common.IsValidiumMode]`
- `[Common.ContractVersions]`
- `[Common.Translator]`
- `[Etherman]` (only top level configuration section, it is still actual for the `EthTxManager`)
- `[NetworkConfig.L1]` -> `[L1NetworkConfig]`
- `[NetworkConfig.L1.PolAddr]` -> `[L1NetworkConfig.POLTokenAddr]`
- `[NetworkConfig.L1.ZkEVMAddr]` -> `[L1NetworkConfig.RollupAddr]`

Fixes #574


joanestebanr (commented):
I think that NetworkConfig.L1 can be totally removed instead of renamed. 
There are a global vars `[L1Config]` and `[L2Config]` that can be used to set the configuration, but the struct `NetworkConfig` I think that only etherman use it

joanestebanr:
-  I think that it can be delete the whole section `[Common]`

Stefan-Ethernal:
> * I think that it can be delete the whole section `[Common]`

I don't understand, those two config params are used:
- [NetworkID](https://github.com/agglayer/aggkit/blob/434d8e906bfb7daf0637ac7589f772410b167322/cmd/run.go#L111)
- [L2RPC](https://github.com/agglayer/aggkit/blob/434d8e906bfb7daf0637ac7589f772410b167322/cmd/run.go#L65).

Ok we have the following vars, but it is unclear to me how to reference them from the code (namely `run.go`):
- [NetworkID](https://github.com/agglayer/aggkit/blob/434d8e906bfb7daf0637ac7589f772410b167322/config/default.go#L14)
- [L2RPC](https://github.com/agglayer/aggkit/blob/434d8e906bfb7daf0637ac7589f772410b167322/config/default.go#L42)

Stefan-Ethernal:
> I think that NetworkConfig.L1 can be totally removed instead of renamed. There are a global vars `[L1Config]` and `[L2Config]` that can be used to set the configuration, but the struct `NetworkConfig` I think that only etherman use it

Again unclear to me how to use the variables in the default config to resolve the config?

cmd/run.go
```diff
@@ -75,21 +77,32 @@ func start(cliCtx *cli.Context) error {
 		}
 	}()
 
-	rollupID := getRollUpIDIfNeeded(components, cfg.NetworkConfig.L1Config, l1Client)
+	ethermanClient, err := etherman.NewClient(cfg.L1NetworkConfig,
```
goran-ethernal: Just check if this can be instantiated only if needed.

temaniarpit27: +1

temaniarpit27 (approved):
💯

cmd/run.go
Stefan-Ethernal: Fixed https://github.com/agglayer/aggkit/pull/570/commits/91e0525c465ee60e4df738f51680f4b188c812c5

goran-ethernal (approved):
LGTM 💯

vcastellm (approved):
LGTM 💯

cmd/run.go
```diff
@@ -74,20 +76,24 @@ func start(cliCtx *cli.Context) error {
 		}
 	}()
 
-	rollupID := getRollUpIDIfNeeded(components, cfg.NetworkConfig.L1Config, l1Client)
+	ethermanClient, err := initEthermanClient(cfg.L1NetworkConfig, components)
```
joanestebanr: It looks like the `ethermanClient` is only used to get the `RollupID.`. Before this rollupid was obtanied with the function `getRollUpIDIfNeeded`. My point is that maybe we can delete totally the etherman code, WDYT?

joanestebanr:
> > * I think that it can be delete the whole section `[Common]`
> 
> I don't understand, those two config params are used:
> 
> * [NetworkID](https://github.com/agglayer/aggkit/blob/434d8e906bfb7daf0637ac7589f772410b167322/cmd/run.go#L111)
> * [L2RPC](https://github.com/agglayer/aggkit/blob/434d8e906bfb7daf0637ac7589f772410b167322/cmd/run.go#L65).
> 
> Ok we have the following vars, but it is unclear to me how to reference them from the code (namely `run.go`):
> 
> * [NetworkID](https://github.com/agglayer/aggkit/blob/434d8e906bfb7daf0637ac7589f772410b167322/config/default.go#L14)
> * [L2RPC](https://github.com/agglayer/aggkit/blob/434d8e906bfb7daf0637ac7589f772410b167322/config/default.go#L42)

I see your point, my idea was to duplicate the configuration into the specific component. But for URL it's not possible because it's a common value (just create once)

joanestebanr:
Why renaming `[NetworkConfig.L1]` to  `[L1NetworkConfig]` it's the same, but just add friction to the user because make the previous configfile incompatible, WDTY?

Stefan-Ethernal:
> Why renaming `[NetworkConfig.L1]` to `[L1NetworkConfig]` it's the same, but just add friction to the user because make the previous configfile incompatible, WDTY?

Once `state.Genesis` was removed from the `NetworkConfig`, this was the resulting struct. So it didn't make sense to have struct that only wraps another struct.

```go
// NetworkConfig is the configuration struct for the different environments
type NetworkConfig struct {
	// L1: Configuration related to L1
	L1Config ethermanconfig.L1Config `mapstructure:"L1"`
}
```

That's why I have decided to clean it up and "promote" the `ethermanconfig.L1Config` into `L1NetworkConfig`. So it is not only because of renames, but removes unnecessary struct.

Notice that since we are removing fields from configuration, we are anyway going to break backward compatibility (since those parameters needs to be removed from the current config), and therefore it is a good moment to clean up this as well.

If you still have strong opinion why we should keep it the way it was `NetworkConfig.L1`, let me know and I'll revert it back.

cmd/run.go
Stefan-Ethernal: I don't have a strong opinion, just thinking out loud:
in case we remove the etherman, we still need the [logic to retrieve the l2 chain id](https://github.com/agglayer/aggkit/blob/79d81c4724bb1b12857c4bb6d6b0dd4cc831006c/etherman/etherman.go#L79-L91).

Let me figure out if there is a better approach if we can do something better, but seems to me that it is better to have this logic encapsulated within some struct.

joanestebanr: Maybe we can set directly to the function ` getRollUpIDIfNeeded` https://github.com/agglayer/aggkit/blob/05c02973fe2855fd68ca0bd3de5ab7e56a55a745/cmd/run.go#L415

Basically it's just a call to a contract.
No strong opinion either... it's just to reduce a bit more the code

joanestebanr:
> > Why renaming `[NetworkConfig.L1]` to `[L1NetworkConfig]` it's the same, but just add friction to the user because make the previous configfile incompatible, WDTY?
> 
> Once `state.Genesis` was removed from the `NetworkConfig`, this was the resulting struct. So it didn't make sense to have struct that only wraps another struct.
> 
> ```go
> // NetworkConfig is the configuration struct for the different environments
> type NetworkConfig struct {
> 	// L1: Configuration related to L1
> 	L1Config ethermanconfig.L1Config `mapstructure:"L1"`
> }
> ```
> 
> That's why I have decided to clean it up and "promote" the `ethermanconfig.L1Config` into `L1NetworkConfig`. So it is not only because of renames, but removes unnecessary struct.
> 
> Notice that since we are removing fields from configuration, we are anyway going to break backward compatibility (since those parameters needs to be removed from the current config), and therefore it is a good moment to clean up this as well.
> 
> If you still have strong opinion why we should keep it the way it was `NetworkConfig.L1`, let me know and I'll revert it back.

No strong opinion, my only concern is breaking configfile compatibility... it's a trade off.....

cmd/run.go
Stefan-Ethernal: I'll leave the `etherman` and even reuse it in the [lerDataQuerier](https://github.com/agglayer/aggkit/blob/a69bec63a7fa3cf18f4d69f4ed02e39a41b15933/aggsender/query/ler_query.go#L50-L66). With that having in mind, I think it makes sense to have the etherman, in order not to duplicate the logic.

Issue #559, opened 2025-06-02
L1 → L2_C (native/WETH) → L2_A


rachit77:
L1 → L2_C (native/WETH) → L2_A

Native token or WETH bridge to L2_C
Claim
Bridge from L2_C to L2_A
Claim

PR #583, opened 2025-06-04
feat: add testcase

temaniarpit27:
#559


.github/workflows/test-e2e.yml
```diff
@@ -154,16 +160,16 @@ jobs:
...
-      agglayer-e2e-ref: 44083e2c2d2a5e84164edbbda0128253f56a6f55
+      agglayer-e2e-ref: 7c81bd4a91818f8ed713601e0b9b6ecfd2dd539e
       kurtosis-cdk-enclave-name: aggkit
       aggsender-find-imported-bridge-artifact: aggsender_find_imported_bridge
-      kurtosis-cdk-args-1: ${{ needs.read-aggkit-args.outputs.kurtosis-cdk-args-1 }}
-      kurtosis-cdk-args-2: ${{ needs.read-aggkit-args.outputs.kurtosis-cdk-args-2 }}
-      kurtosis-cdk-args-3: ${{ needs.read-aggkit-args.outputs.kurtosis-cdk-args-3 }}
+      kurtosis-cdk-args-1: ${{ needs.read-aggkit-args.outputs.kurtosis-cdk-args-3 }}
+      kurtosis-cdk-args-2: ${{ needs.read-aggkit-args.outputs.kurtosis-cdk-args-4 }}
+      kurtosis-cdk-args-3: ${{ needs.read-aggkit-args.outputs.kurtosis-cdk-args-5 }}
```
Stefan-Ethernal, line 172: Couldn't we reuse the ones for 1st and 2nd chain in 2 networks setup, and just amend the 3rd args (instead of creating 4th and 5th args file)? What is the difference there, I have noticed `gasTokenEnabled` differ (and I would put it the same for all the l2 networks if feasible), but are there any others?

.github/test_e2e_multi_chains_args_4.json
```diff
@@ -0,0 +1,27 @@
...
+  },
+  "args": {
+    "verbosity": "debug",
+    "deployment_suffix": "-002",
+    "zkevm_rollup_chain_id": 20202,
+    "zkevm_rollup_id": 2,
+    "cdk_node_image": "aggkit:local",
+    "agglayer_image": "ghcr.io/agglayer/agglayer:0.3.0-rc.21",
+    "cdk_erigon_node_image": "hermeznetwork/cdk-erigon:v2.61.19",
+    "zkevm_contracts_image": "jhkimqd/zkevm-contracts:v10.0.0-rc.6-fork.12",
```
Stefan-Ethernal: wdyt (and use the same everywhere)?
```suggestion
    "zkevm_contracts_image": "jhkimqd/zkevm-contracts:v10.1.0-rc.5-fork.12",
```

Stefan-Ethernal (approved):
LGTM, however wondering if we can reuse the args-1 and args-2 to act as args for first two l2 chains in 3 chains setup networks.

.github/workflows/test-e2e.yml
temaniarpit27: only gas token is different
the testcases given by laia are such that we have different gas tokens on different networks
1st and 2nd have enabled, 3rd disabled

temaniarpit27: couldn't find a good way to use 1 and 2 instead of 3 and 4 and make only few fields different.

rachit77: I agree with Stefan on that, if possible, we should consider using a unified template to render all configs maybe

rachit77 (approved):
💯

.github/workflows/test-e2e.yml
temaniarpit27: https://github.com/agglayer/aggkit/issues/616
created an issue to tackle it
need some thinking on it

PR #585, opened 2025-06-04
fix: return error when limiting cert size to empty certificate

goran-ethernal:
## Description

This PR is a fix of an audit report, where if we do not allow an empty certificate (no bridge exits), and we limit cert size all the way to a block where we do not have any bridge exits, we need to return an error.

Fixes # (issue)

### DIFF
--- a/aggsender/flows/flow_base.go
+++ b/aggsender/flows/flow_base.go
@@ -100,9 +100,9 @@ func (f *baseFlow) limitCertSize(
 
 	for {
 		if currentCert.NumberOfBridges() == 0 && !allowEmptyCert {
-			f.log.Warnf("Minimum certificate size reached. Estimated size: %d > max size: %d",
-				currentCert.EstimatedSize(), f.maxCertSize)
-			return currentCert, nil
+			return nil, fmt.Errorf("error on reducing the certificate size. "+
+				"No bridge exits found in range from: %d, to: %d and empty certificate is not allowed",
+				currentCert.FromBlock, currentCert.ToBlock)
 		}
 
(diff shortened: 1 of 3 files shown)


temaniarpit27 (approved):
💯

Stefan-Ethernal (approved):
:100:

joanestebanr (commented):
Maybe make sense to return the previous cert with `NumberOfBridges()>0`. The point is that if we return an error in that situation the aggsender is going to get stuck. If we return the previous calculated cert maybe it fit the maximum size (because this limit is based on an estimatition)... WDTY?

For sure, if there are no previous cert with bridge>0 it require to return an error

goran-ethernal:
> Maybe make sense to return the previous cert with `NumberOfBridges()>0`. The point is that if we return an error in that situation the aggsender is going to get stuck. If we return the previous calculated cert maybe it fit the maximum size (because this limit is based on an estimatition)... WDTY?
> 
> For sure, if there are no previous cert with bridge>0 it require to return an error

But the previous one does not fit the `maxCertSize`, because if it did fit the max size, this condition would be fulfilled:
```
if f.maxCertSize == 0 || currentCert.EstimatedSize() <= f.maxCertSize {
	return currentCert, nil
}
```
I agree its an estimation, but if we return the previously limited size, then I don't see the point on estimating it at all. To me, if someone is trying to send an overly large certificate, without tweaking the max size to allow this, it makes sense that the aggsender keeps failing to send it if resize it to the point where it has no bridge exits. This can happen either in an upgrade, or if aggsender was down for quite some time. In both cases I think we accept that a manual step is needed.

joanestebanr (approved):

PR #596, opened 2025-06-06
feat: `hermez` to `pp` upgrade patch (DO NOT MERGE)

goran-ethernal:
## Description

This PR is just a quick patch for `Hermez` to `PP` upgrade, where `aggsender` will be provided with an end block until which it will create one large certificate needed for the upgrade, and after that certificate gets settled on `agglayer`, the `aggsender` will halt.

After this `upgrade` certificate is settled and this patched `aggsender` is halted, the regular image for `aggsender` will be started, and the PP network begins.

**IMPORTANT NOTE**: DO NOT MERGE THIS. THIS IS JUST A TEMPORARY SOLUTION FOR UPGRADE. THIS WILL NOT BE MERGED, BUT DISCARDED AFTER WE HAVE A FINAL SOLUTION.

Fixes # (issue)

### DIFF
--- a/aggsender/aggsender.go
+++ b/aggsender/aggsender.go
@@ -147,7 +147,90 @@ func (a *AggSender) Start(ctx context.Context) {
 	if err := a.flow.CheckInitialStatus(ctx); err != nil {
 		a.log.Panicf("error checking flow Initial Status: %v", err)
 	}
+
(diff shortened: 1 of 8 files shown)


goran-ethernal:
Closing this, since it was just a quick temp feature. We eventually implemented it in a proper manner: https://github.com/agglayer/aggkit/pull/648.

PR #604, opened 2025-06-09
feat: direct logs to stdout

### LINKED ISSUE #603: fix local run logging
sometimes the logs prints on stdout, sometimes in file
i think its better to print on stdout

rachit77:
Fixes #603

### DIFF
--- a/test/run-local-e2e.sh
+++ b/test/run-local-e2e.sh
@@ -5,8 +5,8 @@ GREEN='\033[0;32m'
 RED='\033[0;31m'
 NC='\033[0m'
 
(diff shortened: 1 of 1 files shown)


temaniarpit27 (approved):
💯

Stefan-Ethernal (approved):
Does it make sense to have an option storing logs to the file or on stdout?

rachit77:
> Does it make sense to have an option storing logs to the file or on stdout?

Sure, I can add that option if needed. However, I prefer keeping the script simple by default. If someone wants logs in a file, they can easily redirect stdout when running the script. 
wdyt?

PR #615, opened 2025-06-10
feat: use start LER from contract

### LINKED ISSUE #21: aggsender/bridgesync: first LER must be request to smart contract
blocked by AL, needs clarification by protocol

goran-ethernal:
### Description

This PR implements the logic of querying the start `LocalExitRoot` from the `RollupManager` contract for given network.

This start LER will be used as the starting LER in the first certificate on the network.

### Configuration Changes ⚙️
- introduced `RollupManagerAddr` in `aggsender` config. It is an address to the `RollupManager` contract on `L1`.
- introduce `L1GenesisBlock` in `aggsender` config. It is the block number where the `polygonZKEVM` (rollup) smc was deployed on `L1`

Fixes #21


config/default.go
```diff
@@ -217,7 +217,8 @@ PolAddr = "{{L1Config.polTokenAddress}}"
 ZkEVMAddr = "{{L1Config.polygonZkEVMAddress}}"
 RollupManagerAddr = "{{L1Config.polygonRollupManagerAddress}}"
 GlobalExitRootManagerAddr = "{{L1Config.polygonZkEVMGlobalExitRootAddress}}"
-
+[NetworkConfig.Genesis]
+BlockNumber = {{genesisBlockNumber}}
```
joanestebanr: I think that make more sense to set this config to aggsender that is the one that use it

etherman/etherman.go
```diff
@@ -103,6 +103,34 @@ func GetRollupID(l1Config config.L1Config, rollupAddr common.Address, ethClient
 	return rollupID, nil
 }
 
+func GetLastLocalExitRoot(
```
joanestebanr: This function maybe make more sense to be in aggsender, WDTY?

cmd/run.go
```diff
@@ -75,6 +76,7 @@ func start(cliCtx *cli.Context) error {
 	}()
 
 	rollupID := getRollUpIDIfNeeded(components, cfg.NetworkConfig.L1Config, l1Client)
+	lastLER := getRollupLastExitRootIfNeeded(components, cfg.NetworkConfig, l1Client, rollupID)
```
joanestebanr: maybe is better to require this value when is needed (from aggsender) because is only require in the first certificate. In this case you are asking always that is not required

joanestebanr (commented):
In fact, I'm thinking that you don't need the genesis block. You can ask for the LER in the starting block of the certificate, WDTY?

cmd/run.go
goran-ethernal: Hmm it might be a good place, but, how often do we start and restart aggsender. Plus, this makes the testing easier if aggsender gets this outside of its initialization.

goran-ethernal:
> In fact, I'm thinking that you don't need the genesis block. You can ask for the LER in the starting block of the certificate, WDTY?

You can not, because this is saved on the rollup manager contract on L1. We need some starting block on L1 to ask for it, that is why we use genesis block.

Stefan-Ethernal (commented):
It probably makes sense to base this PR on top of the #570, because of refactored etherman logic. WDYT?

aggsender/flows/flow_base.go
```diff
@@ -404,14 +410,22 @@ func (f *baseFlow) getImportedBridgeExits(
 	return importedBridgeExits, nil
 }
 
+// getStartLER returns the last local exit root (LER) based on the configuration
+func (f *baseFlow) getStartLER() common.Hash {
+	if f.cfg.StartLER == (common.Hash{}) {
+		return zeroLER
```
Stefan-Ethernal: I was about to propose simplification, but I was thinking that the `zeroLER` is `0x0`, since it is not the case, wdyt about renaming it to `emptyLER`?

etherman/etherman.go
Stefan-Ethernal: OTOH if this is about to remain in the etherman, consider to make it a method instead (in order to avoid instantiation of contracts etc.)

As a heads up, the `Contracts` reference is removed in #570, and only ref to `RollupManager` SC is kept.

Stefan-Ethernal (approved):

goran-ethernal:
> It probably makes sense to base this PR on top of the #570, because of refactored etherman logic. WDYT?

I will move this from the `etherman` logic, to the `aggsender` package, as @joanestebanr suggested.

aggsender/flows/flow_base.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/615/commits/afe3c3767976d882b5e73275c250464ed1caf1e1

aggsender/flows/flow_base.go
```diff
@@ -404,14 +410,22 @@ func (f *baseFlow) getImportedBridgeExits(
 	return importedBridgeExits, nil
 }
 
+// getStartLER returns the last local exit root (LER) based on the configuration
+func (f *baseFlow) getStartLER() common.Hash {
+	if f.cfg.StartLER == (common.Hash{}) {
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/615/commits/afe3c3767976d882b5e73275c250464ed1caf1e1

cmd/run.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/615/commits/afe3c3767976d882b5e73275c250464ed1caf1e1

etherman/etherman.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/615/commits/afe3c3767976d882b5e73275c250464ed1caf1e1

config/default.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/615/commits/afe3c3767976d882b5e73275c250464ed1caf1e1

temaniarpit27 (approved):

aggsender/config/config.go
```diff
@@ -68,6 +68,10 @@ type Config struct {
 	RequireNoFEPBlockGap bool `mapstructure:"RequireNoFEPBlockGap"`
 	// OptimisticModeConfig is the configuration for optimistic mode (required by FEP mode)
 	OptimisticModeConfig optimistic.Config `mapstructure:"OptimisticModeConfig"`
+	// RollupManagerAddr is the address of the RollupManager contract on L1
+	RollupManagerAddr ethCommon.Address `mapstructure:"RollupManagerAddr"`
+	// RollupCreationBlock is the block number when the rollup was created on L1
+	RollupCreationBlock uint64 `mapstructure:"RollupCreationBlock"`
```
joanestebanr: Maybe adding L1 to the name is going to make it more clear... WDTY?

joanestebanr (approved):

aggsender/config/config.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/615/commits/b4b43b08daf1ecebaa33e9fe7d5270faa8f95984

PR #617, opened 2025-06-10
fix: false warning log

### LINKED ISSUE #164: aggsender: false warning if aggsender block finality it's configured as  `finalized`
# Bug Report

The object `BlockNotifierPolling` have a warning if there are a jump of block >=1 and that always happen with that finality

goran-ethernal:
## Description

`block_notifier_polling` logs a false warning log when block finality is configured as Finalized or Safe for L1.

If block finality is `finalized` or `safe`, almost in all cases, the last seen finalized or safe block is behind of what is marked as the new `finalized` or `safe` block, since this is how that finality should work. Its not sequential, but if block 10 is marked as finalized, the next finalized block will be 15 for example. In this case, the warning for missed blocks can be misleading, and should not be reported.

Fixes #164

### DIFF
--- a/aggsender/block_notifier_polling_test.go
+++ b/aggsender/block_notifier_polling_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/agglayer/aggkit/aggsender/mocks"
 	aggsendertypes "github.com/agglayer/aggkit/aggsender/types"
+	aggkitcommon "github.com/agglayer/aggkit/common"
 	"github.com/agglayer/aggkit/etherman"
 	"github.com/agglayer/aggkit/log"
 	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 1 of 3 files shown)


aggsender/block_notifier_polling_test.go
```diff
@@ -45,9 +46,11 @@ func TestBlockNotifierPollingStep(t *testing.T) {
 	tests := []struct {
 		name                      string
 		previousStatus            *blockNotifierPollingInternalStatus
+		cfg                       *ConfigBlockNotifierPolling
 		HeaderByNumberError       bool
 		HeaderByNumberErrorNumber uint64
```
Stefan-Ethernal: Why are these fields public?


```suggestion
		headerByNumberError       bool
		headerByNumberErrorNumber uint64
```

Stefan-Ethernal (approved):
:100:

aggsender/block_notifier_polling_test.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/617/commits/43247a30e199ed5b7e5dc715789ec4837a04c980

joanestebanr (approved):

PR #621, opened 2025-06-11
fix: multiline swagger descriptions

### LINKED ISSUE #622: Fix multiline swagger descriptions

Stefan-Ethernal:
## Description

Fix the multiline comments in the Swagger docs for Bridge service. In cases there were multiple lines, only first one was rendered.

Fixes #622

### DIFF
--- a/bridgeservice/docs/docs.go
+++ b/bridgeservice/docs/docs.go
@@ -125,7 +125,7 @@ const docTemplate = `{
         },
         "/claim-proof": {
             "get": {
-                "description": "Returns the Merkle proofs (local and rollup exit root) and",
+                "description": "Returns the Merkle proofs (local and rollup exit root) and\nthe corresponding L1 info tree leaf needed to verify a claim.",
                 "produces": [
                     "application/json"
                 ],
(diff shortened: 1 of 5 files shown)


bridgeservice/docs/docs.go
```diff
@@ -125,7 +125,7 @@ const docTemplate = `{
         },
         "/claim-proof": {
             "get": {
-                "description": "Returns the Merkle proofs (local and rollup exit root) and",
+                "description": "Returns the Merkle proofs (local and rollup exit root) and\nthe corresponding L1 info tree leaf needed to verify a claim.",
```
temaniarpit27, line 128: is there some automated way of multiline instead of putting `/n`?

Stefan-Ethernal: Well, this is what swaggo generated. FWIW, it is rendered properly now. 

image

Seems that the common approach is to have a one liner description, but then linter complains about line being too long...

temaniarpit27 (approved):
💯

rachit77 (approved):

PR #623, opened 2025-06-11
fix: grpc retry policies validation of request timeout

### LINKED ISSUE #624: GRPC retry request timeout formula is incorrect
Validation of request timeout when grpc retries are in place is incorrect.

The current formula:
```go
minTimeout = initialBackoff - (1 - backoffMultiplier ^ maxAttempts) / (1 - backoffMultiplier)
```

Instead it should just sum all the backoffs (skip the first one, because for the first call there is no delay), multiplied by backoff multiplier and that should be the minimum request timeout that needs to be configured.

```go
for i := 1; i < N; i++ {
    delay := d0 * math.Pow(r, float64(i-1))
    if delay > dMax {
        delay = dMax
    }
    totalBackoff += delay
}
```

Stefan-Ethernal:
## Description

This PR fixes the validation of minimum request timeout formula which calculates the least required request timeout to accommodate for all the retries.

Fixes #624

### DIFF
--- a/config/default.go
+++ b/config/default.go
@@ -230,11 +230,6 @@ RequireNoFEPBlockGap = true
 		MinConnectTimeout = "5s"
 		RequestTimeout = "{{GenerateAggchainProofTimeout}}"
 		UseTLS = false
-		[AggSender.AggkitProverClient.Retry]
-			InitialBackoff = "1s"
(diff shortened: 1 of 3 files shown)


rachit77 (approved):
💯

PR #637, opened 2025-06-12
fix: add where clause to the get latest info until block

### LINKED ISSUE #639: Filter until block number in l1 info tree syncer
`GetLatestInfoUntilBlock` should include the filtering until block number, so that function name reflects what it is really doing.

Currently it is always returning the latest l1 info tree leaf.

Stefan-Ethernal:
## Description

Add where clause to include filtering until provided block number in the `GetLatestInfoUntilBlock` function, so that it reflects the name.

Fixes #639


temaniarpit27 (approved):
💯

rachit77 (approved):

PR #647, opened 2025-06-13
fix: check if there are bridge transactions in gap

### LINKED ISSUE #498: feat: upgrade PP to FEP: check that block gap  doesn't have any relevant data that is going to be missed
# Feature Request

The idea is that if we upgrade from PP to FEP, the blocks that we are going to skip doesn't contain anything relevant.
 - It detects the blocks between: if there are relevant bridges or claim and `RequireNoFEPBlockGap` is a error if not is a warning

goran-ethernal:
### Description

This PR makes the gap check dependent on the number of bridge transactions in the gap as well.

If the gap between two certificates has no bridge transactions, we can allow it, if not, the `aggsender` will keep reporting errors on certificate building or on startup if it detects it has a gap.

### Configuration changes
- `DelayBeetweenRetries` config parameter is renamed to `DelayBetweenRetries` (because of the typo).

Fixes #498


aggsender/config/config.go
```diff
@@ -63,9 +63,6 @@ type Config struct {
 	// RequireStorageContentCompatibility is true it's mandatory that data stored in the database
 	// is compatible with the running environment
 	RequireStorageContentCompatibility bool `mapstructure:"RequireStorageContentCompatibility"`
-	// RequireNoFEPBlockGap is true if the AggSender should not accept a gap between
-	// lastBlock from lastCertificate and first block of FEP
-	RequireNoFEPBlockGap bool `mapstructure:"RequireNoFEPBlockGap"`
```
joanestebanr: It must keep the flag to skip the sanity check of the gap

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -103,50 +80,13 @@ func NewAggchainProverFlow(
...
-	lastSentCertificate, err := a.storage.GetLastSentCertificateHeader()
-	if err != nil {
-		return fmt.Errorf("aggchainProverFlow - error getting last sent certificate: %w", err)
-	}
-	return a.sanityCheckNoBlockGaps(lastSentCertificate)
-}
-
-// sanityCheckNoBlockGaps checks that there are no gaps in the block range for next certificate
-// #436. Don't allow gaps updating from PP to FEP
-func (a *AggchainProverFlow) sanityCheckNoBlockGaps(lastSentCertificate *types.CertificateHeader) error {
```
joanestebanr, line 138: Why do you delete these function? `CheckInitialStatus` and `sanityCheckNoBlockGaps`. The idea is that each flow have his own sanity check here. 
You still have to check the  Gap but instead of just a block range you have to check the content inside the gap

joanestebanr (commented):
The PR  #516 have a early implementation of that

aggsender/config/config.go
goran-ethernal: To me this check should be dynamic without the need for a flag. If there are bridge transactions in the gap we can not allow it to continue since proofs will not be verified even if we skip it. If there is no bridge transactions, then we can allow it. This way, there is no need to have a flag.

We can maybe keep it if someone wants to forbid gaps even when there are no bridge transactions in them. But I do not see the use case for this.

aggsender/flows/flow_aggchain_prover.go
goran-ethernal: This sanity check makes sense to be for every certificate, not just on start. That is why, I removed this function, and just add it to the place where other sanity checks on certificates are performed.

aggsender/query/bridge_query.go
```diff
@@ -87,3 +116,30 @@ func (b *bridgeDataQuerier) GetLastProcessedBlock(ctx context.Context) (uint64,
 func (b *bridgeDataQuerier) OriginNetwork() uint32 {
 	return b.originNetwork
 }
+
+func (b *bridgeDataQuerier) waitForSyncerToCatchUp(ctx context.Context, block uint64) error {
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/647/commits/88ff8df9e4c48595443a1688aa00fe00a0f18ca2

config/default.go
```diff
@@ -173,7 +173,7 @@ AggsenderPrivateKey = {{AggsenderPrivateKey}}
 BlockFinality = "LatestBlock"
 EpochNotificationPercentage = 50
 MaxRetriesStoreCertificate = 3
-DelayBeetweenRetries = "60s"
+DelayBeetweenRetries = "5s"
```
Stefan-Ethernal: There is a typo in this parameter, it should read the `DelayBetweenRetries`. Maybe it is a good time to change this (and deprecate the wrong naming), as v0.5.0 delivers config changes anyway. WDYT?

config/config.go
```diff
@@ -61,7 +61,8 @@ const (
 	ethermanDeprecatedHint                = "Etherman config is deprecated, remove it from configuration"
 	networkConfigDeprecatedHint           = "NetworkConfig is deprecated, use L1NetworkConfig instead"
 	l1NetworkConfigUsePolTokenAddrHint    = "Use L1NetworkConfig.POLTokenAddr instead"
-	l1NetworkConfigUseRollupAddrrHint     = "Use L1NetworkConfig.RollupAddr instead"
+	l1NetworkConfigUseRollupAddrHint      = "Use L1NetworkConfig.RollupAddr instead"
+	aggsenderRequireNoFEPBlockGapHint     = "AggSender.RequireNoFEPBlockGap is deprecated and no longer used"
```
Stefan-Ethernal: WDYT?
```suggestion
	aggsenderRequireNoFEPBlockGapHint     = "AggSender.RequireNoFEPBlockGap is deprecated, remove it from configuration"
```

aggsender/flows/flow_aggchain_prover.go
joanestebanr: Before your change it was done at start and for each certificate (both cases). The reason to do it at start is for a fast-fail check, basically if when we start the aggsender and we known that we can't create a certificate doesn't make sense to wait until reach an epoch

aggsender/config/config.go
joanestebanr: Ok, if you can't forseen any case that we would need to override this check it's ok. but let's remember that the first check we put on the no gap check didn't seem like there was any reason to skip it either until we needed to.

aggsender/flows/flow_aggchain_prover.go
goran-ethernal: I will bring the check on startup back.

aggsender/flows/flow_base.go
```diff
@@ -475,6 +486,59 @@ func (f *baseFlow) verifyClaimGERs(claims []bridgesync.Claim) error {
...
+
+	nextBlockRange := types.NewBlockRange(buildParams.FromBlock, buildParams.ToBlock)
+	lastBlockRange := types.NewBlockRange(lastSettledFromBlock, lastSettledToBlock)
+
+	// case 2: is a new cert but is not contiguous to previous one
+	gap := nextBlockRange.Gap(lastBlockRange)
+	if gap.IsEmpty() {
+		return nil
+	}
+	bridgeDataInTheGap, claimDataInTheGap, err := f.l2BridgeQuerier.NumOfBridgeTransactions(
```
joanestebanr: Why don't use directy `GetBridgesAndClaims` function instead of creating a new one? I see that is not for optimization because the underlaying function is calling to `GetBridgesAndClaims`.
Another point is that you add a `waitForSyncerToCatchUp`, I think that make more sense to responde an error and retry like any other error. It's very uncommon that you reach this point and the bridge doesn't catch up the 'toBlock', 

Also, it's a bit disturbing that a call like `verifyBlockRangeGaps` can block the execution of the code, another point is that maybe require to set a timeout in the context to avoid to be there forever

WDTY?

goran-ethernal: I agree with you when it is a regular certificate per certificate case. But, when the case is in upgrade from some PP block, to start FEP block, then we need to wait for the syncer to catch up, because, they stop the aggsender in the upgrade process, so at least on the startup, we need to check if there are bridge transactions in the last PP to first FEP blocks, since aggsender was down, and the syncers need to catch up.

Let me fix this, and use this only for the startup check.

aggsender/config/config.go
goran-ethernal: It was because they had empty blocks between the last settled certificate and `startL2Block` for FEP network. We kept returning errors in this case, and we needed a way to ignore this empty gap.

Maybe it makes sense to keep this field if someone wants to force even empty blocks to be settled in the FEP network. WDYT?

aggsender/flows/flow_base.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/647/commits/32ed3efddad5229533804927f45560615501b979

aggsender/flows/flow_aggchain_prover.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/647/commits/32ed3efddad5229533804927f45560615501b979

aggsender/config/config.go
goran-ethernal: I have kept the config parameter so we can configure if we want empty gaps or not for FEP:

https://github.com/agglayer/aggkit/pull/647/commits/07efc0269aa32b4b77c9d7b5a6f190314625ddb2

aggsender/flows/flow_base.go
joanestebanr: I still see that a bit strange, maybe you can publish the new  function in `bridge_query` `WaitForSyncerToCatchUp` and you can call in the initial check if you need, instead of passing a boolean thought several functions.  Also, you can take advantage of this new function in another calls or situations that you required it.

By other hand,  I think  that you can avoid adding the  function `NumOfBridgeTransactions` using the old one `GetBridgesAndClaims`.

WDTY?

joanestebanr (commented):
See next comment: https://github.com/agglayer/aggkit/pull/647#discussion_r2167054965

aggsender/config/config.go
joanestebanr: I see your point, maybe we can have a huddle to discuss it, WDTY?

aggsender/flows/flow_base.go
goran-ethernal: Ok, fair point. I will do as suggested.

goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/647/commits/3a917b847d4edbdb6eb44a51276489db106bac6e

config/config.go
goran-ethernal: I've brought back this parameter, so this is removed now from code.

config/default.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/647/commits/b35fe8b3d3143b38516ecd131c442b67991a15bf

aggsender/query/bridge_query.go
```diff
@@ -87,3 +96,40 @@ func (b *bridgeDataQuerier) GetLastProcessedBlock(ctx context.Context) (uint64,
...
+		if err != nil {
+			return fmt.Errorf("bridgeDataQuerier - error getting last processed block: %w", err)
+		}
+
+		if lastProcessedBlock >= block {
+			b.log.Infof("bridgeDataQuerier - L2 syncer caught up to block: %d", block)
+			return nil
+		}
+
+		b.log.Debugf("bridgeDataQuerier - waiting for L2 syncer to catch up to block: %d, current last processed block: %d",
```
joanestebanr: I would set this as Info, because if not we are not going to see anything in logs and it could look like the aggsender is halted

config/default.go
```diff
@@ -173,7 +173,7 @@ AggsenderPrivateKey = {{AggsenderPrivateKey}}
 BlockFinality = "LatestBlock"
 EpochNotificationPercentage = 50
 MaxRetriesStoreCertificate = 3
-DelayBeetweenRetries = "60s"
+DelayBetweenRetries = "5s"
```
joanestebanr: 5s? is not too fast?

goran-ethernal: I think this is reasonable in the terms of code. Its not too fast too have too many retries.

aggsender/query/bridge_query.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/647/commits/98582cf7c04e5139b92a8ae840ad9a99570a318c

config/default.go
goran-ethernal: I increased the default to 30s: https://github.com/agglayer/aggkit/pull/647/commits/377e93e65bb6515f51a142d73abc3ca8a6d23670

joanestebanr (approved):

Stefan-Ethernal (approved):

PR #619, opened 2025-06-10
fix: make bridge exit limitation configurable

goran-ethernal:
### Description

`Agglayer` introduced certificate height in the pessimistic proof program, so they do not require one bridge exit to be present in the `PP` certificate.

This PR makes this limitation configurable on `Aggsender`, where it is turned off by default, meaning, PP certificate can be sent with only claims.

### Configuration Changes ⚙️
 - introduced `RequireOneBridgeInPPCertificate` `bool` field to the `aggsender` config which is a flag to force the `aggSender` to have at least one bridge exit in the `Pessimistic Proof` certificates. By default it is `true`.

Fixes #618

PR #648, opened 2025-06-13
feat: add MaxL2BlockNumber config param

joanestebanr:
## Description

It required a way to generate certificate until reach a L2 blockNumber. This is useful, for instance, if there are a fork  in L2 and the we need to verify up to a block to be able to update after that.
The `aggsender`  stop depending on  `StopOnFinishedSendingAllCertificates` when reach this final state, also it shows a log like `"all certs send, no more certificates can be sent`. 

## Configuration
```
[AggSender]
MaxL2BlockNumber = 0 
StopOnFinishedSendingAllCertificates = false
```
If `MaxL2BlockNumber`  is equal to **0** means disabled. If not this is the maximum block that can be included in a certificate. 

- Updated go to 1.24.4 due vulnerabitlies


config/default.go
```diff
@@ -244,6 +244,7 @@ RequireStorageContentCompatibility = {{RequireStorageContentCompatibility}}
 UseAgglayerTLS = false
 UseAggkitProverTLS = false
 RequireNoFEPBlockGap = true
+LastL2BlockInCertificate = 0
```
goran-ethernal: Rename it to match the field you added in aggsender config.

aggsender/flows/flow_pp.go
```diff
@@ -27,12 +28,20 @@ func NewPPFlow(log types.Logger,
 	storage db.AggSenderStorage,
 	l1InfoTreeQuerier types.L1InfoTreeDataQuerier,
 	l2BridgeQuerier types.BridgeQuerier,
-	signer signertypes.Signer) *PPFlow {
+	signer signertypes.Signer,
+	maxL2BlockNumber uint64) *PPFlow {
+	feature := &FeatureMaxL2BlockNumber{
```
goran-ethernal: Use the constructor function.

aggsender/flows/flow_pp.go
```diff
@@ -56,6 +65,14 @@ func (p *PPFlow) GetCertificateBuildParams(ctx context.Context) (*types.Certific
 		return nil, err
 	}
 
+	if p.featureMaxL2Block != nil {
+		// If the feature is enabled, we need to adapt the build params
+		buildParams, err = p.featureMaxL2Block.AdaptCertificate(buildParams)
+		if err != nil {
+			return nil, fmt.Errorf("ppFlow - error adapting MaxL2Block certificate: %w", err)
```
goran-ethernal:
```suggestion
			return nil, fmt.Errorf("ppFlow - error adapting certificate to MaxL2Block: %w", err)
```

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -270,8 +297,9 @@ func (a *AggchainProverFlow) verifyBuildParamsAndGenerateProof(
 				lastProvenBlock, buildParams.ToBlock)
 			return nil, nil
 		}
-
-		return nil, fmt.Errorf("aggchainProverFlow - error generating aggchain proof: %w", err)
+		errNew := fmt.Errorf("aggchainProverFlow - error generating aggchain proof: %w", err)
+		a.log.Error(errNew.Error())
```
goran-ethernal: This will be logged in the function that calls this. I think we should remove this log here.

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -236,9 +256,16 @@ func (a *AggchainProverFlow) GetCertificateBuildParams(ctx context.Context) (*ty
...
 			return nil, nil
 		}
-
 		return nil, err
 	}
+	if a.featureMaxL2Block != nil {
+		// If the feature is enabled, we need to adapt the build params
+		buildParams, err = a.featureMaxL2Block.AdaptCertificate(buildParams)
+		if err != nil {
+			return nil, fmt.Errorf("aggchainProverFlow - error adapting MaxL2Block certificate: %w", err)
```
```suggestion
			return nil, fmt.Errorf("aggchainProverFlow - error adapting certificate to MaxL2Block: %w", err)
```

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -204,6 +217,13 @@ func (a *AggchainProverFlow) GetCertificateBuildParams(ctx context.Context) (*ty
 			CreatedAt:           lastSentCert.CreatedAt,
 			CertificateType:     typeCert,
 		}
+		if a.featureMaxL2Block != nil {
+			// If the feature is enabled, we need to adapt the build params
+			buildParams, err = a.featureMaxL2Block.AdaptCertificate(buildParams)
+			if err != nil {
+				return nil, fmt.Errorf("aggchainProverFlow - error adapting MaxL2Block certificate: %w", err)
```
```suggestion
				return nil, fmt.Errorf("aggchainProverFlow - error adapting certificate to MaxL2Block: %w", err)
```

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -95,18 +100,25 @@ func NewAggchainProverFlow(
 	optimisticModeQuerier types.OptimisticModeQuerier,
 	optimisticSigner types.OptimisticSigner,
 ) *AggchainProverFlow {
+	feature := &FeatureMaxL2BlockNumber{
```
goran-ethernal: Use constructor function.

aggsender/flows/feature_maxl2blocknumber.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/agglayer/aggkit/aggsender/types"
+)
+
+var (
+	ErrMaxL2BlockNumberExceededInARetryCert = errors.New("max L2 block number exceeded in a retry certificate")
+	ErrComplete                             = errors.New("all certs send, no more certificates can be sent")
+	ErrBuildParamsIsNil                     = errors.New("buildParams is nil")
+)
+
+type FeatureMaxL2BlockNumber struct {
```
goran-ethernal: To me this makes more sense than naming a feature: FeatureN

```suggestion
type MaxL2BlockNumberLimiter struct {
```

aggsender/flows/feature_maxl2blocknumber.go
```diff
@@ -0,0 +1,122 @@
...
+func (f *FeatureMaxL2BlockNumber) isUpcomingNextRange(
+	fromBlock, toBlock uint64) bool {
+	if !f.IsEnabled() {
+		return false
+	}
+	// This is exactly the upcoming next range after the last sent certificate.
+	// e.g: max= 150
+	// last sent cert: fromBlock= 100, toBlock= 150
+	// upcoming next range: fromBlock= 151, toBlock= 200
+	if fromBlock == f.maxL2BlockNumber+1 && toBlock > f.maxL2BlockNumber {
```
goran-ethernal:
```suggestion
	return fromBlock == f.maxL2BlockNumber+1 && toBlock > f.maxL2BlockNumber
```

aggsender/flows/feature_maxl2blocknumber.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	f.log.Infof("Adapting certificate build params. "+
+		"maxL2BlockNumber: %d, FromBlock: %d, ToBlock: %d",
+		f.maxL2BlockNumber, buildParams.FromBlock, buildParams.ToBlock)
+	if buildParams.IsARetry() && !f.allowToResizeRetryCert {
+		// We have a buildParams that we can't change the range, so it's an error
+		err := fmt.Errorf("can't adapt the retry certificate, "+
+			"the ToBlock %d is greater than the maxL2BlockNumber %d. Err: %w",
+			buildParams.ToBlock, f.maxL2BlockNumber, ErrMaxL2BlockNumberExceededInARetryCert)
+		f.log.Error(err)
```
goran-ethernal: This will be logged as well in the caller. I think we can remove this log.

aggsender/flows/feature_maxl2blocknumber.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func (f *FeatureMaxL2BlockNumber) IsAllowedBlockNumber(toBlock uint64) bool {
+	if !f.IsEnabled() {
+		return false
+	}
+	return toBlock <= f.maxL2BlockNumber
+}
+
+func (f *FeatureMaxL2BlockNumber) isUpcomingNextRange(
```
goran-ethernal:
```suggestion
func (f *FeatureMaxL2BlockNumber) isMaxRangeReached(
```

aggsender/flows/feature_maxl2blocknumber.go
```diff
@@ -0,0 +1,122 @@
...
+		f.log.Warn(err)
+		return nil, err
+	}
+	newBuildParams, err := buildParams.Range(buildParams.FromBlock, f.maxL2BlockNumber)
+	if err != nil {
+		return nil, fmt.Errorf("error adjusting the ToBlock of the certificate  %d -> %d: %w",
+			buildParams.ToBlock, f.maxL2BlockNumber,
+			err)
+	}
+	if f.allowToSendNoBridgesCert && newBuildParams.IsEmpty() {
```
goran-ethernal: In PR #619 , we added a config parameter `RequireOneBridgeInPPCertificate`. We need to include this as well in this function. My suggestion is, rename `allowToSendNoBridgesCert` to `allowToSendEmptyCertificate`, and add also that `RequireOneBridgeInPPCertificate` to the config of this feature, so if the `allowEmptyCertificate == false` (PP certificate), check if it has any bridge if `RequireOneBridgeInPPCertificate` is `true`. If it is not true, return an error. If it is true, then check if it has claims. If it is empty, return an error, or nil as build params.

config/default.go
joanestebanr: good catch!

aggsender/flows/feature_maxl2blocknumber.go
joanestebanr: I prefer not to delegate loggging, it's going to be a double log for that but is better than none.

aggsender/flows/feature_maxl2blocknumber.go
joanestebanr: I prefer the name of `Feature` because it means that is a optional behaviour

aggsender/flows/feature_maxl2blocknumber.go
joanestebanr: The point that is exactly the next range after the last one, I think that your name `isMaxRangeReached` doesn't means that... if you have another suggestion

aggsender/flows/feature_maxl2blocknumber.go
joanestebanr: The current code doesn't have this distintion beetween empty or no bridges. Basically if we allow to send a certificate with no bridges also means that we can send totally empty.  So I'm going to rename the internal field as `requireOneBridgeInCertificate` (without PP because this code is generic) to match the PR #619

I undertand that there no restriction in PP to send a empty certificate, isn't? as you mentioned in the PR agglayer have introduce the height in the pessimistic proof, so is no longer a requirement. So we don't need the `allowEmptyCertificate`  just `RequireOneBridgeInPPCertificate`, isn't? WDTY

aggsender/flows/feature_maxl2blocknumber.go
vcastellm: I agree with Goran, prefixing `Feature` is kind of weird, we don't have any other "feature" prefixed like that

aggsender/aggsender.go
```diff
@@ -159,6 +160,17 @@ func (a *AggSender) checkDBCompatibility(ctx context.Context) {
...
 	}
 }
+func (a *AggSender) checkSendCertificateStopConfidition(err error) {
+	if errors.Is(err, flows.ErrComplete) {
+		a.log.Infof("AggSender reached the end of the certificates to send")
+		if a.cfg.StopOnFinishedSendingAllCertificates {
+			a.log.Info("Stopping AggSender because StopOnFinishedSendingAllCertificates is true")
+			// That is the fastest way to stop the process. Currently there are no way of gracefully stopping the AggSender
+			// because the run.go launch the components with a goroutine but doesn't check any return value
+			panic("Stopping AggSender because StopOnFinishedSendingAllCertificates is true")
```
vcastellm, line 170: We don't need the stack trace, I recommed to use a `log.Fatal` here instead of a panic to indicate something controlled.

joanestebanr: It's better  `panic` in golang (even that the name is a bit missleading)  because you can unittest it, also you can recovery with a defer. `log.Fatal` calls to os.Exit and it's not possible to test or capture it

aggsender/flows/feature_maxl2blocknumber.go
joanestebanr: The reason to call this a feature is because all the logic are inside this object. Also you can instantiate or not (it acts as a featureFlag). But, anyway, if the name is confusing, no problem I'll rename it

docs/common_config.md
```diff
@@ -0,0 +1,55 @@
...
+AggsenderPrivateKey = { Method="local", Path="/opt/private_key.kestore", Password="xdP6G8gV9PYs" }
+```
+
+### Google Cloud KMS (GCP)
+Use this method to sign using the Google Cloud KMS infrastructure.
+
+| Name | Type | Example | Description |
+| -----|------|---------|-------------|
+| Method | string | `GCP` | Must be `GCP` |
+| KeyName | string |  | id of the key in Google Cloud |
```
rachit77:
```suggestion
| KeyName | string |  | ID of the key in Google Cloud |
```

docs/aggsender.md
```diff
@@ -149,26 +149,33 @@ The certificate is the data submitted to `Agglayer`. Must be signed to be accept
...
+| GlobalExitRootL2Addr              | Address                   | Address of the GlobalExitRootManager contract on L2 sovereign chain (needed for AggchainProof mode)             |
+| SovereignRollupAddr               | Address                   | Address of the sovereign rollup contract on L1                                                                  |
+| RequireStorageContentCompatibility| bool                      | If true, data stored in the database must be compatible with the running environment                            |
+| RequireNoFEPBlockGap              | bool                      | If true, AggSender should not accept a gap between lastBlock from lastCertificate and first block of FEP        |
+| OptimisticModeConfig              | optimistic.Config         | Configuration for optimistic mode (required by FEP mode)                                                        |
+| MaxL2BlockNumber                  | uint64                    | Set the last block the be included in a certificate (0 disabled)
```
rachit77:
```suggestion
| MaxL2BlockNumber                  | uint64                    | Set the last block to be included in a certificate (0 = disabled)
```

docs/aggsender.md
```diff
@@ -149,26 +149,33 @@ The certificate is the data submitted to `Agglayer`. Must be signed to be accept
...
+|StopOnFinishedSendingAllCertificates| bool                      | When there are no more certificates to be send due MaxL2BlockNumber stop
```
rachit77:
```suggestion
|StopOnFinishedSendingAllCertificates| bool                      | Stop when there are no more certificates to send due to MaxL2BlockNumber
```

aggsender/flows/feature_maxl2blocknumber.go
rachit77: How about  `isNextRangePastLimit` ?

The name `isUpcomingNextRange` isn’t very intuitive.

aggsender/flows/feature_maxl2blocknumber.go
joanestebanr: Renamed: 890fa2b48ed875bd0379674470a53baa17863e9d

aggsender/flows/feature_maxl2blocknumber.go
joanestebanr: I still think doesn't match the name  is not PastLimit is exactly the next range after the limit

.github/workflows/build-push-docker-image.yml
```diff
@@ -0,0 +1,123 @@
+name: Build and Push Docker Image
```
temaniarpit27, line 1: isnt this already present in develop?
@Stefan-Ethernal did this and afaik it is also in use

aggsender/aggsender.go
```diff
@@ -159,6 +160,17 @@ func (a *AggSender) checkDBCompatibility(ctx context.Context) {
 		a.log.Panicf("error checking compatibility data in DB, you can bypass this check using config file. Err: %w", err)
 	}
 }
+func (a *AggSender) checkSendCertificateStopConfidition(err error) {
+	if errors.Is(err, flows.ErrComplete) {
```
temaniarpit27, line 164: maybe combine both ifs?

aggsender/config/config.go
```diff
@@ -73,6 +73,12 @@ type Config struct {
 	RequireNoFEPBlockGap bool `mapstructure:"RequireNoFEPBlockGap"`
 	// OptimisticModeConfig is the configuration for optimistic mode (required by FEP mode)
 	OptimisticModeConfig optimistic.Config `mapstructure:"OptimisticModeConfig"`
+	// MaxL2BlockNumber is the last L2 block number that is going to be included in a certificate
+	// 0 means disabled
+	MaxL2BlockNumber uint64 `mapstructure:"MaxL2BlockNumber"`
+	// StopOnFinishedSendingAllCertificates is a flag to stop the AggSender when it finishes sending all certificates
+	//  up to MaxL2BlockNumber
```
temaniarpit27: extra space here i think

rachit77 (approved):
💯

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -236,9 +256,16 @@ func (a *AggchainProverFlow) GetCertificateBuildParams(ctx context.Context) (*ty
...
 		}
-
 		return nil, err
 	}
+	if a.featureMaxL2Block != nil {
+		// If the feature is enabled, we need to adapt the build params
+		buildParams, err = a.featureMaxL2Block.AdaptCertificate(buildParams)
+		if err != nil {
+			return nil, fmt.Errorf("aggchainProverFlow - error adapting certificate to MaxL2Block. Err: %w", err)
+		}
```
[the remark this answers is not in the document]
joanestebanr: It would break the flow, and basically it's called twice because there are 2 places in the function that create a `CertificateBuildParams`

.github/workflows/build-push-docker-image.yml
joanestebanr: The target of this PR is `release/0.4` and doesn't have the @Stefan-Ethernal action. I required to pre-build a version and have to add

aggsender/aggsender.go
joanestebanr: Between both ifs there are a log that it's important

temaniarpit27 (approved):

PR #660, opened 2025-06-18
fix: migration 0.2 to 0.3 issue

joanestebanr:
## Description

The migration from aggkit `v0.2.0` to `v0.3.0` fails if there are any certificate in database. 

Example of error: 
```
sql: Scan error on column index 4, name "previous_local_exit_root": converting NULL to string is unsupported
```

###  Meddler don't support `*common.Hash`
This require fix local meddler function but also the same functions in `zkevm-ethtx-manager` library (check this [CDK PR](https://github.com/0xPolygon/cdk/pull/285))

###  New db field  `l1_info_tree_leaf_count` is a nullable `INTERGER`  but the reception type in code `uint32` (notice that have no pointer). 
So when execute migration the old data is set to `NULL` but it's not possible to read it because the code field is `uint32` instead of `*uint32`. 
Checking the code this value is only used from DB for retry a FEP certificate, so, as a trade off, we delete the non settle certificates from DB and set the default value 0.

### DIFF
--- a/aggsender/db/aggsender_db_storage_test.go
+++ b/aggsender/db/aggsender_db_storage_test.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"math/big"
+	"os"
 	"path"
 	"testing"
 	"time"
@@ -21,6 +23,22 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func Test_StorageExploratory(t *testing.T) {
+	t.Skip()
+	path := os.Getenv("DB_AGGSENDER_0_2")
+	if path == "" {
+		t.Fatalf("environment variable DB_AGGSENDER_0_2 is not set")
+	}
+	cfg := AggSenderSQLStorageConfig{
+		DBPath:                  path,
+		KeepCertificatesHistory: true,
+	}
+	storage, err := NewAggSenderSQLStorage(log.WithFields("aggsender-db"), cfg)
+	require.NoError(t, err)
+	cert, err := storage.GetLastSentCertificate()
+	require.NoError(t, err)
+	fmt.Print(cert)
+}
 func Test_Storage(t *testing.T) {
 	ctx := context.Background()
 
--- a/aggsender/db/aggsender_db_storage_test.go
+++ b/aggsender/db/aggsender_db_storage_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"math/big"
 	"os"
 	"path"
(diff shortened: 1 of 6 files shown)


aggsender/db/aggsender_db_storage_test.go
```diff
@@ -21,6 +23,22 @@ import (
...
+	}
+	cfg := AggSenderSQLStorageConfig{
+		DBPath:                  path,
+		KeepCertificatesHistory: true,
+	}
+	storage, err := NewAggSenderSQLStorage(log.WithFields("aggsender-db"), cfg)
+	require.NoError(t, err)
+	cert, err := storage.GetLastSentCertificate()
+	require.NoError(t, err)
+	fmt.Print(cert)
```
temaniarpit27: do we need to print cert here?

db/meddler.go
```diff
@@ -152,41 +152,58 @@ func (b MerkleProofMeddler) PreWrite(fieldPtr interface{}) (saveValue interface{
 type HashMeddler struct{}
 
 // PreRead is called before a Scan operation for fields that have the HashMeddler
-func (b HashMeddler) PreRead(fieldAddr interface{}) (scanTarget interface{}, err error) {
+func (m HashMeddler) PreRead(fieldAddr interface{}) (scanTarget interface{}, err error) {
```
temaniarpit27, line 155: `m` or `b`?
other places have `b`

db/meddler.go
```diff
@@ -152,41 +152,58 @@ func (b MerkleProofMeddler) PreWrite(fieldPtr interface{}) (saveValue interface{
...
 	// give a pointer to a byte buffer to grab the raw data
+	_, ok := fieldAddr.(**common.Hash)
+	if ok {
+		// This is becase if not the rows.Scan fails 'converting NULL to string is unsupported'
```
temaniarpit27, line 159: because

db/meddler_test.go
```diff
@@ -62,3 +66,72 @@ func TestHashMeddler_PreWrite(t *testing.T) {
...
+	require.ErrorContains(t, err, "converting NULL to string is unsupported")
+}
+
+func TestMeddlerHashPointerIsNotNull(t *testing.T) {
+	db := createExampleDB(t)
+	var certificateInfo certificateInfo
+	err := meddler.QueryRow(db, &certificateInfo, "SELECT * FROM certificate_info where height=1;")
+	require.NoError(t, err, "data case")
+	require.NotNil(t, certificateInfo.FinalizedL1InfoTreeRoot, "FinalizedL1InfoTreeRoot should not be nil for height 1")
+	fmt.Print(certificateInfo)
```
temaniarpit27: print here

temaniarpit27 (approved):
few comments
otherwise LGTM

vcastellm:
Late LGTM

joanestebanr:
This PR have been reverted from branch `release/0.4`

PR #661, opened 2025-06-19
fix: migration 0.2 to 0.3 issue V2

joanestebanr:
## Description

The migration from aggkit `v0.2.0` to `v0.3.0` fails if there are any certificate in database. 

Example of error: 
```
sql: Scan error on column index 4, name "previous_local_exit_root": converting NULL to string is unsupported
```

###  Meddler don't support `*common.Hash`
This require fix local meddler function but also the same functions in `zkevm-ethtx-manager` library (check this [CDK PR](https://github.com/0xPolygon/cdk/pull/285))

###  New db field  `l1_info_tree_leaf_count` is a nullable `INTERGER`  but the reception type in code `uint32` (notice that have no pointer). 
So when execute migration the old data is set to `NULL` but it's not possible to read it because the code field is `uint32` instead of `*uint32`. 
Checking the code this value is only used from DB for retry a FEP certificate, so, as a trade off, we delete the non settle certificates from DB and set the rest to 0 (that is safe because this value is not used) 

**NOTE**: the difference with PR #660 is that instead of modifying migration `0002.sql` it add a new migration  `0004.sql` and doesn't change the type of the column `certificate_info.l1_info_tree_leaf_count` neither `certificate_info_history.l1_info_tree_leaf_count`. This means that the column is still nullable, but sqlite doesn't allow to modify that after the creation of the table (it require a copy to another table)

### DIFF
--- a/aggsender/db/migrations/0004.sql
+++ b/aggsender/db/migrations/0004.sql
@@ -1,0 +1,14 @@
+--- Fix migration from 0001 to 0002 that field l1_inffo_tree_leaf_count can't be read
+-- +migrate Down
+
+
+-- +migrate Up
+
+-- The field l1_info_tree_leaf_count is unknown for previous data stored in DB
+-- the only place where is used is to retry a FEP. So we delete all non-finalized certificates
+-- so it can't produce the error keeping as much data as possible.
+DELETE FROM certificate_info WHERE  status != 4 AND l1_info_tree_leaf_count IS NULL; -- Remove non-finalized certificates with NULL l1_info_tree_leaf_count
+DELETE FROM certificate_info_history WHERE status != 4 AND l1_info_tree_leaf_count IS NULL; -- Remove non-finalized certificates with NULL l1_info_tree_leaf_count
+--- The rest of l1info_tree_leaf_count is not used in the code, so we can safely set to 0
+UPDATE certificate_info SET l1_info_tree_leaf_count = 0 WHERE l1_info_tree_leaf_count IS NULL;
+UPDATE certificate_info_history SET l1_info_tree_leaf_count = 0 WHERE l1_info_tree_leaf_count IS NULL;
--- a/aggsender/db/migrations/0004.sql
+++ b/aggsender/db/migrations/0004.sql
@@ -1,4 +1,4 @@
---- Fix migration from 0001 to 0002 that field l1_inffo_tree_leaf_count can't be read
+--- Fix migration from 0001 to 0002 that field l1_info_tree_leaf_count can't be read
 -- +migrate Down
 
 
--- a/db/meddler.go
+++ b/db/meddler.go
@@ -156,7 +156,7 @@ func (m HashMeddler) PreRead(fieldAddr interface{}) (scanTarget interface{}, err
 	// give a pointer to a byte buffer to grab the raw data
 	_, ok := fieldAddr.(**common.Hash)
 	if ok {
-		// This is becase if not the rows.Scan fails 'converting NULL to string is unsupported'
+		// This is because if not the rows.Scan fails 'converting NULL to string is unsupported'
 		return new(*string), nil
 	}
 	return new(string), nil
--- a/db/meddler.go
+++ b/db/meddler.go
@@ -152,13 +152,45 @@ func (b MerkleProofMeddler) PreWrite(fieldPtr interface{}) (saveValue interface{
 type HashMeddler struct{}
 
 // PreRead is called before a Scan operation for fields that have the HashMeddler
-func (b HashMeddler) PreRead(fieldAddr interface{}) (scanTarget interface{}, err error) {
+func (m HashMeddler) PreRead(fieldAddr interface{}) (scanTarget interface{}, err error) {
 	// give a pointer to a byte buffer to grab the raw data
+	_, ok := fieldAddr.(**common.Hash)
+	if ok {
+		// This is becase if not the rows.Scan fails 'converting NULL to string is unsupported'
+		return new(*string), nil
+	}
 	return new(string), nil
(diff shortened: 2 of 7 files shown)


temaniarpit27 (approved):

db/meddler.go
```diff
@@ -152,45 +152,62 @@ func (b MerkleProofMeddler) PreWrite(fieldPtr interface{}) (saveValue interface{
...
 		return nil
 	}
+	return errors.New("fieldPtr is not **common.Hash")
+}
 
-	// If fieldPtr is neither a *common.Hash nor a **common.Hash, return an error
-	return errors.New("fieldPtr is not *common.Hash or **common.Hash")
+// PostRead is called after a Scan operation for fields that have the HashMeddler
+func (m HashMeddler) PostRead(fieldPtr, scanTarget interface{}) error {
+	_, ok := scanTarget.(**string)
```
goran-ethernal, line 190: The double pointer is getting confusing 😵‍💫. The logic is wuite complex. I am thinking,would it be cleaner to we have a nullable and non nullable meddler 🤔.

aggsender/db/migrations/0004.sql
```diff
@@ -0,0 +1,14 @@
+--- Fix migration from 0001 to 0002 that field l1_info_tree_leaf_count can't be read
+-- +migrate Down
+
+
+-- +migrate Up
+
+-- The field l1_info_tree_leaf_count is unknown for previous data stored in DB
+-- the only place where is used is to retry a FEP. So we delete all non-finalized certificates
+-- so it can't produce the error keeping as much data as possible.
+DELETE FROM certificate_info WHERE  status != 4 AND l1_info_tree_leaf_count IS NULL; -- Remove non-finalized certificates with NULL l1_info_tree_leaf_count
```
goran-ethernal, line 10: Could this be solved with a nullable meddler for this field that returns 0 if it is null? And not have a migration file that removes records?

joanestebanr: The problem is when this value is used by aggsender:  if it retry a certificate using DB data you are going to use `l1_info_tree_leaf_count` as zero when it's not a valid value. And there are no problem to remove this rows because is going to get it from `agglayer`

db/meddler.go
joanestebanr: It's confusing, but required, it's the diference beetween a nullable or not field. I think that creating another type is  going to be as confuse as this one. Because, if you see the code at the begining it decide if it's a double pointer or not.
Also I think that having another meddler is going to be error prone

aggsender/db/migrations/0004.sql
goran-ethernal: But that is something to be solved in the code and not in migration in my opinion.

If someone was using aggsender on 0.2.0 which does not have this field, it is expected that this field is 0 (I agree, the migration file should have put a DEFAULT 0).

Since it is only needed for the FEP certificates, so why not check its validity in code. We could even use the CertType field, so if it is CertType = Unknown (which in this case is true because its a  certificate from an old version) and if its a retry, and leaf count is zero, get it again. This way, the migration does not remove any records. WDYT?

joanestebanr: To allow to `flow_aggchain_prover.go` manage this situation we need to change: 
-  modify struct `db.certificateInfo` changing ` L1InfoTreeLeafCount` as   `* uint32`. 
- modify struct `types.CertificateHeader` changing  ` L1InfoTreeLeafCount` as   `* uint32`. 

This produce a explosion of changes along all the code, that only have sense to mitigate the migration problem in `0002.sql` and we have to maintain all this code. 
Doing it as a migration  is so much clean, and we don't have to maintain a strange code in the future that it's difficult to remember why we have added.

Maybe we can do a huddle to discuss it. WDYT?

vcastellm:
It's normal that you cannot change the nullable property of an existing column, but it can be done with a new column (not table), WDYT if the migration:

1. Creates a new column that is not nullable with 0 as default value
2. copy the values from the old column to the new one (in case there are)
3. Drop the old column

By doing this we won't need to delete records.

joanestebanr:
> It's normal that you cannot change the nullable property of an existing column, but it can be done with a new column (not table), WDYT if the migration:
> 
> 1. Creates a new column that is not nullable with 0 as default value
> 2. copy the values from the old column to the new one (in case there are)
> 3. Drop the old column
> 
> By doing this we won't need to delete records.

Good point!, we can apply that in another PR to fix the nullable issue. But then we can rename the new column with the old name or we have to keep the new field name? 
 
But it still require to delete rows, because 0 it's not a valid value, and in some cases the code use this value so the best approach is delete the row or, as @goran-ethernal suggest have code that deal with the null value and reject to use

vcastellm (approved):
> But then we can rename the new column with the old name or we have to keep the new field name?

true, I missed one step:
4. Rename new column to old name

This is usually how I did this in the past.

> because 0 it's not a valid value

Bc 0 is not valid from the logic perspective I guess, but then we need to deal with 0 in the code but we won't have issue with the uint32.

If you want to to this in an upcoming PR it's ok, then this PR LGTM but consider fixing the debt introduced here as commented

joanestebanr:
> Bc 0 is not valid from the logic perspective I guess, but then we need to deal with 0 in the code but we won't have issue with the uint32.

I mean that we don't known if 0 is the valid value or just the default value due the migration. I guess that 0 from a point of view of logic it's a valid value (I'm not sure about this point, if it's a forbidden value then everything will be more easy, for sure). So the only way to known that is convert the `uint32` field to `* uint32` to be sure that the value is not informed in DB, in that case  it make sense to keep the field as nullable, WDTY?

PR #662, opened 2025-06-19
feat: handle gi

temaniarpit27:


agglayer/types/types.go
```diff
@@ -440,20 +440,26 @@ func (t *TokenInfo) String() string {
 
 // GlobalIndex represents the global index of an imported bridge exit
 type GlobalIndex struct {
-	MainnetFlag bool   `json:"mainnet_flag"`
-	RollupIndex uint32 `json:"rollup_index"`
-	LeafIndex   uint32 `json:"leaf_index"`
+	FirstUnusedBits *big.Int `json:"first_unused_bits,omitempty"`
```
Stefan-Ethernal, line 443: Aren't the most significant 191 bits padded with 0s always? If that is the case, we probably do not need to capture it in the field and with that assumption logic could be probably simplified. WDYT?

bridgesync/processor.go
```diff
@@ -864,56 +870,81 @@ func buildNetworkIDsFilter(networkIDs []uint32, networkIDColumn string) string {
...
+	rollupIndexBig := new(big.Int).And(rollupIndexShifted, rollupIndexMask)
+	rollupIndex = uint32(rollupIndexBig.Uint64())
+
+	// Extract local exit root index from bits 0-31 (least significant bits)
+	localExitRootIndexMask := new(big.Int).Sub(
+		new(big.Int).Lsh(big.NewInt(1), globalIndexLocalExitRootBits),
+		big.NewInt(1),
+	)
+	localExitRootIndexBig := new(big.Int).And(globalIndex, localExitRootIndexMask)
+	localExitRootIndex = uint32(localExitRootIndexBig.Uint64())
```
Stefan-Ethernal, line 945: Couldn't we stick with the previous logic (which extracts different portions of global index)? IMO it was cleaner and more verbose. WDYT?

agglayer/types/types.go
Stefan-Ethernal: Should we add some validation which enforces the first 191 bits to be 0s always? AFAIK we don't care about them on the SC?

Stefan-Ethernal (commented):
As a side note, it could be much better if we have added validation to the smart contract, although I do understand that they are already audited and changes are not welcome.

Issue #640, opened 2025-06-13
option to run only tests in run-local-e2e.sh


temaniarpit27:
currently we dont have a way to run tests with existing setup
add option to run either setup, tests or both

PR #671, opened 2025-06-23
feat: provide option to run kurtosis or tests or both

temaniarpit27:
#640

### DIFF
--- a/test/run-local-e2e.sh
+++ b/test/run-local-e2e.sh
@@ -10,82 +10,115 @@ log_error() { echo -e "${RED}[ERROR]${NC} $*"; }
 
 trap 'log_error "Script failed at line $LINENO"' ERR
 
-if [ "$#" -lt 2 ]; then
-    echo "Usage: $0 <test_type: single-l2-network-fork12-op-succinct | single-l2-network-fork12-pessimistic | multi-l2-networks-2-chains | multi-l2-networks-3-chains> <kurtosis_repo> [e2e_repo]"
+if [ "$#" -lt 3 ]; then
+    echo "Usage: $0 <test_type: single-l2-network-fork12-op-succinct | single-l2-network-fork12-pessimistic | multi-l2-networks-2-chains | multi-l2-networks-3-chains> <kurtosis_repo_path> <e2e_repo_path>"
+    echo ""
+    echo "Arguments:"
+    echo "  test_type           Type of test to run"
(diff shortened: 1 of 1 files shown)


Stefan-Ethernal (approved):

test/run-local-e2e.sh
```diff
@@ -10,82 +10,115 @@ log_error() { echo -e "${RED}[ERROR]${NC} $*"; }
...
     ;;
 esac
-log_info "$ENCLAVE_NAME enclave started successfully."
-popd >/dev/null
 
-if [ -n "$E2E_FOLDER" ]; then
-    if [ ! -d "$E2E_FOLDER" ]; then
-        log_error "The provided E2E folder does not exist: $E2E_FOLDER"
+# Run Kurtosis setup if path is provided and not '-'
+if [ "$KURTOSIS_REPO_PATH" != "-" ]; then
```
goran-ethernal, line 56: Should we check if the `KURTOSIS_REPO_PATH` is empty as well?

temaniarpit27: https://github.com/agglayer/aggkit/pull/671/files#diff-eae4732425264cca57337c15a9473714d14b4015ba5a896c01125683a78df6a4R57
doing it here

goran-ethernal: Hmm, doesn't that mean if I do not provide a flag, like `kurtosis_repo_path` (meaning its empty ""), it will enter here and log an error?

temaniarpit27: yes it will enter here and log and error

goran-ethernal: But aren't we making them optional? To me it makes sense, if I do not define the `kurtosis_repo_path` flag, it will not run kurtosis. Its kind of easier than to have to put a `-` for that flag. Just thinking from a user perspective.

Same for `e2e_repo_path`.

goran-ethernal: But, I see that you always expect 3 parameters. Its easier for the code probably. Which makes sense, since only we are going to use this. So, I am ok than to leave it as it is.

temaniarpit27: it is not optional now.
earlier only e2e_path was optional
but now we need a way to determine what we want to do, either run only setup, only test or both
if we leave lets say kurtosis_path empty, our script wont be able to determine whether it is kurtosis path or e2e path

goran-ethernal (approved):

Issue #627, opened 2025-06-11
[FEATURE] add counter on get logs to avoid endless recursion


arnaubennassar:
# Feature Request

`func (d *EVMDownloaderImplementation) GetEventsByBlockRange(ctx context.Context, fromBlock, toBlock uint64) EVMBlocks` is a recursive function in the case of a block hash missmatch. In the case of a permanent missmatch (very unlikely to happen) this function will call itself until out of memory.

Add a counter on the recursion and limit how many times it can call itself

PR #672, opened 2025-06-23
fix: GetEventsByBlockRange retry

temaniarpit27:
#627


sync/evmdownloader.go
```diff
@@ -337,10 +342,14 @@ func (d *EVMDownloaderImplementation) GetEventsByBlockRange(ctx context.Context,
 				if b.Hash != l.BlockHash {
 					d.log.Infof(
 						"there has been a block hash change between the event query and the block query "+
-							"for block %d: %s vs %s. Retrying.",
-						l.BlockNumber, b.Hash, l.BlockHash,
+							"for block %d: %s vs %s. Retrying attempt %d.",
+						l.BlockNumber, b.Hash, l.BlockHash, retryCount,
```
Stefan-Ethernal: WDYT?
```suggestion
							"for block %d: %s vs %s. Retrying attempt %d/%d.",
						l.BlockNumber, b.Hash, l.BlockHash, retryCount, MaxRetryCountBlockHashMismatch
```

sync/evmdownloader.go
```diff
@@ -17,6 +17,7 @@ import (
 
 const (
 	DefaultWaitPeriodBlockNotFound = time.Millisecond * 100
+	MaxRetryCountBlockHashMismatch = 3
```
Stefan-Ethernal: Probably makes sense to introduce this as a configuration param. WDYT?

Stefan-Ethernal (approved):
:100:

goran-ethernal (approved):
💯

sync/evmdownloader.go
temaniarpit27: i feel this is more of a internal logic and not to be configured by client

temaniarpit27: i increased it to 5 for more chances

PR #675, opened 2025-06-25
fix: remove root access when building docker image

### LINKED ISSUE #674: Remove root access requirements when building Docker image
Remove root access requirements inside the Docker container (aggsender/aggkit) and remove shell access inside the Docker container for `aggsender`.

Consider making 2 `Dockerfile(s)`:
- `Dockerfile.release` - that removes the shell
- `Dockerfile.development` - that has the shell exposed

Note: figure out if there is necessity to generate 2 Docker images for each release: with and without shell exposed.

Stefan-Ethernal:
This pull request introduces significant enhancements to Docker workflows, refines the Dockerfile for production and development environments, and removes unused dependencies from the `go.mod` file. The most important changes include adding support for production and development variants in Docker workflows, improving the Dockerfile with conditional shell inclusion, and cleaning up unused Go dependencies.

### Docker Workflow Enhancements:
* [`.github/workflows/build-push-docker-image.yml`](diffhunk://#diff-16a6973e9a12990dc9265a3cf2ce7a82225ed139a55e85fb6e86e6d131afe127R27-R36): Added support for production and development variants in the build matrix, with distinct image tags and caching scopes. Outputs now include separate image names and tags for production and development builds. [[1]](diffhunk://#diff-16a6973e9a12990dc9265a3cf2ce7a82225ed139a55e85fb6e86e6d131afe127R27-R36) [[2]](diffhunk://#diff-16a6973e9a12990dc9265a3cf2ce7a82225ed139a55e85fb6e86e6d131afe127L48-R71) [[3]](diffhunk://#diff-16a6973e9a12990dc9265a3cf2ce7a82225ed139a55e85fb6e86e6d131afe127L65-R144) [[4]](diffhunk://#diff-16a6973e9a12990dc9265a3cf2ce7a82225ed139a55e85fb6e86e6d131afe127L110-R183)
* [`.github/workflows/release.yml`](diffhunk://#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R33-R61): Updated release workflow to handle production and development variants, including separate metadata, image builders, and digest handling. [[1]](diffhunk://#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R33-R61) [[2]](diffhunk://#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34L55-R96) [[3]](diffhunk://#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34L81-R133) [[4]](diffhunk://#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R142-R152) [[5]](diffhunk://#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34L125-R179)

### Dockerfile Improvements:
* [`Dockerfile`](diffhunk://#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L1-R58): Refactored to use Alpine as the base image for both build and runtime stages. Introduced conditional shell inclusion for development environments, improved dependency management, and enhanced security by removing shells in production builds.

### Dependency Cleanup:
* [`go.mod`](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L48): Removed unused dependencies, including `modernc.org/sqlite` and related libraries, as well as other indirect dependencies that are no longer required. [[1]](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L48) [[2]](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L102) [[3]](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L165) [[4]](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L178) [[5]](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L222-L224)

### Makefile Update:
* [`Makefile`](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R78-R81): Added a new `build-docker-ci` target to build Docker images specifically for CI environments with shell inclusion.

### Minor Code Adjustments:
* [`bridgesync/processor.go`](diffhunk://#diff-18eb3102bf480ea624d689abce486ee67aa3ec80ca545c2d7952ff12dfffeea6L28): Removed the unused import of `modernc.org/sqlite`.

Fixes #674


Dockerfile
```diff
@@ -1,31 +1,39 @@
...
-RUN addgroup --system appgroup && adduser --system --ingroup appgroup appuser
-USER appuser
+# Copy source and build
+COPY --chown=appuser:appgroup . .
+RUN make build-aggkit
 
+# ================================
+# STAGE 2: Final runtime image (distroless)
+# ================================
+FROM gcr.io/distroless/static-debian12:nonroot
```
joanestebanr: This distro doesn't have any shell, package manager,  .... this is going to prevent from inspect  inspect the database (also the sqlite client have been removed) ... maybe make sense to have some tools inside the docker. 
I see that is a marked a best practice employed by Google, but I don't known if fit or needs. It's very common to get into the docker, modify files, or just check the running process inside.. 

 WDYT?

joanestebanr (commented):
I see you create a docker with not possible interaction, no tools, no shell... I'm wondering if we are ready for that

Dockerfile
Stefan-Ethernal: This is an issue that was reported by the Gateway folks:

### 1. root access requirement
> the issue was that we were not able to sign transactions with AWS KMS using the appuser in the container and had to run the container with root user to be able to sign. We suspect that this happens because the appuser is created as system user and does not have a home directory and aws signing library needs home. Any reason this user is created as system (-S option)?. I would suggest adding -m option (or maybe -h for Alpine, not exactly sure) to create a home directory for app user, we can test it when the build is available.

Comment from my end: it seems like we can easily address this, by creating non system user and exposing the home dir for it.

### 2. disable shell
> Regarding the shell access - we need to run aggsender fully locked out, so no on can shell into the container. We have no issues accessing sqlite files, since they are mounted to the container in a vlume and can be accesses on the host system

Comment from my end: when I have [switched](https://github.com/agglayer/aggkit/pull/675/commits/12be61a60b500d299a40a0ea8931a9e6a511ab8f) to `distroless` as a base image, the e2e tests have failed with the following error message
```bash
[FATAL tini (7)] exec sh failed: No such file or directory
```
which indicates that we are probably not ready to disable the shell (I'm unaware which component is using the shell, but AFAIK there is no the `aggkit` component that is explicitly using it).

revitteth:
@Stefan-Ethernal I will push on this one - I have a potential solution!

Stefan-Ethernal:
> @Stefan-Ethernal I will push on this one - I have a potential solution!

Awesome stuff @revitteth, it seems as an elegant approach. Thanks! 

I was experimenting with removing the shell even from the ci (but leaving it for development purposes, because of debugging and troubleshooting). However, it requires some tweaks on the Kurtosis CDK front as well and I didn't quite make it to work yet. I will try some more next week.

revitteth:
@Stefan-Ethernal glad it's good, but yeah I'm unaware of the 'contexts beyond CI' in which the container might be shelled - I guess whatever is doing that (i.e. Kurtosis) should do it's checks some other way (RPC, etc.)?

vcastellm:
@revitteth what work do you need to do forward? I've been assigned to the task

vcastellm:
@revitteth I see that you've set it up to conditionally setup the shell and with a non-login non-shell user, e2e tests are now passing ok, that's great but we'll need a full fledged image for development purposes, we can conditionally setup:

1. the user with shell too
2. Install sqlite

I think that should cover our needs, but we'll need two different images to be released.

Dockerfile
```diff
@@ -1,31 +1,61 @@
...
+FROM alpine:3.22
+
+# Build argument to control shell installation
+ARG INCLUDE_SHELL=false
+
+# Install runtime dependencies
+RUN apk add --no-cache sqlite-libs ca-certificates && \
+    if [ "$INCLUDE_SHELL" = "true" ]; then \
+        echo "Including shell and sqlite CLI for CI/dev environment" && \
+        apk add --no-cache sqlite; \
```
joanestebanr: Missing some tools that had previous image in -dev docker, like procps? or it's installed in another way?

Dockerfile
```diff
@@ -1,31 +1,61 @@
...
-    procps \
-    libssl-dev && \
-    rm -rf /var/lib/apt/lists/*
-COPY --from=build /app/target/aggkit /usr/local/bin/
-
-# ADD NON-ROOT USER
-RUN addgroup --system appgroup && adduser --system --ingroup appgroup appuser
+
+# Compile binary
+RUN make build-aggkit
```
joanestebanr: In dev imange missing  aggkit tools installed

bridgesync/processor.go
```diff
@@ -25,7 +25,6 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/iden3/go-iden3-crypto/keccak256"
 	"github.com/russross/meddler"
-	_ "modernc.org/sqlite"
```
joanestebanr, line 28: maybe, as you point to do this version without changing code, unless this change is mandatory make sense to revert to avoid any code change. WDYT?

Stefan-Ethernal: My two cents, because I've made this change, is that it does not harm removing this dependency, as it is redundant with the `github.com/mattn/go-sqlite3`. At first, I thought removing the `github.com/mattn/go-sqlite3` (which would probably eliminate the need for gcc and CGO cross compile), but it turned out that it is somewhat slower to the `modernc.org/sqlite`.

As a bottom line, in case the CI doesn't prove different, we should merge it now, as it is pretty isolated change (and we have plenty of bridge service e2e tests). WDYT?

Dockerfile
vcastellm: Let's add them

Dockerfile
vcastellm: Which tools?

joanestebanr: `make build-tools` generate the binary `aggsender_find_imported_bridge`

Stefan-Ethernal:
:heavy_check_mark: LGTM I guess (I cannot approve it because I've created the PR) :slightly_smiling_face:

.github/workflows/build-push-docker-image.yml
```diff
@@ -62,39 +82,66 @@ jobs:
...
+
+      - name: Download digests for production
+        if: matrix.variant.suffix == ''
+        uses: actions/download-artifact@v4
+        with:
+          path: /tmp/digests
+          pattern: digests-linux-amd64
+          merge-multiple: true
+
+      - name: Download digests for production
```
goran-ethernal, line 131: This looks duplicated.

Dockerfile
```diff
@@ -1,31 +1,61 @@
...
+      echo "Removing shell for production security" && \
+      rm -f /bin/sh /bin/bash /bin/ash; \
+    fi
+
+# Set the working directory and user
+WORKDIR /home/appuser
 USER appuser
 
+# Copy the built binary from the builder stage
+COPY --from=builder /app/target/aggkit /usr/local/bin/aggkit
```
joanestebanr, line 57: Could you copy also the tool `aggsender_find_imported_bridge` into the docker if include shell? I see that this is also a issue in previous DockerFile

Dockerfile
```diff
@@ -1,31 +1,61 @@
...
+    else \
+        adduser -D -G appgroup -h /home/appuser -s /bin/false appuser; \
+    fi && \
+    mkdir -p /home/appuser && \
+    chown -R appuser:appgroup /home/appuser
+
+# Remove shell for production security (only if not INCLUDE_SHELL)
+RUN if [ "$INCLUDE_SHELL" != "true" ]; then \
+      echo "Removing shell for production security" && \
+      rm -f /bin/sh /bin/bash /bin/ash; \
```
joanestebanr: Even removing this link you can still open a shell: 
```
docker run -it --entrypoint /bin/busybox  aggkit:local sh
```

.github/workflows/build-push-docker-image.yml
vcastellm: Only one pattern can be used at a time, one is for `amd64` and other for `arm64`

vcastellm (approved):

joanestebanr (approved):

goran-ethernal (approved):

PR #677, opened 2025-06-25
feat: use minimal args

temaniarpit27:
#616


.github/test_e2e_multi_chains_args_1_base.json
Stefan-Ethernal, line 1: Can we rename this file to e.g. `test_e2e_args_base.json`, since this one is used for all the other "full/merged" args files. WDYT?

.github/test_e2e_multi_chains_args_1_base.json
```diff
@@ -0,0 +1,15 @@
+{
+  "deployment_stages": {
+    "deploy_l2_contracts": true,
+    "deploy_cdk_bridge_infra": false,
+    "deploy_optimism_rollup": false
+  },
+  "args": {
+    "verbosity": "debug",
+    "cdk_node_image": "aggkit:local",
```
Stefan-Ethernal: It might be a good idea to include image versions for the other docker images as well (at least agglayer, whereas the aggkit-prover and cdk-erigon can be enumerated in specific args files)?

I know that you probably wanted to rely on `input-parser.star` in kurtosis cdk, but maybe it is better to pin the versions here. WDYT (as I don't have a strong opinion)?

.github/test_e2e_single_chain_fork12_op_succinct_args.json
```diff
@@ -18,15 +14,12 @@
     "op_succinct_max_concurrent_proof_requests": "1",
     "op_succinct_max_concurrent_witness_gen": "1",
     "op_succinct_range_proof_interval": "60",
-    "l1_seconds_per_slot": 2,
     "additional_services": [
-      "bridge_spammer"
```
Stefan-Ethernal, line 23: We would need the bridge_spammer service, because it is dependant upon it in the optimistic mode tests (https://github.com/agglayer/e2e/blob/dcdf2b4481ef962374c827d22b64337261104621/tests/op/optimistic-mode.bats#L110)

.github/test_e2e_single_chain_fork12_op_succinct_args.json
```diff
@@ -6,10 +6,6 @@
   "args": {
```
Stefan-Ethernal, line 6: Don't you need to include as well to `deployment_stages`?
```json
"deploy_optimism_rollup": "true",
```

Stefan-Ethernal (approved):
Generally LGTM 💯

.github/test_e2e_single_chain_fork12_op_succinct_args.json
temaniarpit27: this is by default true

.github/test_e2e_multi_chains_args_1_base.json
temaniarpit27: in my opinion we should use whatever images are present in kurtosis-cdk. anyway we are pinning a kurtosis-cdk version in our CI. so that should take care of image versions automatically.
i am in favour of keeping files minimal. also, the new image versions are tested in kurtosis-cdk CI. where they dont fix an image version. so we are somewhat safe with new image versions.

.github/test_e2e_multi_chains_args_2.json
```diff
@@ -12,16 +11,9 @@
...
     "cdk_node_image": "aggkit:local",
-    "agglayer_image": "ghcr.io/agglayer/agglayer:0.3.0-rc.21",
-    "cdk_erigon_node_image": "hermeznetwork/cdk-erigon:v2.61.19",
-    "zkevm_contracts_image": "jhkimqd/zkevm-contracts:v10.1.0-rc.5-fork.12",
     "additional_services": [],
-    "consensus_contract_type": "pessimistic",
-    "sequencer_type": "erigon",
     "erigon_strict_mode": false,
-    "gas_token_enabled": false,
     "enable_normalcy": true,
```
temaniarpit27, line 16: @Stefan-Ethernal any idea what this flag is?

.github/test_e2e_single_chain_fork12_op_succinct_args.json
rachit77: i would suggest to add this explicitly because in the past default behaviour was changed

temaniarpit27: my opinion would be use default values as much as possible as kurtosis and minimal here.

rachit77 (approved):
💯

PR #682, opened 2025-06-27
fix: add lock to loadTrackedHeaders in reorg detector

### LINKED ISSUE #680: [BUG] subscriptions race condition in ReorgDetector at startup
add `rd.subscriptionsLock` [here](https://github.com/agglayer/aggkit/blob/develop/reorgdetector/reorgdetector.go#L263)

Stefan-Ethernal:
## Description

Fix race condition in loadTrackedHeaders in reorg detector on startup.
Also fix typo in function name.

Fixes #680 , #683

### DIFF
--- a/lastgersync/processor.go
+++ b/lastgersync/processor.go
@@ -74,8 +74,10 @@ func (p *processor) ProcessBlock(ctx context.Context, block sync.Block) error {
 	if err != nil {
 		return fmt.Errorf("failed to start transaction: %w", err)
 	}
+
+	shouldRollback := true
 	defer func() {
-		if err != nil {
+		if shouldRollback {
 			p.log.Errorf("transaction rollback due to error: %v", err)
 			if errRollback := tx.Rollback(); errRollback != nil {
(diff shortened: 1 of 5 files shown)


lastgersync/processor.go
```diff
@@ -74,8 +74,10 @@ func (p *processor) ProcessBlock(ctx context.Context, block sync.Block) error {
 	if err != nil {
 		return fmt.Errorf("failed to start transaction: %w", err)
 	}
+
+	shouldRollback := true
 	defer func() {
-		if err != nil {
+		if shouldRollback {
```
temaniarpit27, line 80: p.rollbackTransaction(tx)
similar func here?

temaniarpit27 (approved):
💯

lastgersync/processor.go
Stefan-Ethernal: Probably doesn't make sense at this point, because it is invoked only from a single place.

goran-ethernal (approved):

joanestebanr (approved):

PR #685, opened 2025-06-27
test: handle sub calls

rachit77:
## Description

Check CI for e2e internal claims test: https://github.com/agglayer/e2e/pull/86


bridgeservice/bridge.go
```diff
@@ -41,14 +42,15 @@ const (
...
-	globalIndexParam  = "global_index"
+	networkIDParam      = "network_id"
+	networkIDsParam     = "network_ids"
+	pageNumberParam     = "page_number"
+	pageSizeParam       = "page_size"
+	depositCountParam   = "deposit_count"
+	fromAddressParam    = "from_address"
+	leafIndexParam      = "leaf_index"
+	globalIndexParam    = "global_index"
+	populateProofsParam = "populate_proofs"
```
temaniarpit27: we can also convert this to `fullResponse` in which we will get all claims field and not just this data.

temaniarpit27 (approved):
💯

bridgeservice/bridge.go
rachit77: if you set `populate_proofs=true` in the request than you get all the claims field in the response

temaniarpit27: then rename it to full_response/include_all_fields, something like that. not restrict to specific fields

joanestebanr (approved):

bridgeservice/bridge.go
rachit77: done: 0f323c93d0ea42ac5a73b03c2c5310bf2f82d7fc

PR #691, opened 2025-07-01
feat: panic bridge service db when data needs to resynced (2)

### LINKED ISSUE #657: panic if data needs to be resynched
If a version update (such as `v0.5`) the node should panic if there is existing data on the relevant DBs

rachit77:
## Description

**Approach: 2**

This PR implements a database compatibility checking mechanism for the bridge service that prevents data corruption during version upgrades.

Fixes https://github.com/agglayer/aggkit/issues/657


bridgesync/migrations/bridgesync0004.sql
```diff
@@ -0,0 +1,11 @@
...
+DROP TABLE IF EXISTS resync_counter;
+
+-- +migrate Up
+CREATE TABLE resync_counter (
+    key VARCHAR PRIMARY KEY,
+    value INTEGER NOT NULL
+);
+
+-- Insert initial resync counter value
+INSERT INTO resync_counter (key, value) VALUES ('resync_version', 1);
```
temaniarpit27: always append newline at the end of file. there is a setting in vscode/cursor which automatically does that

bridgesync/processor.go
```diff
@@ -945,3 +958,59 @@ func (p *processor) isHalted() bool {
 	defer p.mu.RUnlock()
 	return p.halted
 }
+
+func checkAndHandleResync(db *sql.DB, logger *log.Logger, name string) error {
+	ctx := context.Background()
```
temaniarpit27: not sure if we have usecase for this ctx

bridgesync/processor.go
```diff
@@ -945,3 +958,59 @@ func (p *processor) isHalted() bool {
...
+		}
+	}
+
+	return updateResyncCounter(db, logger, name)
+}
+
+func updateResyncCounter(db *sql.DB, logger *log.Logger, name string) error {
+	ctx := context.Background()
+
+	_, err := db.ExecContext(ctx, `
```
temaniarpit27: is there no upsert command?

temaniarpit27 (approved):
LGTM

bridgesync/processor.go
rachit77: nope i don't see a direct upsert command available here

joanestebanr (commented):
I don't see how it works. I understand that we want to force a resync from 0.4 to 0.5? isn't? but I think that we are going to update from 0.4, insert a 1 in new table `resync_counter` so that match the internal const `ResyncCounterVersion`... so it's not going to be detected. 

I think that a best approach is use the dbCompatibility mechanism that will protect against incompatibility data and also will detect resync situations. 

Let me known if you want to discuss this topic

PR #711, opened 2025-07-07
feat: add `aggsender-verifier` `gRPC` service

goran-ethernal:
## Description

This PR adds a support to create a `gRPC` server and to register any services on it.

It also adds a new service called `AggsenderVerifier` that will act as a `beethoven` style permissionless `aggsender`, that a chain's aggsender will call to verify and sign a new certificate, before it sends it to the `agglayer`.

## `grpc` package

In the common `grpc` package, a new file is added called `server.go` which contains the `gRPC` `ServerConfig`:

```go
type ServerConfig struct {
	// Host is the address to bind the gRPC server
	Host string `mapstructure:"Host"`
	// Port is the port to bind the gRPC server
	Port int `mapstructure:"Port"`
	// EnableReflection indicates whether gRPC server reflection is enabled
	EnableReflection bool `mapstructure:"EnableReflection"`
}

```

And the functions to create a new `gRPC` server.

## Aggsender Validator

A new package is added in the `aggsender` main package, called `validator` that contains logic for the validator service as well as the `.proto` definition of the service.

The validator service only has one function, which will either return an error or not, based on the validity of the provided certificate:

```
service AggsenderValidator {
    rpc ValidateCertificate(ValidateCertificateRequest) returns (google.protobuf.Empty);
}

// Request to validate a certificate
message ValidateCertificateRequest {
  // Previous certificate identifier
  agglayer.node.types.v1.CertificateId previous_certificate_id = 1;
  // Certificate to be validated
  agglayer.node.types.v1.Certificate certificate = 2;
}

```

The validator service and the function to verify the certificate can be called using `grpcurl` for testing, once the server is up and running (note, you have to have reflection enabled to not have to provide the `.proto` files to the `grpcurl`):

```
grpcurl -plaintext \
  -d '{
    "previous_certificate_id": {
      // fill with fields from CertificateId, e.g. "id": "abc123"
    },
    "certificate": {
      // fill with fields from Certificate, e.g. "field1": "value1"
    }
  }' \
  localhost:50051 \
  aggkit.aggsender.validator.v1.AggsenderValidator/ValidateCertificate
```

The service uses the already defined `.proto` definition of a certificate from the [buf.build](https://buf.build/agglayer/agglayer/file/main:agglayer/node/types/v1/certificate.proto).

## `Buf.Build`
This PR also adds support for `buf.build` and code generation from `.proto` files.

`buf.build` supports adding dependencies to other repos and their `.proto` files using `buf.yaml` and `buf.gen.yaml` files, where we can specify on which repos we have dependencies for their `.proto` files (in this case `agglayer` since it has `certificate.proto`), without the need to clone the given repo.

Fixes # (issue)


grpc/server.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/agglayer/aggkit/log"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+)
+
+type ServerConfig struct {
+	// Host is the address to bind the gRPC server
+	Host string `mapstructure:"Host"`
+	// Port is the port to bind the gRPC server
+	Port int `mapstructure:"Port"`
```
joanestebanr, line 17: I see that this field after is concatenated to <Host>:<Port>... why don't set the port directly in the `Host` string?  WDTY?

goran-ethernal: I thought about this, but went with this approach, because, if you see the `default.go`, for any service we went with `Host` and `Port` as separated fields, so I followed the same pattern.

.gitignore
```diff
@@ -14,3 +14,4 @@ coverage.html
 .idea/*
 build/*
 data
+buf.lock
```
temaniarpit27: should we push this as well?
similar to go.sum?

aggsender/validator/proto/v1/validator.proto
```diff
@@ -0,0 +1,22 @@
...
+    rpc ValidateCertificate(ValidateCertificateRequest) returns (google.protobuf.Empty);
+}
+
+// Request to validate a certificate
+message ValidateCertificateRequest {
+  // Previous certificate identifier.
+  agglayer.node.types.v1.CertificateId previous_certificate_id = 1;
+  // Certificate to be validated
+  agglayer.node.types.v1.Certificate certificate = 2;
+}
```
temaniarpit27: new line at end of file?

buf.gen.yaml
```diff
@@ -0,0 +1,26 @@
+version: v2
+plugins:
+  - local: protoc-gen-go
+    out: .
+    opt:
+      - paths=source_relative
+      - Magglayer/interop/types/v1/bytes.proto=buf.build/gen/go/agglayer/interop/protocolbuffers/go/agglayer/interop/types/v1
```
temaniarpit27, line 7: `Magglayer`?

aggsender/validator/proto/v1/validator.proto
```diff
@@ -0,0 +1,22 @@
...
+
+// Service for validating new certificates
+service AggsenderValidator {
+    // Method to validate a new certificate
+    rpc ValidateCertificate(ValidateCertificateRequest) returns (google.protobuf.Empty);
+}
+
+// Request to validate a certificate
+message ValidateCertificateRequest {
+  // Previous certificate identifier.
```
Stefan-Ethernal: Nitpick: remove the dot because of consistency with the other comments
```suggestion
  // Previous certificate identifier
```

buf.gen.yaml
goran-ethernal: That is how you say to the `protoc-gen-go` to replace the proto namespace `this` with `this` `go` package namespace. You start with `M` and then provide the namespaces: `M{protoNamespace}=`goNamespace`.

Makefile
```diff
@@ -33,7 +33,10 @@ check-docker: ## Check if docker is installed
 # Check for Protoc
 .PHONY: check-protoc
 check-protoc: ## Check if protoc is installed
-	@which protoc > /dev/null || (echo "Error: Protoc is not installed" && exit 1)
+	@which protoc > /dev/null || (echo "Error: protoc is not installed" && exit 1)
```
Stefan-Ethernal: Consider adding a hint here as well (as it is done for the `protoc-gen-go`, `protoc-gen-go-grpc` etc.), how to install the `protoc`. WDYT?

.gitignore
goran-ethernal: Hmm, it probably makes sense. Let me include it.

goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/711/commits/b9904b4e7fe110ee31e193844da5da78d3176afd

aggsender/validator/proto/v1/validator.proto
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/711/commits/b9904b4e7fe110ee31e193844da5da78d3176afd

grpc/server.go
```diff
@@ -0,0 +1,87 @@
...
+	listener, err := net.Listen("tcp", addr)
+	if err != nil {
+		return nil, err
+	}
+
+	server := grpc.NewServer(opts...)
+
+	if cfg.EnableReflection {
+		reflection.Register(server) // Register reflection service on gRPC server
+	}
```
Stefan-Ethernal, line 50: WDYT about enabling this only in development/debugging environments? I'm suggesting to rely on logger parameter that has production and development values to determine which setup we are running currently?

grpc/server.go
```diff
@@ -0,0 +1,87 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+)
+
+type ServerConfig struct {
+	// Host is the address to bind the gRPC server
+	Host string `mapstructure:"Host"`
+	// Port is the port to bind the gRPC server
+	Port int `mapstructure:"Port"`
+	// EnableReflection indicates whether gRPC server reflection is enabled
```
Stefan-Ethernal, line 18: If we opt to keep this parameter, instead on relying on production vs development parameter, then it is worth explaining that it enables exposure of service methods, etc. WDYT?

grpc/server.go
```diff
@@ -0,0 +1,87 @@
...
+
+	cfg ServerConfig
+}
+
+// NewServer creates and returns a new gRPC Server instance listening on the specified address.
+// It accepts an address string, a boolean to enable server reflection, and optional gRPC server options.
+// If enableReflection is true, the server will register the reflection service for introspection.
+// Returns a pointer to the Server and an error if the listener cannot be created.
+func NewServer(cfg ServerConfig, opts ...grpc.ServerOption) (*Server, error) {
+	addr := fmt.Sprintf("%s:%d", cfg.Host, cfg.Port)
```
Stefan-Ethernal, line 38: Hmm I remember previously we were stripping the `http://` and `https://` prefixes https://github.com/agglayer/aggkit/blob/d34eb05ebb901737bcab1573b335262cb3089d5c/grpc/client.go#L254-L255

Is it relevant here as well?

Stefan-Ethernal (approved):

joanestebanr (commented):
The PR description no longer match the code (new param on service)

aggsender/validator/proto/v1/validator.proto
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/711/commits/8426eb824e2e249d622e420bc3f19ae221fbf9a0

Makefile
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/711/commits/8426eb824e2e249d622e420bc3f19ae221fbf9a0

grpc/server.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/711/commits/8426eb824e2e249d622e420bc3f19ae221fbf9a0

grpc/server.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/711/commits/8426eb824e2e249d622e420bc3f19ae221fbf9a0

grpc/server.go
```diff
@@ -0,0 +1,87 @@
...
+// It blocks until the provided context is cancelled, at which point it gracefully stops the server.
+// If the server fails to start, an error is logged.
+func (s *Server) Start(ctx context.Context) {
+	go func() {
+		<-ctx.Done()
+		s.stop()
+	}()
+
+	if err := s.grpcServer.Serve(s.listener); err != nil {
+		log.Errorf("failed to start gRPC server: %w", err)
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/711/commits/8426eb824e2e249d622e420bc3f19ae221fbf9a0

grpc/server.go
goran-ethernal: I would like to keep it. It is more easier and more flexible like this.

goran-ethernal:
> The PR description no longer match the code (new param on service)

Updated the description @joanestebanr .

joanestebanr (approved):

PR #720, opened 2025-07-08
refactor: simplify the `aggoracle` logic

Stefan-Ethernal:
## Description

- Simplify the aggoracle logic to rely on the latest indexed l1 info tree leaf and send the latest global exit root, instead of querying for the last finalized block number and extracting the global exit root up to that one.
- Also add some more logs and lower verbosity of certain logs to debug level.
- For L1 syncers (namely l1 info tree syncer and bridge l1 syncer), use `FinalizedBlock` commitment level always and remove it from configuration.
- Make GER injection asynchronous

### :wrench:  Config changes
- Obsolete `[AggOracle.BlockFinality]` parameter
- Obsolete `[L1InfoTreeSync.BlockFinality]` parameter
- Obsolete `[BridgeL1Sync.BlockFinality]` parameter

Fixes #494


l1infotreesync/processor.go
```diff
@@ -193,20 +190,33 @@ func (p *processor) GetLatestInfoUntilBlock(ctx context.Context, blockNum uint64
...
+	if blockNum != nil {
+		if *blockNum == 0 {
+			return nil, ErrNoBlock0
+		}
+		lpb, err := p.getLastProcessedBlockWithTx(tx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to retrieve the last processed block: %w", err)
+		}
+		if lpb < *blockNum {
+			return nil, ErrBlockNotProcessed
```
goran-ethernal, line 202: Do we have to return an error here? The function description says it will return the most recent `L1InfoTreeLeaf` that occurred before or at `blockNum`. If block is not yet processed, shouldn't we return what is the latest we have?

joanestebanr (commented):
Maybe `reorgDetectorL1`  can be removed also , WDTY?

l1infotreesync/processor.go
joanestebanr: The previous behaviour was returning the same error `ErrBlockNotProcessed`, if it's changed maybe we can have unexpected bug, WDYT?

Stefan-Ethernal: I would probably leave it like this, because it was the behavior from before. Although the both ways have sense to some extent. However, probably this approach comes as a less surprise (IMO it is better to return an error, in case certain target block is not indexed than to return a wrong data).

Stefan-Ethernal:
> Maybe `reorgDetectorL1` can be removed also , WDTY?

Since this requires some more refactorings, I would leave it for some future work (#737), if you agree.

rachit77 (approved):
LGTM 💯

On a side note:
You might want to take a look at the [aggOracle docs](https://github.com/agglayer/aggkit/blob/fix/always-query-for-latest-block-that-defines-upper-bound-for-ger/docs/aggoracle.md). It may need an update to reflect this refactor — at the very least, we should add a note mentioning that aggOracle now uses the finalized block by default.

Stefan-Ethernal:
> LGTM 💯
> 
> On a side note: You might want to take a look at the [aggOracle docs](https://github.com/agglayer/aggkit/blob/fix/always-query-for-latest-block-that-defines-upper-bound-for-ger/docs/aggoracle.md). It may need an update to reflect this refactor — at the very least, we should add a note mentioning that aggOracle now uses the finalized block by default.

Thanks for reminder. Fixed https://github.com/agglayer/aggkit/pull/720/commits/2f141037e7aa77145264e2ad945f8801ae0b718c

l1infotreesync/processor.go
goran-ethernal: To me it just doesn't make sense, because in the query, we also say `WHERE block_num <= $1` which means, it will get the latest it has until that block.

goran-ethernal (approved):
LGTM 💯

PR #721, opened 2025-07-09
feat: certificate header cache

goran-ethernal:
## Description

This PR adds a `CertificateHeader` cache which will be used by the `aggsender validator` to retrieve previously settled certificates.

The `CertificateHeaderCache` is an in memory cache, that uses the `Time To Live` cache in the background, which can be configured to store `n` number of certificates in memory, and each item will be held for the configured amount of time.

After the configured time expires, the item gets removed from the cache.

If the capacity is hit on the cache when getting a new item, the oldest item will get evicted from the cache, and the new one will be added to it.

Fixes # (issue)


aggsender/validator/cache/certificate_cache.go
```diff
@@ -0,0 +1,70 @@
...
+// NewCertificateCache creates and returns a new CertificateCache instance with the specified
+// Agglayer client, time-to-live (TTL) duration for cache entries, and maximum cache capacity.
+// The cache stores certificate headers indexed by their hash and automatically evicts entries
+// based on the provided TTL and capacity constraints.
+func NewCertificateCache(
+	agglayerClient agglayer.AgglayerClientInterface,
+	ttl time.Duration,
+	capacity uint64) *CertificateCache {
+	c := ttlcache.New(
+		ttlcache.WithTTL[common.Hash, *agglayertypes.CertificateHeader](ttl),
```
joanestebanr: I think that is better to store on cache the obj not a pointer, because if we have a pointer, some other parts of code can modify it and we would have some bugs, WDTY?

aggsender/validator/cache/certificate_cache.go
```diff
@@ -0,0 +1,70 @@
...
+// Parameters:
+//   - ctx: The context for controlling cancellation and deadlines.
+//   - certificateID: The unique identifier (hash) of the certificate.
+//
+// Returns:
+//   - *agglayertypes.CertificateHeader: The retrieved certificate header.
+//   - error: An error if the certificate header could not be retrieved.
+func (c *CertificateCache) GetCertificateHeader(
+	ctx context.Context, certificateID common.Hash) (*agglayertypes.CertificateHeader, error) {
+	if c.cache.Has(certificateID) {
```
joanestebanr: When are discarded old objects?

aggsender/validator/cache/certificate_cache.go
```diff
@@ -0,0 +1,70 @@
...
+		return c.cache.Get(certificateID).Value(), nil
+	}
+
+	certificateHeader, err := c.agglayerClient.GetCertificateHeader(ctx, certificateID)
```
joanestebanr: make sense to cache also the 'not found' answer from Agglayer to avoid spamming? WDYT?

aggsender/validator/cache/certificate_cache.go
```diff
@@ -0,0 +1,70 @@
...
+// Agglayer client, time-to-live (TTL) duration for cache entries, and maximum cache capacity.
+// The cache stores certificate headers indexed by their hash and automatically evicts entries
+// based on the provided TTL and capacity constraints.
+func NewCertificateCache(
+	agglayerClient agglayer.AgglayerClientInterface,
+	ttl time.Duration,
+	capacity uint64) *CertificateCache {
+	c := ttlcache.New(
+		ttlcache.WithTTL[common.Hash, *agglayertypes.CertificateHeader](ttl),
+		ttlcache.WithCapacity[common.Hash, *agglayertypes.CertificateHeader](capacity),
```
joanestebanr: `capaticy` params is bytes or num of items?

goran-ethernal: Number of items.

aggsender/validator/cache/certificate_cache.go
goran-ethernal: They are discarded in two ways:
- lazy discarding - whenever you call functions like `cache.Has`, `cache.Get`, `cache.Set` it will remove items that are expired.
- automatic discarding - by calling the `cache.Start()` function, which I didn't do in this case, but rely on lazy discarding.

aggsender/validator/cache/certificate_cache.go
goran-ethernal: Hmm, since it is basically `map[certificateID]CertificateHeader`, I would need to store this in a separate cache or map. Which makes sense, but, I wouldn't overcomplicate it. I think we will need some rate limiter on top of our `gRPC` servers in the near future to stop malicious actors spamming the validator with calls. I think that would be a more cleaner solution.

aggsender/validator/cache/certificate_cache.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/721/commits/d51b6b00870cb0f50ba7336dad96132b7458ae50

aggsender/validator/cache/certificate_cache.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	certificateHeader, err := c.agglayerClient.GetCertificateHeader(ctx, certificateID)
+	if err != nil {
+		return agglayertypes.CertificateHeader{}, err
+	}
+
+	// if DefaultTTL is set, the cache will use the TTL from the cache configuration
+	// defined in the NewCertificateCache function
+	c.cache.Set(certificateID, *certificateHeader, ttlcache.DefaultTTL)
```
Stefan-Ethernal: Do you need to provide here the ttl parameter retrieved from the [configuration](https://github.com/agglayer/aggkit/blob/d51b6b00870cb0f50ba7336dad96132b7458ae50/aggsender/validator/cache/certificate_cache.go#L29)?

As a side note, is this ttl which we are providing here, an override per item in case we want to store it shorter or longer than the [global cache config](https://github.com/agglayer/aggkit/blob/d51b6b00870cb0f50ba7336dad96132b7458ae50/aggsender/validator/cache/certificate_cache.go)?

aggsender/validator/cache/certificate_cache.go
```diff
@@ -0,0 +1,70 @@
...
+//   - ctx: The context for controlling cancellation and deadlines.
+//   - certificateID: The unique identifier (hash) of the certificate.
+//
+// Returns:
+//   - *agglayertypes.CertificateHeader: The retrieved certificate header.
+//   - error: An error if the certificate header could not be retrieved.
+func (c *CertificateCache) GetCertificateHeader(
+	ctx context.Context, certificateID common.Hash) (agglayertypes.CertificateHeader, error) {
+	if c.cache.Has(certificateID) {
+		return c.cache.Get(certificateID).Value(), nil
```
Stefan-Ethernal: Is it safe to invoke the `Get` straight away, without checking if it exists?

joanestebanr (approved):
Maybe doesn't make too much sense to have that in folder `aggsender/validator/cache` because is generic and we can use for regular usage also, WDTY?

goran-ethernal:
> Maybe doesn't make too much sense to have that in folder `aggsender/validator/cache` because is generic and we can use for regular usage also, WDTY?

@joanestebanr  I will move it to the `aggsender/cache` folder.

aggsender/validator/cache/certificate_cache.go
Stefan-Ethernal: EDIT: It seems that the comment above was added after I left this comment, so I think this confirms my second question, so everything is fine!

goran-ethernal: Yes, the ttl parameter will come from the `Aggsender Validator` configuration, which we will add in subsequent PRs.

This `ttlcache.DefaultTTL` is just used, as a way to tell the cache, if TTL is not provided by the config, just use the default one. But if it is provided, it will use that one, no matter what we pass to the `.Set()` function.

aggsender/validator/cache/certificate_cache.go
goran-ethernal: It will return a nil value if doesn't find it. So it is more cleaner to me to ask if it has the item, since the cache is basically just a map, so it should be fast, plus, you see, once the `Get()` function returns a value, it needs to be accessed through the `.Value()` function which will break with nil pointer exception if `Get()` returns nil.

Stefan-Ethernal (approved):
:100:

PR #739, opened 2025-07-11
fix: DFS search

rachit77:
## Description

This PR  fixes bugs in syncer and adds reentrancy test.
e2e test: https://github.com/agglayer/e2e/pull/99


temaniarpit27 (approved):

.github/test_e2e_single_chain_fork12_op_succinct_args.json
```diff
@@ -15,12 +15,11 @@
     "op_succinct_max_concurrent_witness_gen": "1",
     "op_succinct_range_proof_interval": "60",
     "additional_services": [
-      "bridge_spammer"
```
Stefan-Ethernal: As a heads up, we are relying on `bridge_spammer` in the optimistic-mode e2e tests (https://github.com/agglayer/e2e/blob/cb516a358d66bccd229504467d41001d56a864b0/tests/op/optimistic-mode.bats#L19-L42). The test is currently disabled as it discovered the issue on the aggkit prover. Can we make your test work with the `bridge_spammer` enabled...?

bridgesync/downloader_test.go
```diff
@@ -326,28 +328,139 @@ func TestFindCall(t *testing.T) {
...
+		},
+	}
+
+	// Test that findCall continues searching and finds the first valid claim method
+	found, err := findCall(rootCall, bridgeAddr, func(call call) (bool, error) {
+		// Simulate tryDecodeClaimCalldata behavior
+		if len(call.Input) < 4 {
+			return false, fmt.Errorf("input too short")
+		}
+		methodID := call.Input[:4]
```
Stefan-Ethernal:
```suggestion
		if len(call.Input) < methodIDLength {
			return false, fmt.Errorf("input too short")
		}
		methodID := call.Input[:methodIDLength]
```

bridgesync/downloader_test.go
```diff
@@ -326,28 +328,139 @@ func TestFindCall(t *testing.T) {
...
+
+		if bytes.Equal(methodID, claimAssetEtrogMethodID) || bytes.Equal(methodID, claimMessageEtrogMethodID) {
+			return true, nil // Found a valid claim method
+		}
+
+		// Unrecognized method ID - return false, nil to continue searching
+		return false, nil
```
Stefan-Ethernal: WDYT, to reduce a couple of lines of code?
```suggestion
		isClaimInvoked := bytes.Equal(methodID, claimAssetEtrogMethodID) || bytes.Equal(methodID, claimMessageEtrogMethodID)

		return isClaimInvoked, nil
```

Stefan-Ethernal (approved):
Generally LGTM, but be mindful about `bridge_spammer`. Can we expose it, but in case they interfere with your tests, disable it in your test only (or even better if we can make it work independently from `bridge_spammer`). WDYT?

.github/test_e2e_single_chain_fork12_op_succinct_args.json
rachit77: why `bridge_spammer` is needed for optimistic-mode e2e tests ?
@Stefan-Ethernal

rachit77: or perhaps we should enable the bridge spammer only for the optimistic-mode E2E tests, since it's only required by optimistic mode

Stefan-Ethernal: I haven't implemented those tests myself, but I think the @jhkimqd would like to have some load on the network and it is easier to piggyback on the `bridge_spammer` and let it run automatically.

Are we confident that in case we are disabling the `bridge_spammer`, we are not hiding some issue on the aggkit side?

rachit77: i don't think removing `bridge_spammer` at this point would have any impact on other component. 
Moreover i will add task to investigate make `bridge_spammer` work with existing tests.
wdys?

PR #741, opened 2025-07-14
feat: integrate `aggoracle` committee

rachit77:
#732 

## Description

New Smart Contract PR: https://github.com/agglayer/agglayer-contracts/pull/499/files

### Config Changes
- Added `EnableAggOracleCommittee = false` to `[AggOracle]` section
- Added `AggOracleCommitteeAddr = "{{L2Config.AggOracleCommitteeAddr}}"` to `[AggOracle.EVMSender]` section

### Purpose
These configs enable switching between direct GER injection mode (default) and committee-based validation mode. When `EnableAggOracleCommittee` is true, GER proposals must be validated by committee members before injecting to `GlobalExitRootManagerL2SovereignChain.sol` contract.

### Usage
```toml
[AggOracle]
EnableAggOracleCommittee = true  # Enable committee mode
  [AggOracle.EVMSender]
    AggOracleCommitteeAddr = "0x..." # Committee contract address
```


cmd/run.go
```diff
@@ -384,6 +386,7 @@ func createAggoracle(
 		l1Client,
 		l1InfoTreeSyncer,
 		cfg.AggOracle.WaitPeriodNextGER.Duration,
+		cfg.AggOracle.EnableAggOracleQuorum,
```
temaniarpit27: `EnableAggOracleQuorum` -> `EnableAggOracleCommittee`

test/helpers/simulated.go
```diff
@@ -171,14 +174,26 @@ func NewSimulatedBackend(t *testing.T,
 	balances[deployerAuth.From] = types.Account{Balance: balance}
 	balances[precalculatedBridgeAddr] = types.Account{Balance: balance}
 
+	var precalculatedAggOracleCommitteeAddr common.Address
```
goran-ethernal:
```suggestion
	var (
	         precalculatedAggOracleCommitteeAddr common.Address
	         precalculatedAggOracleCommitteeProxyAddr common.Address
	 )
```

bridgesync/docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 services:
   test-claimdata-l1:
     container_name: test-claimdata-l1
-    image: hermeznetwork/geth-zkevm-contracts:elderberry-fork.9-geth1.13.11
+    image: hermeznetwork/geth-zkevm-contracts:latest
```
goran-ethernal, line 4: Do we want to always get the latest image of the contracts? Seems like it might cause issues with the test if something changes in the contracts.

aggoracle/chaingersender/evm.go
```diff
@@ -41,48 +64,92 @@ type EVMChainGERSender struct {
 func NewEVMChainGERSender(
 	logger *log.Logger,
 	l2GERManagerAddr common.Address,
+	aggOracleCommitteeAddr common.Address,
 	l2Client aggkittypes.BaseEthereumClienter,
 	ethTxMan types.EthTxManager,
 	gasOffset uint64,
 	waitPeriodMonitorTx time.Duration,
+	enableAggOracleCommittee bool,
```
goran-ethernal, line 73: Why not directly pass the mode?

aggoracle/chaingersender/evm.go
```diff
@@ -99,32 +166,89 @@ func (c *EVMChainGERSender) IsGERInjected(ger common.Hash) (bool, error) {
...
+
+	lastProposedGER, err := c.aggOracleCommittee.AddressToLastProposedGER(
+		&bind.CallOpts{Pending: false}, c.ethTxMan.From())
+	if err != nil {
+		return false, fmt.Errorf("failed to check last proposed GER for oracle committee member %s: %w",
+			c.ethTxMan.From(), err)
+	}
+
+	lastProposedGERHash := common.Hash(lastProposedGER)
+	return lastProposedGERHash == ger, nil
```
goran-ethernal, line 178: What happens if you propose a GER-1, then propose GER-2, then try to propose again GER-1? Its not something that will happen with honest actors, but what happens if someone tries to do this as a malicious actor?

aggoracle/e2e_test.go
```diff
@@ -12,11 +12,12 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestEVM(t *testing.T) {
+func TestEVM_DirectInjectionMode(t *testing.T) {
 	setup := helpers.NewE2EEnvWithEVML2(t, helpers.DefaultEnvironmentConfig())
 
-	for i := 0; i < 10; i++ {
-		_, err := setup.L1Environment.GERContract.UpdateExitRoot(setup.L1Environment.Auth, common.HexToHash(strconv.Itoa(i)))
+	for i := 2; i < 12; i++ {
```
goran-ethernal: Just curious, why go from 2 to 12?

aggoracle/chaingersender/evm.go
jesusholoo: https://github.com/agglayer/specs/blob/5eb68833d1254a134efc85bc9594d0822f21f7fb/specs/contracts.md?plain=1#L962

aggoracle/chaingersender/evm.go
```diff
@@ -99,32 +166,89 @@ func (c *EVMChainGERSender) IsGERInjected(ger common.Hash) (bool, error) {
 	return gerIndex.Cmp(common.Big0) == 1, nil
 }
 
+// IsGERProposed checks if the provided global exit root has already been proposed by the oracle committee member
+func (c *EVMChainGERSender) IsGERProposed(ger common.Hash) (bool, error) {
```
jesusholoo, line 169: you coudl check if the GER is consolidated as well, on the globalExitROotMnaager contract. Then you hold more cases

aggoracle/e2e_test.go
```diff
@@ -31,3 +32,59 @@ func TestEVM(t *testing.T) {
...
+		// wait for the GER to be processed by the InfoTree syncer
+		time.Sleep(time.Millisecond * 500)
+		expectedGER, err := setup.L1Environment.GERContract.GetLastGlobalExitRoot(&bind.CallOpts{Pending: false})
+		require.NoError(t, err)
+
+		isInjected, err := setup.L2Environment.AggoracleSender.IsGERInjected(expectedGER)
+		require.NoError(t, err)
+
+		// fetch proposedGERToReport from committee contract
+		proposedGERToReport, err := setup.L2Environment.AggOracleCommitteeContract.ProposedGERToReport(nil, expectedGER)
```
jesusholoo: shouldnt be better to check the previosu events?¿ instead of checking that something is 0

bridgesync/docker-compose.yml
jesusholoo: is this docker used nowadays!?

aggoracle/e2e_test.go
rachit77: 0 to 10 will also work.
I will make the loop from 0 to 10

aggoracle/chaingersender/evm.go
rachit77: we are checking that here https://github.com/agglayer/aggkit/blob/07fe727d08669208ece65b1b1330760fca2c8004/aggoracle/oracle.go#L86-L87

The [IsGERInjected](https://github.com/agglayer/aggkit/blob/07fe727d08669208ece65b1b1330760fca2c8004/aggoracle/chaingersender/evm.go#L158-L167) functions checks if the GER has been already consolidated on globalExitROotMnaager contract or not

aggoracle/e2e_test.go
rachit77: which event do you mean? @invocamanman 
are you referring to use `addressToLastProposedGER` ?

aggoracle/chaingersender/evm.go
rachit77: I don’t have a strong preference here — I just thought using a boolean in the config file might be more intuitive and user-friendly.

rachit77: as discussed will keep this as boolean as it is right now

bridgesync/docker-compose.yml
rachit77: @invocamanman not sure but it doesn't seem like this docker is used nowdays atleast from docker hub.
and apart from latest tag the second most recent image is [fork.11-geth1.13.11](https://hub.docker.com/layers/hermeznetwork/geth-zkevm-contracts/fork.11-geth1.13.11/images/sha256-879220aa137c9e83524dd71ef5df810a01e68a9735f948c15dac6cc0c0e9ff1e)

aggoracle/chaingersender/evm.go
jesusholoo: So my point is, why is not checked if the ger is already injected here `ProposeGER`
Put it whathever you want, but the check should be there

aggoracle/e2e_test.go
jesusholoo: events from contracts.

```
        emit ProposedGlobalExitRoot(proposedGlobalExitRoot, msg.sender);
        emit ConsolidatedGlobalExitRoot(globalExitRoot);
```

aggoracle/chaingersender/evm.go
```diff
@@ -99,32 +166,89 @@ func (c *EVMChainGERSender) IsGERInjected(ger common.Hash) (bool, error) {
...
+			c.ethTxMan.From(), err)
+	}
+
+	lastProposedGERHash := common.Hash(lastProposedGER)
+	return lastProposedGERHash == ger, nil
+}
+
 // InjectGER injects the provided global exit root into the L2 GER manager contract
 func (c *EVMChainGERSender) InjectGER(ctx context.Context, ger common.Hash) error {
+	if c.mode != DirectInjectionMode {
```
temaniarpit27: i dont think this check is needed

aggoracle/chaingersender/evm.go
```diff
@@ -99,32 +166,89 @@ func (c *EVMChainGERSender) IsGERInjected(ger common.Hash) (bool, error) {
...
+		return fmt.Errorf("InjectGER is only available in direct injection mode, current mode: %s", c.mode)
+	}
+
+	return c.submitTransaction(ctx, &c.l2GERManagerAddr, c.l2GERManagerAbi, insertGERFuncName, ger, "inject")
+}
+
+// ProposeGER proposes the provided global exit root to the AggOracleCommittee contract
+func (c *EVMChainGERSender) ProposeGER(ctx context.Context, ger common.Hash) error {
+	if c.mode != AggOracleCommitteeMode {
```
temaniarpit27: i dont think this check is needed

aggoracle/e2e_test.go
```diff
@@ -12,11 +12,12 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestEVM(t *testing.T) {
+func TestEVM_DirectInjectionMode(t *testing.T) {
```
temaniarpit27: check if this and below test can be combined

aggoracle/e2e_test.go
```diff
@@ -31,3 +32,59 @@ func TestEVM(t *testing.T) {
...
 	}
 }
+
+func TestEVM_AggOracleCommitteeMode(t *testing.T) {
+	cfg := helpers.DefaultEnvironmentConfig()
+	cfg.AggoraclecommitteeConfig.EnableAggOracleCommittee = true
+	setup := helpers.NewE2EEnvWithEVML2(t, cfg)
+
+	for i := 0; i < 10; i++ {
+		rootHash := common.HexToHash(strconv.Itoa(i))
```
temaniarpit27: i = 1 wont collide with the initial proposed ger?

config/default.go
```diff
@@ -32,7 +32,7 @@ genesisBlockNumber = 0
 
 [L2Config]
 	GlobalExitRootAddr = "0x0000000000000000000000000000000000000000"
-
+	AggOracleCommitteeAddr = "0x0000000000000000000000000000000000000000"
```
temaniarpit27, line 35: do we need this here if we already have config below?

aggoracle/types/types.go
```diff
@@ -43,3 +43,17 @@ type L2GERManagerContract interface {
 	)
 	GlobalExitRootUpdater(opts *bind.CallOpts) (common.Address, error)
 }
+
+// AggOracleCommitteeContract is an interface to interact with the AggOracleCommittee contract
+type AggOracleCommitteeContract interface {
+	ProposeGlobalExitRoot(opts *bind.TransactOpts, proposedGlobalExitRoot [32]byte) (*types.Transaction, error)
```
Stefan-Ethernal: This is unused, should we remove it? If not, let's add the comment, documenting it.

```suggestion
```

aggoracle/e2e_test.go
rachit77: nopes as `rootHash` is not GER

GER = Hash(RER, MER)

config/default.go
rachit77: I'd prefer to consolidate all contract addresses into a single location

wdyt?

bridgesync/docker-compose.yml
temaniarpit27: we need a latest image of the same. There are only 2 tags present and used `latest`.
https://hub.docker.com/r/hermeznetwork/geth-zkevm-contracts/tags

aggoracle/chaingersender/evm.go
temaniarpit27: done

aggoracle/e2e_test.go
temaniarpit27: i guess its fine since we are checking if its injected or not as well

temaniarpit27 (approved):

aggoracle/chaingersender/evm.go
```diff
@@ -41,48 +64,92 @@ type EVMChainGERSender struct {
...
 
-// validateGERSender validates whether the provided GER sender is allowed to send and remove GERs
-func validateGERSender(gerSender common.Address, l2GERManagerSC types.L2GERManagerContract) error {
-	zeroAddr := common.Address{}
-	gerUpdater, err := l2GERManagerSC.GlobalExitRootUpdater(nil)
+// initializeMode initializes mode-specific components and validations
+func (c *EVMChainGERSender) initializeMode() error {
+	switch c.mode {
+	case DirectInjectionMode:
+		return c.initializeDirectInjectionMode()
```
Stefan-Ethernal: We probably don't need the `initializeDirectInjectionMode`, and can be simplified this way. WDYT?

```suggestion
		return validateGERSender(c.ethTxMan.From(), c.l2GERManager)
```

aggoracle/chaingersender/evm.go
```diff
@@ -41,48 +64,92 @@ type EVMChainGERSender struct {
...
+
+	c.aggOracleCommittee = aggOracleCommittee
+	c.aggOracleCommitteeAbi = aggOracleCommitteeAbi
+
+	// Validate GER proposer
+	if err := validateGERProposer(c.ethTxMan.From(), aggOracleCommittee); err != nil {
+		return err
 	}
 
 	return nil
```
Stefan-Ethernal: Let's validate early and if we detect any issues, it is not necessary to fetch the abi and set the fields, WDYT?

```suggestion
	aggOracleCommittee, err := aggoraclecommittee.NewAggoraclecommittee(c.aggOracleCommitteeAddr, c.l2Client)
	if err != nil {
		return fmt.Errorf("failed to create binding for AggOracleCommittee (SC address: %s): %w",
			c.aggOracleCommitteeAddr, err)
	}

    // Validate GER proposer
	if err := validateGERProposer(c.ethTxMan.From(), aggOracleCommittee); err != nil {
		return err
	}

	// Get the ABI for AggOracleCommittee
	aggOracleCommitteeAbi, err := aggoraclecommittee.AggoraclecommitteeMetaData.GetAbi()
	if err != nil {
		return fmt.Errorf("failed to retrieve AggOracleCommittee ABI: %w", err)
	}

	c.aggOracleCommittee = aggOracleCommittee
	c.aggOracleCommitteeAbi = aggOracleCommitteeAbi

	return nil
```

aggoracle/chaingersender/evm.go
```diff
@@ -41,48 +64,92 @@ type EVMChainGERSender struct {
...
+		return fmt.Errorf("unknown GER mode: %s", c.mode)
+	}
+}
+
+func (c *EVMChainGERSender) initializeDirectInjectionMode() error {
+	if err := validateGERSender(c.ethTxMan.From(), c.l2GERManager); err != nil {
+		return err
+	}
+	return nil
+}
```
Stefan-Ethernal: I would remove this and invoke the `validateGERSender` instead. WDYT?
```suggestion
```

aggoracle/chaingersender/evm.go
```diff
@@ -99,32 +166,87 @@ func (c *EVMChainGERSender) IsGERInjected(ger common.Hash) (bool, error) {
 	return gerIndex.Cmp(common.Big0) == 1, nil
 }
 
+// IsGERProposed checks if the provided global exit root has already been proposed by the oracle committee member
+func (c *EVMChainGERSender) IsGERProposed(ger common.Hash) (bool, error) {
+	isInjected, err := c.IsGERInjected(ger)
+	if err != nil {
+		return false, fmt.Errorf("failed to check if global exit root is injected %s: %w", ger, err)
```
Stefan-Ethernal: It doesn't make sense to provide the same error message wrapped twice, so probably just propagate the retrieved error instead. WDYT?

```suggestion
		return false, err
```

aggoracle/chaingersender/evm.go
```diff
@@ -99,32 +166,87 @@ func (c *EVMChainGERSender) IsGERInjected(ger common.Hash) (bool, error) {
...
+
+// ProposeGER proposes the provided global exit root to the AggOracleCommittee contract
+func (c *EVMChainGERSender) ProposeGER(ctx context.Context, ger common.Hash) error {
+	// Check if the GER has already been proposed by the oracle committee member
+	isProposed, err := c.IsGERProposed(ger)
+	if err != nil {
+		return err
+	}
+	if isProposed {
+		c.logger.Infof("GER %s has already been proposed by the oracle committee member", ger.Hex())
```
Stefan-Ethernal:
```suggestion
		c.logger.Infof("GER %s has already been proposed by the aggoracle committee member", ger.Hex())
```

aggoracle/chaingersender/evm.go
```diff
@@ -99,32 +166,87 @@ func (c *EVMChainGERSender) IsGERInjected(ger common.Hash) (bool, error) {
...
+			c.ethTxMan.From(), err)
+	}
+
+	lastProposedGERHash := common.Hash(lastProposedGER)
+	return lastProposedGERHash == ger, nil
+}
+
 // InjectGER injects the provided global exit root into the L2 GER manager contract
 func (c *EVMChainGERSender) InjectGER(ctx context.Context, ger common.Hash) error {
+	return c.submitTransaction(ctx, &c.l2GERManagerAddr, c.l2GERManagerAbi, insertGERFuncName, ger, "inject")
```
Stefan-Ethernal, line 192: Consider moving this [check](https://github.com/agglayer/aggkit/blob/e306654226e3711be77a5c45adb187c955c17f1e/aggoracle/oracle.go#L86-L94) about is ger injected here to the `InjectGER`, so that the `ProposeGER` and `InjectGER` are as similar as possible (since they are using the pretty similar thing). WDYT?

aggoracle/config.go
```diff
@@ -16,8 +16,9 @@ var (
...
 type Config struct {
-	TargetChainType   TargetChainType          `mapstructure:"TargetChainType"`
-	URLRPCL1          string                   `mapstructure:"URLRPCL1"`
-	WaitPeriodNextGER types.Duration           `mapstructure:"WaitPeriodNextGER"`
-	EVMSender         chaingersender.EVMConfig `mapstructure:"EVMSender"`
+	TargetChainType          TargetChainType          `mapstructure:"TargetChainType"`
+	URLRPCL1                 string                   `mapstructure:"URLRPCL1"`
+	WaitPeriodNextGER        types.Duration           `mapstructure:"WaitPeriodNextGER"`
+	EVMSender                chaingersender.EVMConfig `mapstructure:"EVMSender"`
+	EnableAggOracleCommittee bool                     `mapstructure:"EnableAggOracleCommittee"`
```
Stefan-Ethernal, line 23: Thinking out loud:
do we need this parameter? Isn't it safe to asume that if the [AggOracleCommitteeAddr](https://github.com/agglayer/aggkit/blob/e306654226e3711be77a5c45adb187c955c17f1e/aggoracle/chaingersender/evm.go#L28) is not `0x0`, we are working in the `comittee` mode?

The aggkit config is pretty complex already and errors will and can happen easily, so IMO it is better if we can derive some parameters automatically, but also enforce the configuration to be valid.

aggoracle/chaingersender/evm.go
Stefan-Ethernal: Consider to mark the mode as `aggoracle_committee` in case `aggOracleCommitteeAddr != 0x0`. WDYT?

aggoracle/types/types.go
```diff
@@ -43,3 +43,17 @@ type L2GERManagerContract interface {
...
+
+	// View method to check oracle manager state
+	GetAggOracleMemberIndex(opts *bind.CallOpts, oracleMember common.Address) (*big.Int, error)
+
+	// View method to check the last proposed GER by an oracle member
+	AddressToLastProposedGER(opts *bind.CallOpts, oracleMember common.Address) ([32]byte, error)
+
+	// View method to get all oracle committee members
+	GetAllAggOracleMembers(opts *bind.CallOpts) ([]common.Address, error)
```
Stefan-Ethernal: Unused, remove it. WDYT?
```suggestion
```

test/helpers/e2e.go
```diff
@@ -60,20 +67,26 @@ type L1Environment struct {
...
+	GERContract                *globalexitrootmanagerl2sovereignchain.Globalexitrootmanagerl2sovereignchain
+	AggOracleCommitteeContract *aggoraclecommittee.Aggoraclecommittee
+	AggoracleSender            aggoracle.ChainSender
+	EthTxManagerMock           *EthTxManager
 }
 
 type EnvironmentConfig struct {
 	L1RPCClient aggkittypes.RPCClienter
 	L2RPCClient aggkittypes.RPCClienter
+	AggoraclecommitteeConfig
```
Stefan-Ethernal: Why we are embedding this struct? IMO it is cumbersome and would make sense to have a field associated. WDYT?


```suggestion
	aggoracleCommitteeCfg AggoraclecommitteeConfig
```

test/helpers/e2e.go
```diff
@@ -343,5 +368,48 @@ func newSimulatedEVML2SovereignChain(t *testing.T) (
 	require.NoError(t, err)
 	require.Equal(t, gerProxyAddr, bridgeGERAddr)
 
-	return client, setup.UserAuth, gerProxyAddr, gerL2Contract, setup.BridgeProxyAddr, setup.BridgeProxyContract
+	// Deploy AggOracleCommittee contract if EnableAggOracleCommittee is true
+	var aggOracleCommitteeProxyAddr common.Address
+	var aggOracleCommitteeContract *aggoraclecommittee.Aggoraclecommittee
```
Stefan-Ethernal: WDYT?
```suggestion
	var (
		aggOracleCommitteeProxyAddr common.Address
		aggOracleCommitteeContract  *aggoraclecommittee.Aggoraclecommittee
	)
```

test/helpers/e2e.go
```diff
@@ -343,5 +368,48 @@ func newSimulatedEVML2SovereignChain(t *testing.T) (
...
+		aggOracleCommitteeAbi, err := aggoraclecommittee.AggoraclecommitteeMetaData.GetAbi()
+		require.NoError(t, err)
+		require.NotNil(t, aggOracleCommitteeAbi)
+
+		// aggOracleMembers are the addresses of the aggoracle committee members
+		aggOracleMembers := []common.Address{setup.UserAuth.From}
+
+		// add other aggoracle committee members to the aggOracleMembers slice based on the quorum
+		for i := 0; i < int(aggOracleCommitteeConfig.Quorum)-1; i++ {
+			aggOracleMembers = append(aggOracleMembers, crypto.CreateAddress(setup.DeployerAuth.From, aggOracleCommitteeNonce+uint64(i+1)))
```
Stefan-Ethernal: Linter complains for me here that the line is too long


```suggestion
			aggOracleMembers = append(aggOracleMembers,
				crypto.CreateAddress(setup.DeployerAuth.From, aggOracleCommitteeNonce+uint64(i+1)))
```

test/helpers/e2e.go
```diff
@@ -343,5 +368,48 @@ func newSimulatedEVML2SovereignChain(t *testing.T) (
...
+		}
+
+		aggOracleCommitteeInitData, err := aggOracleCommitteeAbi.Pack("initialize", setup.DeployerAuth.From, aggOracleMembers, aggOracleCommitteeConfig.Quorum)
```
Stefan-Ethernal: Warning that the line is too long, consider wrapping it for better readability.

```suggestion
		aggOracleCommitteeInitData, err := aggOracleCommitteeAbi.Pack(
			"initialize",
			setup.DeployerAuth.From,
			aggOracleMembers,
			aggOracleCommitteeConfig.Quorum)
```

test/helpers/e2e.go
```diff
@@ -303,7 +320,7 @@ func newSimulatedEVML2SovereignChain(t *testing.T) (
 	genesisAllocMap := map[common.Address]types.Account{
 		l2BridgeProxyAddr: {Balance: premineBalance},
 	}
-	client, setup := NewSimulatedBackend(t, genesisAllocMap, deployerAuth)
+	client, setup := NewSimulatedBackend(t, genesisAllocMap, deployerAuth, aggOracleCommitteeConfig.EnableAggOracleCommittee)
```
Stefan-Ethernal: Consider wrapping the line, because linter complains.
```suggestion
	client, setup := NewSimulatedBackend(t, genesisAllocMap, deployerAuth, 
    aggOracleCommitteeConfig.EnableAggOracleCommittee)
```

test/helpers/e2e.go
```diff
@@ -184,14 +197,14 @@ func L2Setup(t *testing.T, cfg *EnvironmentConfig) *L2Environment {
 	t.Helper()
 
 	l2Client, authL2, gerL2Addr, gerL2Contract,
-		bridgeL2Addr, bridgeL2Contract := newSimulatedEVML2SovereignChain(t)
+		bridgeL2Addr, bridgeL2Contract, aggOracleCommitteeAddr, aggOracleCommitteeContract := newSimulatedEVML2SovereignChain(t, cfg.AggoraclecommitteeConfig)
```
Stefan-Ethernal: Linter complains about line being too long.
```suggestion
	l2Client, authL2, gerL2Addr, gerL2Contract,
		bridgeL2Addr, bridgeL2Contract,
		aggOracleCommitteeAddr, aggOracleCommitteeContract := newSimulatedEVML2SovereignChain(t, cfg.AggoraclecommitteeConfig)
```

test/helpers/e2e.go
```diff
@@ -343,5 +368,48 @@ func newSimulatedEVML2SovereignChain(t *testing.T) (
...
+		require.NoError(t, err)
+		client.Commit()
+
+		// Create aggoracle committee contract binding
+		aggOracleCommitteeContract, err = aggoraclecommittee.NewAggoraclecommittee(
+			aggOracleCommitteeProxyAddr, client.Client())
+		require.NoError(t, err)
+	}
+
+	return client, setup.UserAuth, gerProxyAddr, gerL2Contract, setup.BridgeProxyAddr, setup.BridgeProxyContract, aggOracleCommitteeProxyAddr, aggOracleCommitteeContract
```
Stefan-Ethernal: Consider wrapping the line, as it is too long and linter in VSC complains

```suggestion
	return client, setup.UserAuth, gerProxyAddr, gerL2Contract,
		setup.BridgeProxyAddr, setup.BridgeProxyContract,
		aggOracleCommitteeProxyAddr, aggOracleCommitteeContract
```

Stefan-Ethernal (approved):
Generally LGTM, leaving a couple of comments to consider.

aggoracle/types/types.go
temaniarpit27: done

test/helpers/simulated.go
```diff
@@ -171,14 +174,29 @@ func NewSimulatedBackend(t *testing.T,
...
+		precalculatedAggOracleCommitteeProxyAddr common.Address
+	)
+
+	if enableAggOracleCommittee {
+		// Create aggoracle committee address from deployerAuth.From and nonce = 4
+		precalculatedAggOracleCommitteeAddr = crypto.CreateAddress(deployerAuth.From, aggOracleCommitteeNonce)
+		precalculatedAggOracleCommitteeProxyAddr = crypto.CreateAddress(deployerAuth.From, aggOracleCommitteeNonce+1)
+		balances[precalculatedAggOracleCommitteeAddr] = types.Account{Balance: balance}
+		balances[precalculatedAggOracleCommitteeProxyAddr] = types.Account{Balance: balance}
+	}
```
Stefan-Ethernal: This can be handled from the outside of example [here](https://github.com/agglayer/aggkit/blob/e306654226e3711be77a5c45adb187c955c17f1e/test/helpers/e2e.go#L323), precalculate addresses and premine it, as it is only necessary for the L2 setup.

aggoracle/chaingersender/evm.go
temaniarpit27: done

aggoracle/chaingersender/evm.go
temaniarpit27: done

test/helpers/simulated.go
```diff
@@ -65,10 +69,12 @@ func (t *TestClient) Call(result any, method string, args ...any) error {
...
-	UserAuth            *bind.TransactOpts
-	DeployerAuth        *bind.TransactOpts
-	BridgeProxyAddr     common.Address
-	BridgeProxyContract *polygonzkevmbridgev2.Polygonzkevmbridgev2
+	UserAuth                    *bind.TransactOpts
+	DeployerAuth                *bind.TransactOpts
+	BridgeProxyAddr             common.Address
+	BridgeProxyContract         *polygonzkevmbridgev2.Polygonzkevmbridgev2
+	AggOracleCommitteeAddr      common.Address
+	AggOracleCommitteeProxyAddr common.Address
```
Stefan-Ethernal: I don't see this belongs here, it should be related only for L2 setup/environment. Can we figure out if we can remove it as my suggestion is to calculate it in the `newSimulatedEVML2SovereignChain` function instead?

```suggestion
```

aggoracle/chaingersender/evm.go
temaniarpit27: done

aggoracle/types/types.go
temaniarpit27: done

test/helpers/e2e.go
temaniarpit27: done

test/helpers/e2e.go
temaniarpit27: done

test/helpers/e2e.go
temaniarpit27: done

test/helpers/e2e.go
temaniarpit27: done

test/helpers/e2e.go
temaniarpit27: done

test/helpers/e2e.go
temaniarpit27: done

aggoracle/chaingersender/evm.go
temaniarpit27: done

test/helpers/e2e.go
temaniarpit27: done

test/helpers/simulated.go
temaniarpit27: done

aggoracle/config.go
temaniarpit27: was using this var to make sure deployer use this feature explicity. wanted to avoid the situation where mistakingly they add something here and thus creates deployment issues.

aggoracle/chaingersender/evm.go
temaniarpit27: done

test/helpers/simulated.go
temaniarpit27: done

aggoracle/config.go
Stefan-Ethernal: In that case, to me it would make sense to ensure that if this value is set to true, we make sure that the `AggoracleComitteeAddr` is non zero address and to try querying some function to make sure that the correct contract is deployed. Wdyt?

temaniarpit27: done

aggoracle/chaingersender/evm.go
temaniarpit27: using bool and checking if addr is != '0x0`

aggoracle/chaingersender/evm.go
```diff
@@ -41,50 +65,92 @@ type EVMChainGERSender struct {
...
 	gasOffset uint64,
 	waitPeriodMonitorTx time.Duration,
+	enableAggOracleCommittee bool,
 ) (*EVMChainGERSender, error) {
+	// Determine mode based on configuration
+	mode := DirectInjectionMode
+	if enableAggOracleCommittee && aggOracleCommitteeAddr != aggkitcommon.ZeroAddress {
+		mode = AggOracleCommitteeMode
+	}
+
```
goran-ethernal, line 81: I would add a log here that says which mode is selected and why.

aggoracle/oracle.go
```diff
@@ -78,36 +82,23 @@ func (a *AggOracle) processLatestGER(ctx context.Context) error {
...
-	if isGERInjected {
-		a.logger.Debugf("GER (%s) is already injected", latestGER.Hex())
-		return nil
-	}
-
 	go func() {
-		if err := a.injectGER(ctx, latestGER); err != nil {
+		// Submit GER based on mode
+		var err error
+		if a.enableAggOracleCommittee {
```
goran-ethernal: We introduce this check in the `chainSender`:

```go
mode := DirectInjectionMode
if enableAggOracleCommittee && aggOracleCommitteeAddr != aggkitcommon.ZeroAddress {
    mode = AggOracleCommitteeMode
}
```

I don't think it is enough to say `if a.enableAggOracleCommittee` here. We might be in the inject mode.

I propose having a function called `ProcessGER` on the `chainSender`, which will based on its mode either call `ProposeGER` or `InjectGER`. I don't think oracle should know which mode on the sender is enabled.

temaniarpit27: done

aggoracle/chaingersender/evm.go
temaniarpit27: done

goran-ethernal (approved):
LGTM 💯

PR #743, opened 2025-07-14
refactor: l2 ger syncer and resolve l2 ger sync mode automatically

### LINKED ISSUE #736: refactor l2 ger syncer and automatically resolve the proper l2 ger downloader

Stefan-Ethernal:
## 🔄 Changes Summary
- Rename the last ger syncer to l2 ger syncer.
- Move the `L2EVMGERReader` to the `l2gersync` (component that on demand reads the inserted and removed global exit roots from L2 GER Manager contract)
- Automatically detect which `SyncMode` does l2 ger syncer needs to be run, based on the deployed l2 ger manager contract.
- The legacy evm downloader in the l2 ger syncer does not store the empty blocks anymore (namely the blocks without the GER updates. Info about block number and block position is extracted from the `L1InfoTreeLeaf`).
- Refactor the component tests framework to accommodate deployment of legacy l2 ger manager contract and test `evmdownloader_legacy.go`

## ⚠️ Breaking Changes
- 🛠️ **Config**:
    - Rename `LastGERSync` to `L2GERSync`
    - Remove `LastGERSync.SyncMode` parameter (as it is determined automatically, by the deployed L2 GER manager contract)
- 🔌 **API/CLI**:
    - the L2 GER syncer database is renamed to `l2gersync.sqlite`, and this would cause the re-indexing. It is safe to remove the `lastgersync.sqlite`, as it is not in use anymore, after upgrading to this version.
    - the database migration files are also renamed, to be aligned with the db name. In case the aggkit fails to start after the upgrade (because of the rename), the remediation should be deletion of `lastgersync.sqlite` and reindexing the content into `l2gersync.sqlite`.
    - the `block_pos` column is [added](https://github.com/agglayer/aggkit/blob/eeaaf058d560296745e537c91ebae4af586cf79e/l2gersync/migrations/l2gersync0003.sql#L5) to the `imported_global_exit_root` table

## 📋 Config Updates
- 🧾 **Diff/Config snippet**:
    - [Diff](https://github.com/agglayer/aggkit/pull/743/files#diff-61cfb7e001b72e449356ab5369f659461155cdd3c1b162d0ada0529a836d8374)

    - Example config section
```toml
[L2GERSync]
    DBPath = "{{PathRWData}}/l2gersync.sqlite"
    BlockFinality = "LatestBlock"
    InitialBlockNum = 0
    GlobalExitRootL2Addr = "{{L2Config.GlobalExitRootAddr}}"
    RetryAfterErrorPeriod = "1s"
    WaitForNewBlocksPeriod = "1s"
    DownloadBufferSize = 100
```

## :link: Related PRs
- https://github.com/0xPolygon/kurtosis-cdk/pull/709

Fixes #736


l2gersync/l2gersyncer.go
```diff
@@ -0,0 +1,175 @@
...
+		globalexitrootmanagerl2sovereignchain.NewGlobalexitrootmanagerl2sovereignchain(address, backend)
+	if err == nil {
+		updater, err := sovereignGERManager.GlobalExitRootUpdater(&bind.CallOpts{Context: ctx})
+		if err == nil {
+			log.Debugf("Detected: GlobalExitRootManagerL2SovereignChain, GlobalExitRootUpdater = %s", updater.Hex())
+			return SovereignChain, nil
+		}
+
+		if !strings.Contains(err.Error(), gethvm.ErrExecutionReverted.Error()) &&
+			!strings.Contains(err.Error(), "function selector was not recognized") {
```
goran-ethernal: Maybe move this "function selector was not recognized" to a string constant.

l2gersync/l2gersyncer.go
```diff
@@ -0,0 +1,175 @@
...
+	sovereignGERManager, err :=
+		globalexitrootmanagerl2sovereignchain.NewGlobalexitrootmanagerl2sovereignchain(address, backend)
+	if err == nil {
+		updater, err := sovereignGERManager.GlobalExitRootUpdater(&bind.CallOpts{Context: ctx})
+		if err == nil {
+			log.Debugf("Detected: GlobalExitRootManagerL2SovereignChain, GlobalExitRootUpdater = %s", updater.Hex())
+			return SovereignChain, nil
+		}
+
+		if !strings.Contains(err.Error(), gethvm.ErrExecutionReverted.Error()) &&
```
goran-ethernal: What happens if it is not one of these two errors?

l2gersync/migrations/lastgersync0003.sql
```diff
@@ -0,0 +1,5 @@
+-- +migrate Down
+ALTER TABLE imported_global_exit_root DROP COLUMN block_pos;
+
+-- +migrate Up
+ALTER TABLE imported_global_exit_root ADD block_pos INTEGER NULL;
```
goran-ethernal: Thinking out loud, do we need a way to update existing events in the `db`, or since you changed the file name, maybe we need to resync from scratch?

l2gersync/l2gersyncer.go
Stefan-Ethernal: In fact, I'm going to remove this, because it doesn't really makes sense.

The logic is as following:
- in case there is `execution reverted` error message received, it means that the `GlobalExitRootManagerL2SovereignChain` is not deployed
- therefore we need to fallback and try with `PolygonZkEVMGlobalExitRootv2` (aka legacy L2 GER manager contract)

l2gersync/l2gersyncer.go
Stefan-Ethernal: The logic is as following:

- in case there is execution reverted error message received, it means that the `GlobalExitRootManagerL2SovereignChain` is not deployed
- therefore we need to fallback and try with `PolygonZkEVMGlobalExitRootv2` (aka legacy L2 GER manager contract)

l2gersync/migrations/lastgersync0003.sql
Stefan-Ethernal: > do we need a way to update existing events in the db

First of all, the `block_pos` was added because of this part in the [aggsender](https://github.com/agglayer/aggkit/blob/fe95ab0a3f31f249456d9a64c31e937db1311b07/aggsender/query/ger_query.go#L68). You are probably right, we might not need to resync entire database, but to drop the latest N entries (which are required for the currently pending certificate that is about to be sent). Now I'm unsure how to communicate this to the chain operators :thinking_face:

> or since you changed the file name, maybe we need to resync from scratch?

I haven't changed the name of the migration file, because of the upgradeability issues, I have just changed the name of the root folder, but I suppose it doesn't matter. It would be much nicer if I could rename the migration file name (which I did at first), but because the applied migrations are persisted by the migration file names, I found that it probably is not a safe to do so.

goran-ethernal: I eas talking about the default db file name. If someone was using the old db, and it didnt specify the name, we will create a new one with new name, which will make it resync everything from scratch, which is good, so we can have block position populated.

cmd/run.go
```diff
@@ -574,18 +582,18 @@ func runReorgDetectorL2IfNeeded(
...
 	ctx context.Context,
 	components []string,
-	cfg lastgersync.Config,
+	cfg l2gersync.Config,
 	reorgDetectorL2 *reorgdetector.ReorgDetector,
 	l2Client aggkittypes.BaseEthereumClienter,
 	l1InfoTreeSync *l1infotreesync.L1InfoTreeSync,
-) *lastgersync.LastGERSync {
+) *l2gersync.L2GERSync {
 	if !isNeeded([]string{aggkitcommon.BRIDGE}, components) {
```
joanestebanr, line 615: Maybe  you have to update the components that require it

config/config.go
```diff
@@ -192,6 +194,14 @@ var (
 			FieldNamePattern: "BridgeL1Sync.BlockFinality",
 			Reason:           bridgeL1SyncBlockFinalityDeprecated,
 		},
+		{
+			FieldNamePattern: "LastGERSync.SyncMode",
```
joanestebanr, line 200: make sense to include as bad field `L2GERSync.SyncMode`?

goran-ethernal (approved):

config/config.go
Stefan-Ethernal: If you mean to deprecate the `L2GERSync.SyncMode`, it doesn't make sense, because through this PR I have introduced the concept named `L2GERSync` (and it wasn't existing before). WDYT?

cmd/run.go
Stefan-Ethernal: We have agreed to decouple the `AggSender` (and related components, such as `AggsenderValidator` and `AggchainProofTool`) from the l2 ger syncer (especially its database):
- if possible provide a common component for the l2 ger syncer and the aggsender which will index the GER insertion and removal events,
- therefore, we wouldn't need to expand this condition for the `AggSender` (and its related components)

aggoracle/e2e_test.go
```diff
@@ -13,19 +13,19 @@ import (
 )
 
 func TestEVM(t *testing.T) {
-	setup := helpers.NewE2EEnvWithEVML2(t, helpers.DefaultEnvironmentConfig())
+	l1Setup, l2Setup := helpers.NewL1SetupWithL2EVM(t, helpers.DefaultEnvironmentConfig(helpers.SovereignChainL2GERContract))
```
temaniarpit27: NewL1SetupWithL2EVM - maybe a better name for this?

config/config.go
```diff
@@ -68,6 +68,8 @@ const (
 	aggOracleBlockFinalityDeprecated      = "AggOracle.BlockFinality is deprecated, remove it from configuration"
 	l1InfoTreeSyncBlockFinalityDeprecated = "L1InfoTreeSync.BlockFinality is deprecated, remove it from configuration"
 	bridgeL1SyncBlockFinalityDeprecated   = "BridgeL1Sync.BlockFinality is deprecated, remove it from configuration"
+	lastGERSyncDeprecatedHint             = "LastGERSync is deprecated, use L2GERSync instead"
+	lastGERSyncSyncModeDeprecatedHint     = "LastGERSync.SyncMode is deprecated, remove it from configuration"
 )
 
 type DeprecatedFieldsError struct {
```
temaniarpit27: any particular reason we have this?
if we are providing updated config, then maybe completely remove this?
i know its a safe check to keep but not a fan of this. we need to maintain it

temaniarpit27 (approved):
💯 
do we need to make any changes in kurtosis?

Stefan-Ethernal:
> do we need to make any changes in kurtosis?

Yes we do, here is the kurtosis cdk [PR](https://github.com/0xPolygon/kurtosis-cdk/pull/709).

config/config.go
Stefan-Ethernal: > if we are providing updated config, then maybe completely remove this?

We are just providing the defaults, but it can be overridden. Although I agree that it can be tedious to maintain it and it will just grow over time, it is a nice way to fail fast in case configuration is invalid in a transparent way.

aggoracle/e2e_test.go
Stefan-Ethernal: WDYT about `NewSimulatedEnvironmentEVM`? The reason why it was named like this is because we were believing that in the future we might have non-EVM L2 chains and we would need to have separate setups for it

Stefan-Ethernal: Renamed to `NewSimulatedEVMEnvironment` https://github.com/agglayer/aggkit/pull/743/commits/0eb2aa6922cea38182a70b2309b01d36d5db0d91

l2gersync/evm_downloader_legacy.go
```diff
@@ -172,21 +191,17 @@ func (d *downloaderFEP) getGERsFromIndex(
 		if err != nil {
 			return nil, fmt.Errorf("failed to retrieve l1 info tree leaf for index=%d: %w", i, err)
 		}
-		gers = append(gers,
-			&GlobalExitRootInfo{
-				L1InfoTreeIndex: i,
-				GlobalExitRoot:  info.GlobalExitRoot,
-			})
+		gers = append(gers, newGlobalExitRootInfo(info.GlobalExitRoot, i, blockNum, 0))
```
goran-ethernal:
```suggestion
		gers = append(gers, newGlobalExitRootInfo(info.GlobalExitRoot, i, blockNum, info.BlockPosition))
```

goran-ethernal: Why use passed `blockNum` if `info` already has that?

l2gersync/processor.go
```diff
@@ -198,7 +175,7 @@ func (p *processor) GetFirstGERAfterL1InfoTreeIndex(
 	ctx context.Context, l1InfoTreeIndex uint32) (GlobalExitRootInfo, error) {
 	e := GlobalExitRootInfo{}
 	err := meddler.QueryRow(p.database, &e, `
-		SELECT l1_info_tree_index, global_exit_root
+		SELECT l1_info_tree_index, block_num, global_exit_root
```
goran-ethernal: Are you missing block position?

l2gersync/evm_downloader_legacy.go
Stefan-Ethernal: Very good point, I have reused the block number and block position from L1InfoTreeLeaf (which is created by indexing the `UpdateL1InfoTree`/`UpdateL1InfoTreeV2` events). Also changed the logic a bit, so it doesn't report the empty blocks in l2 ger syncers (which I believe doesn't make sense) https://github.com/agglayer/aggkit/pull/743/commits/1deab4898a211274d3302c4b7c09fcd583c82242

Stefan-Ethernal: Very good point, I have reused the block number and block position from L1InfoTreeLeaf (which is created by indexing the `UpdateL1InfoTree`/`UpdateL1InfoTreeV2` events). Also changed the logic a bit, so it doesn't report the empty blocks in l2 ger syncers (which I believe doesn't make sense) https://github.com/agglayer/aggkit/pull/743/commits/1deab4898a211274d3302c4b7c09fcd583c82242

l2gersync/processor.go
Stefan-Ethernal: Yes, thank you. Fixed https://github.com/agglayer/aggkit/pull/743/commits/67a1d9a621de698f9a4831e45baf85dbc47cb42f

l2gersync/l2_ger_syncer.go
```diff
@@ -0,0 +1,173 @@
...
+func (s *L2GERSync) GetFirstGERAfterL1InfoTreeIndex(
+	ctx context.Context, atOrAfterL1InfoTreeIndex uint32,
+) (GlobalExitRootInfo, error) {
+	return s.processor.GetFirstGERAfterL1InfoTreeIndex(ctx, atOrAfterL1InfoTreeIndex)
+}
+
+// GetInjectedGERsForRange retrieves all injected global exit roots within a specified block range.
+// It returns a map where the keys are the global exit root hashes and the values are the
+// corresponding GlobalExitRootInfo containing the L1 info tree index, global exit root and block number.
+func (s *L2GERSync) GetInjectedGERsForRange(ctx context.Context,
```
goran-ethernal, line 165: I think this is now not used anywhere, since you returned the old `GERReader` logic for `aggsender`.

l2gersync/evm_downloader_legacy.go
```diff
@@ -172,21 +200,17 @@ func (d *downloaderFEP) getGERsFromIndex(
...
-				GlobalExitRoot:  info.GlobalExitRoot,
-			})
+		gers = append(gers, newGlobalExitRootInfo(info.GlobalExitRoot, i, info.BlockNumber, info.BlockPosition))
 	}
 
 	return gers, nil
 }
 
-func (d *downloaderFEP) populateGreatestInjectedGER(b *sync.EVMBlock, gerInfos []*GlobalExitRootInfo) {
+func (d *downloaderLegacy) findLatestInjectedGER(gerInfos []*GlobalExitRootInfo) *GlobalExitRootInfo {
```
goran-ethernal, line 210: Are these GERs sorted somehow, by index or block? If not, I don't think our latest injected logic is good.

l2gersync/l2_ger_syncer.go
Stefan-Ethernal: It is easier to expose this function and [test](https://github.com/agglayer/aggkit/blob/3c8dc20afaca8fc83343126febb49708b6bb6279/l2gersync/e2e_test.go#L160) the legacy evm downloader that way, that's why I have opted to keep it (and who knows we might need it in future as well).

l2gersync/evm_downloader_legacy.go
Stefan-Ethernal: They are inherently sorted, since we are iterating from the latest l1 info tree leaf index up to the last root index:
- [for loop](https://github.com/agglayer/aggkit/blob/3c8dc20afaca8fc83343126febb49708b6bb6279/l2gersync/evm_downloader_legacy.go#L198)
- and [this](https://github.com/agglayer/aggkit/blob/3c8dc20afaca8fc83343126febb49708b6bb6279/tree/tree.go#L190) is how the latest root is retrieved (it is sorted by block number and block position in a descending order) 

However, the `findLatestInjectedGER` seems not to return the latest injected ger, but the first one which is injected. Let me try out iterating backwards from the most recent one, until older one and see if it works.

Stefan-Ethernal: Fixed, the latest GER is extracted by iterating the `gerInfos` in reverse order https://github.com/agglayer/aggkit/pull/743/commits/df6c1d27d3faa80e1d300e6c434832df1ba14934

PR #738, opened 2025-07-11
feat: call `aggsender validator` in regular `aggsender`

goran-ethernal:
## Description

This PR implements the real call to `Aggsender Validator` in the regular `Aggsender`.

It also adds the validator signature to the `submitCertificate` request to the `Agglayer` by putting it into the request headers under key: `x-agglayer-extra-certificate-signature`.

### ⚙️ Config Changes
- Added `RequireValidatorCall` to the `Aggsender` config which defines if a call to the `Aggsender Validator` is required before sending the certificate to the `Agglayer`.
- Added `ValidatorClient` config to setup `grpc` client to the `Aggsender Validator` which is in the form of all the `grpc` clients config:
```
[AggSender]
RequireValidatorCall = false
[AggSender.ValidatorClient]
URL = "validatorURL"
MinConnectTimeout = "5s"
RequestTimeout = "300s"
UseTLS = false
[AggSender.ValidatorClient.Retry]
	InitialBackoff = "1s"
	MaxBackoff = "10s"
	BackoffMultiplier = 2.0
	MaxAttempts = 20
```

Fixes #730

PR #744, opened 2025-07-14
feat: Implements grpc service for `aggsender-validator`

### LINKED ISSUE #729: aggsender: New component `aggsender-validator` with gRPC service
- New component `aggsender-validator`: gRPC service that receives a certificate + a previous certificate ID and returns a signature or an errors is fails the validation

joanestebanr:
## Description

- Implements grpc service for `aggsender-validator`
    - Integrate agglayer client cache
- Fix flaky test in aggsender/flow (time dependency)

## 📋 Configuration Updates
- Add new configuration for `validator`:
```
[Aggsender]
	[AggSender.AgglayerClient]
		Cached = true
			[ValidaAggSendertor.AgglayerClient.ConfigurationCache]
				TTL = "15m"
				Capacity = 100
			[AggSender.AgglayerClient.GRPC]
				URL = "{{AggLayerURL}}"
				MinConnectTimeout = "5s"
				RequestTimeout = "300s"
				UseTLS = false
				[AggSender.AgglayerClient.GRPC.Retry]
					InitialBackoff = "1s"
					MaxBackoff = "10s"
					BackoffMultiplier = 2.0
					MaxAttempts = 20

[Validator]
EnableRPC = true
Signer = { Method = "mock" }
MaxCertSize = "{{AggSender.MaxCertSize}}"
MaxL2BlockNumber = "{{AggSender.MaxL2BlockNumber}}"
DelayBetweenRetries = "{{AggSender.DelayBetweenRetries}}"
[Validator.ServerConfig]
	Host = "0.0.0.0"
	Port = 5578
	EnableReflection = true
        MaxDecodingMessageSize = 1073741824  # 1GB
[Validator.LerQuerierConfig]
	RollupManagerAddr = "{{AggSender.RollupManagerAddr}}"
	RollupCreationBlockL1 = "{{AggSender.RollupCreationBlockL1}}"
[Validator.PPConfig]
	RequireOneBridgeInPPCertificate = "{{AggSender.RequireOneBridgeInPPCertificate}}"
[Validator.AgglayerClient]
		Cached = true
		[Validator.AgglayerClient.ConfigurationCache]
			TTL = "15m"
			Capacity = 100
		[Validator.AgglayerClient.GRPC]
		URL = "{{AggSender.AgglayerClient.GRPC.URL}}"
		MinConnectTimeout = "{{AggSender.AgglayerClient.GRPC.MinConnectTimeout}}"
		RequestTimeout = "{{AggSender.AgglayerClient.GRPC.RequestTimeout}}"
		UseTLS = "{{AggSender.AgglayerClient.GRPC.UseTLS}}"
		[Validator.AgglayerClient.GRPC.Retry]
			InitialBackoff = "{{AggSender.AgglayerClient.GRPC.Retry.InitialBackoff}}"
			MaxBackoff = "{{AggSender.AgglayerClient.GRPC.Retry.MaxBackoff}}"
			BackoffMultiplier = "{{AggSender.AgglayerClient.GRPC.Retry.BackoffMultiplier}}"
			MaxAttempts = "{{AggSender.AgglayerClient.GRPC.Retry.MaxAttempts}}"
```

Fixes #729


agglayer/certificate_cache.go
```diff
@@ -52,19 +71,43 @@ func NewCertificateCache(
...
 	// if DefaultTTL is set, the cache will use the TTL from the cache configuration
 	// defined in the NewCertificateCache function
 	c.cache.Set(certificateID, *certificateHeader, ttlcache.DefaultTTL)
 
-	return *certificateHeader, nil
+	return certificateHeader, nil
+}
+
+// SendCertificate sends a certificate to the Agglayer client. (no cache)
+func (c *CertificateCache) SendCertificate(ctx context.Context,
```
goran-ethernal: I am not a fan of having the client functions exposed on the cache.

In your own words, to achieve a better concern separation, I would leave the cache to do only cache things, and if you need the client on other place, just have a reference there to the client.

agglayer/client.go
```diff
@@ -2,19 +2,14 @@ package agglayer
...
 
-	"github.com/0xPolygon/cdk-rpc/rpc"
-	rpcTypes "github.com/0xPolygon/cdk-rpc/types"
+	"github.com/agglayer/aggkit/agglayer/grpc"
 	"github.com/agglayer/aggkit/agglayer/types"
 	"github.com/ethereum/go-ethereum/common"
 )
 
-const errCodeAgglayerRateLimitExceeded int = -10007
+// NOTE:  errCodeAgglayerRateLimitExceeded is -10007
```
goran-ethernal, line 12: Remove?

agglayer/config.go
```diff
@@ -0,0 +1,31 @@
+package agglayer
+
+import (
+	"errors"
+
+	aggkitgrpc "github.com/agglayer/aggkit/grpc"
+)
+
+var (
+	ErrConfigurationCacheRequired = errors.New("configuration cache is required when Cached is true")
```
goran-ethernal:
```suggestion
	errConfigurationCacheRequired = errors.New("configuration cache is required when Cached is true")
```

agglayer/config.go
```diff
@@ -0,0 +1,31 @@
...
+	Cached             bool
+	ConfigurationCache *ConfigurationCache
+}
+
+// Validate checks if the client configuration is valid.
+func (c *ClientConfig) Validate() error {
+	if err := c.GRPC.Validate(); err != nil {
+		return err
+	}
+	if c.Cached && c.ConfigurationCache == nil {
```
goran-ethernal:
```suggestion
	if c.Cached {
	     if c.ConfigurationCache == nil {
	         return ErrConfigurationCacheRequired
	     }
	     
	     return c.ConfigurationCache.Validate()
         }
```

config/default.go
```diff
@@ -239,4 +239,35 @@ GlobalExitRootL2 = "{{L2Config.GlobalExitRootAddr}}"
...
+[Validator]
+EnableRPC = true
+Signer = { Method = "mock" }
+MaxCertSize = "{{AggSender.MaxCertSize}}"
+MaxL2BlockNumber = "{{AggSender.MaxL2BlockNumber}}"
+DelayBetweenRetries = "{{AggSender.DelayBetweenRetries}}"
+[Validator.ServerConfig]
+	Host = "0.0.0.0"
+	Port = 5578
+	EnableReflection = true
```
goran-ethernal, line 257: Should we  put `EnableRPC` and `EnableReflection` as `false` since these are only development flags?

aggsender/validator/validator_service.go
```diff
@@ -3,21 +3,124 @@ package validator
...
 	v1 "github.com/agglayer/aggkit/aggsender/validator/proto/v1"
+	"github.com/agglayer/aggkit/grpc"
 	"github.com/agglayer/aggkit/log"
+	signertypes "github.com/agglayer/go_signer/signer/types"
+	"github.com/ethereum/go-ethereum/common"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/protobuf/types/known/emptypb"
 )
 
+type AgglayerClientInterface = agglayer.AggLayerClientCertificateIDQuerier
```
goran-ethernal: Why do you create a new type? You already have `AgglayerClientInterface` with its mock in the `agglayer` package.

aggsender/validator/config.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/agglayer/aggkit/agglayer"
+	"github.com/agglayer/aggkit/config/types"
+	aggkitgrpc "github.com/agglayer/aggkit/grpc"
+	signertypes "github.com/agglayer/go_signer/signer/types"
+	ethCommon "github.com/ethereum/go-ethereum/common"
+)
+
+// Config defines the configuration for the aggsender validator service.
+type Config struct {
+	// EnableRPC is a flag to enable the RPC for aggsender
```
goran-ethernal:
```suggestion
	// EnableRPC is a flag to enable the RPC for validator
```

aggsender/validator/config.go
```diff
@@ -0,0 +1,48 @@
...
+	EnableRPC bool `mapstructure:"EnableRPC"`
+	// Signer is the key which is used to sign certificates
```
goran-ethernal:
```suggestion
	// Signer is the key which is used to sign valid certificates
```

aggsender/validator/config.go
```diff
@@ -0,0 +1,48 @@
...
+	// MaxL2BlockNumber is the last L2 block number that is going to be included in a certificate
+	// 0 means disabled
+	MaxL2BlockNumber uint64 `mapstructure:"MaxL2BlockNumber"`
+	// DelayBetweenRetries is the delay between retries:
+	//  is used on store Certificate and also in initial check
+	DelayBetweenRetries types.Duration `mapstructure:"DelayBetweenRetries"`
+	// LerQuerier contains the configuration for the LER querier
+	// which is used to query the LER data from the RollupManager contract
+	LerQuerier LerQuerierConfig `mapstructure:"PPConfig"`
+	// PPConfig specific configuration  for Pessimistic mode
```
goran-ethernal:
```suggestion
	// PPConfig specific configuration for Pessimistic mode
```

aggsender/validator/cert_hash.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+const (
+	u32Size = 4
+	u64Size = 8
+)
+
+func HashCertificateToSign(cert *agglayertypes.Certificate) common.Hash {
```
goran-ethernal, line 12:
```suggestion
// HashCertificateToSign is the hash of the certificate that the validator will sign before returning result to the aggsender
func HashCertificateToSign(cert *agglayertypes.Certificate) common.Hash {
```

agglayer/grpc/conversion_to_agglayer.go
```diff
@@ -0,0 +1,298 @@
...
+			"PrevLocalExitRoot, NewLocalExitRoot, or Metadata. %w", ErrNilCertificate)
+	}
+
+	if cert.L1InfoTreeLeafCount == nil {
+		return nil, fmt.Errorf("convertProtoCertToAgglayer.certificate has nil L1InfoTreeLeafCount. %w",
+			ErrNilCertificate)
+	}
+
+	if cert.AggchainData != nil {
+		return nil, fmt.Errorf("convertProtoCertToAgglayer. not supported AggchainData in incomming certificate. %w",
```
goran-ethernal: `AggchainData` is an interface that either contains:
- signature of the certificate - signed by the `aggsender` in `PP` mode.
- aggchain proof - generated by the `aggkit prover` in `FEP` mode.

This check needs to be modified to check if it is nil. If it is nil, return an error. Then, in the validator, check that this is a signature, and not a proof.

I think we should be able to convert to our types without a problem. I can help you with this.

agglayer/grpc/conversion_to_agglayer.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+	importedBridgeExits, err := grpcImportedBridgeExitsToAgglayer(cert.ImportedBridgeExits)
+	if err != nil {
+		return nil, fmt.Errorf("convertProtoCertToAgglayer. error converting grpc imported bridge exits: %w", err)
+	}
+
+	agglayerCert := &agglayertypes.Certificate{
+		NetworkID: cert.NetworkId,
+		Height:    cert.Height,
+		//CertificateID:         cert.CertificateId,
```
goran-ethernal: We can calculate it on our own, but I think we don't need it in the validator, so we can just remove this comment.

agglayer/grpc/conversion_to_agglayer.go
```diff
@@ -0,0 +1,298 @@
...
+	v1nodetypes "buf.build/gen/go/agglayer/agglayer/protocolbuffers/go/agglayer/node/types/v1"
+	v1types "buf.build/gen/go/agglayer/interop/protocolbuffers/go/agglayer/interop/types/v1"
+	agglayertypes "github.com/agglayer/aggkit/agglayer/types"
+	"github.com/agglayer/aggkit/bridgesync"
+	treetypes "github.com/agglayer/aggkit/tree/types"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+var (
+	ErrNotImplemented        = errors.New("aggsender-validator not implemented")
```
goran-ethernal: I think you don't need to export these errors.

```suggestion
	errNotImplemented        = errors.New("aggsender-validator not implemented")
	errNilCertificate            = errors.New("aggsender-validator nil certificate")
```

agglayer/grpc/conversion_to_agglayer.go
```diff
@@ -0,0 +1,298 @@
...
+	ErrNilCertificate        = errors.New("aggsender-validator nil certificate")
+	ErrMetadataNotCompatible = errors.New("aggsender-validator metadata not compatible with the current version")
```
goran-ethernal: This error is not used anywhere.

agglayer/grpc/conversion_to_agglayer.go
```diff
@@ -0,0 +1,298 @@
...
+	if bridgeExit.Amount != nil {
+		amount = amount.SetBytes(bridgeExit.Amount.Value)
+	}
+
+	var metadata []byte
+	if bridgeExit.Metadata != nil {
+		metadata = bridgeExit.Metadata.Value
+	}
+
+	agglayerBrdigeExit := &agglayertypes.BridgeExit{
```
goran-ethernal:
```suggestion
	agglayerBridgeExit := &agglayertypes.BridgeExit{
```

agglayer/grpc/conversion_to_agglayer.go
```diff
@@ -0,0 +1,298 @@
...
+		DestinationNetwork: bridgeExit.DestNetwork,
+		DestinationAddress: common.BytesToAddress(bridgeExit.DestAddress.Value),
+		Amount:             amount,
+		Metadata:           metadata,
+	}
+
+	return agglayerBrdigeExit, nil
+}
+
+// leafTypeToProto converts a leaf type to a proto leaf type
```
goran-ethernal:
```suggestion
// grpcLeafTypeToAgglayer converts a proto leaf type to a agglayer leaf type
```

agglayer/grpc/conversion_to_agglayer.go
```diff
@@ -0,0 +1,298 @@
...
+		ImportedBridgeExits: importedBridgeExits,
+	}
+
+	return agglayerCert, nil
+}
+
+func grpcBridgeExitsToAgglayer(
+	bridgeExits []*v1types.BridgeExit,
+) ([]*agglayertypes.BridgeExit, error) {
+	if bridgeExits == nil {
```
goran-ethernal:
```suggestion
	if len(bridgeExits) == 0 {
```

agglayer/grpc/conversion_to_agglayer.go
```diff
@@ -0,0 +1,298 @@
...
+		return agglayertypes.LeafTypeMessage, nil
+	default:
+		return agglayertypes.LeafTypeAsset, fmt.Errorf("unknown leaf type: %s", leafType)
+	}
+}
+
+func grpcImportedBridgeExitsToAgglayer(
+	importedBridgeExits []*v1types.ImportedBridgeExit,
+) ([]*agglayertypes.ImportedBridgeExit, error) {
+	if importedBridgeExits == nil {
```
goran-ethernal:
```suggestion
	if len(importedBridgeExits) == 0 {
```

agglayer/grpc/conversion_to_agglayer.go
```diff
@@ -0,0 +1,298 @@
...
+
+func grpcL1LeafToAgglayer(l1Leaf *v1types.L1InfoTreeLeafWithContext) (*agglayertypes.L1InfoTreeLeaf, error) {
+	if l1Leaf == nil {
+		return nil, fmt.Errorf("grpcL1LeafToAgglayer. l1 leaf is nil. %w", ErrNilCertificate)
+	}
+	if l1Leaf.Rer == nil || l1Leaf.Mer == nil {
+		return nil, fmt.Errorf("grpcL1LeafToAgglayer. l1 leaf has nil Rer or Mer. %w", ErrNilCertificate)
+	}
+	var inner *agglayertypes.L1InfoTreeLeafInner
+	if l1Leaf.Inner != nil {
```
goran-ethernal: I think `Inner` should never be nil. We should probably introduce a check for it.

aggsender/validator/config.go
```diff
@@ -0,0 +1,48 @@
...
+	MaxCertSize uint `mapstructure:"MaxCertSize"`
+	// MaxL2BlockNumber is the last L2 block number that is going to be included in a certificate
+	// 0 means disabled
+	MaxL2BlockNumber uint64 `mapstructure:"MaxL2BlockNumber"`
+	// DelayBetweenRetries is the delay between retries:
+	//  is used on store Certificate and also in initial check
+	DelayBetweenRetries types.Duration `mapstructure:"DelayBetweenRetries"`
+	// LerQuerier contains the configuration for the LER querier
+	// which is used to query the LER data from the RollupManager contract
+	LerQuerier LerQuerierConfig `mapstructure:"PPConfig"`
```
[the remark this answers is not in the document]
joanestebanr: 1db9010943a00f217d511ef4972e39de339c4dd2

agglayer/client.go
joanestebanr: Yes, it was unused code that lint complains about it, I left as a message just in case

agglayer/config.go
joanestebanr: For me make sense to expose the error, so any other module can use for check `ErrorIs`, WDTY?

agglayer/certificate_cache.go
joanestebanr: I have integrated the cache as a decorator for agglayer client and the creation of the client can automatically integrate the cache.  It make sense FMPOV because any part of the code that use a agglayer client can use the cache also (the idea was to integrate also in aggsender)
What do you mean? 
- You don't like that the New function can add the cache?
- You don't like that the cache act as a decorator of the client?
- You don't like that the movement of the cache to the `agglayer` module?
Please let me known, and we can discuss arround it to find the best fitting solution

agglayer/config.go
joanestebanr: Good point, I have move also the nil check to the `Validate()` function
fix: 021007efe471d0ee8b92fe01cc2d76f0a1a7562e

config/default.go
joanestebanr: good point, I agree about `EnableRPC` but I have doubts about `EnableReflection`... doesn't make sense to expose it to make more easy to use this service? or there are any performance issue?

aggsender/validator/config.go
joanestebanr: 77d9143f430dd707f0e606588e8368868d4d2acc

agglayer/certificate_cache.go
goran-ethernal: Then I would rename it to something else, it is no longer a cache. It has client behavior. We talked about this yesterday, and separation of concerns.

You can rename it to `AgglayerClientCached`, or something like that, but it is no longer only a cache.

agglayer/client.go
goran-ethernal: If it is not used anywhere, I would simply delete it. This looks like something that remained from the old `jsonRPC` logic.

config/default.go
goran-ethernal: Reflection is only enabled in development environments, as per google's specification and recommendations.

agglayer/grpc/conversion_to_agglayer.go
joanestebanr: good catch!, done b1492ff01b5ae302d1cc16d98c019043e6d94aa4

agglayer/grpc/conversion_to_agglayer.go
joanestebanr: I would appreciate your help!

agglayer/certificate_cache.go
joanestebanr: We agree to rename the cache to `AgglayerClientCache`

agglayer/grpc/conversion_to_agglayer.go
joanestebanr: We agree to expose errors as part of the interface

agglayer/grpc/conversion_to_agglayer.go
joanestebanr: removed `ErrMetadataNotCompatible` b117171aaaf69e02ad69eae2cb2fb7301e6a3a6a

config/default.go
joanestebanr: https://grpc.io/docs/guides/reflection/

> If your gRPC API is accessible to public users, you may not want to expose the reflection service, as you may consider this a security issue. Ultimately, you will need to make a call here that strikes the best balance between security and ease-of-use for you and your users.

agglayer/agglayer_client_cache.go
```diff
@@ -0,0 +1,123 @@
...
+	"fmt"
+	"time"
+
+	agglayertypes "github.com/agglayer/aggkit/agglayer/types"
+	"github.com/agglayer/aggkit/config/types"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/jellydator/ttlcache/v3"
+)
+
+type ConfigurationCache struct {
```
goran-ethernal: Maybe rename it to `CacheConfig`. The current name means like cache for configuration.

config/default.go
```diff
@@ -191,15 +191,20 @@ MaxL2BlockNumber = 0
...
-		MinConnectTimeout = "5s"
-		RequestTimeout = "300s"
-		UseTLS = false
-		[AggSender.AgglayerClient.Retry]
-			InitialBackoff = "1s"
-			MaxBackoff = "10s"
-			BackoffMultiplier = 2.0
-			MaxAttempts = 20
+		Cached = false
+		[ValidaAggSendertor.AgglayerClient.ConfigurationCache]
```
goran-ethernal:
```suggestion
		[AggSender.AgglayerClient.ConfigurationCache]
```

goran-ethernal (approved):

config/default.go
joanestebanr: 😅

grpc/server.go
```diff
@@ -18,6 +18,8 @@ type ServerConfig struct {
 	// EnableReflection indicates whether gRPC server reflection is enabled
 	// This allows clients to introspect the server's services and methods.
 	EnableReflection bool `mapstructure:"EnableReflection"`
+	// MaxRecvMsgSize is the maximum size of a message that the server can receive.
```
goran-ethernal:
```suggestion
	// MaxDecodingMessageSize is the maximum size of a message that the server can receive.
```

goran-ethernal (approved):

aggsender/validator/proto/v1/validator.proto
```diff
@@ -27,3 +37,5 @@ message ValidateCertificateResponse {
   // The signature provided by the aggsender validator.
    agglayer.interop.types.v1.FixedBytes65 signature = 1;
 }
+
+
```
Stefan-Ethernal: Nitpick, remove extra blank lines
```suggestion
```

grpc/server.go
```diff
@@ -18,6 +18,8 @@ type ServerConfig struct {
 	// EnableReflection indicates whether gRPC server reflection is enabled
 	// This allows clients to introspect the server's services and methods.
 	EnableReflection bool `mapstructure:"EnableReflection"`
+	// MaxDecodingMessageSize is the maximum size of a message that the server can receive.
+	MaxDecodingMessageSize uint64 `mapstructure:"MaxDecodingMessageSize"`
```
Stefan-Ethernal: WDYT? No need to use uint64, since the grpc library models this as `int` (probably it would make sense to put `int` as well in the config, even though negative values don't make sense, but that way we are secured by the overflows when downcasting from `uint64/uint` to `int`). WDYT?
```suggestion
	MaxDecodingMessageSize uint `mapstructure:"MaxDecodingMessageSize"`
```

config/default.go
```diff
@@ -239,4 +244,36 @@ GlobalExitRootL2 = "{{L2Config.GlobalExitRootAddr}}"
 ProfilingHost = "localhost"
 ProfilingPort = 6060
 ProfilingEnabled = false
+
+[Validator]
+EnableRPC = false
+Signer = { Method = "mock" }
```
Stefan-Ethernal: Can we add in the comment which values are supported for the `Method`, wdyt?

config/default.go
```diff
@@ -239,4 +244,36 @@ GlobalExitRootL2 = "{{L2Config.GlobalExitRootAddr}}"
...
+MaxCertSize = "{{AggSender.MaxCertSize}}"
+MaxL2BlockNumber = "{{AggSender.MaxL2BlockNumber}}"
+DelayBetweenRetries = "{{AggSender.DelayBetweenRetries}}"
+[Validator.ServerConfig]
+	Host = "0.0.0.0"
+	Port = 5578
+	EnableReflection = true
+	MaxDecodingMessageSize = 1073741824  # 1GB
```
Stefan-Ethernal: Just curious, what would be some typical sizes of certificates if you know?

aggsender/validator/local_validator.go
```diff
@@ -40,6 +40,13 @@ func (a *LocalValidator) String() string {
 	return "LocalValidator"
 }
```
Stefan-Ethernal, line 42: We are still planning to remove the LocalValidator implementation at some point, right?

cc @goran-ethernal

aggsender/validator/config.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/agglayer/aggkit/config/types"
+	aggkitgrpc "github.com/agglayer/aggkit/grpc"
+	signertypes "github.com/agglayer/go_signer/signer/types"
+	ethCommon "github.com/ethereum/go-ethereum/common"
+)
+
+// Config defines the configuration for the validator validator service.
+type Config struct {
+	// EnableRPC is a flag to enable the RPC for validator
+	EnableRPC bool `mapstructure:"EnableRPC"`
```
Stefan-Ethernal, line 14: This parameter is unused, at least for now. Are we going to need it?

aggsender/aggsender.go
```diff
@@ -172,6 +172,16 @@ func (a *AggSender) Start(ctx context.Context) {
 	if err := a.flow.CheckInitialStatus(ctx); err != nil {
 		a.log.Panicf("error checking flow Initial Status: %v", err)
 	}
+	if a.validator != nil {
+		status, err := a.validator.HealthCheck(ctx)
+		if err != nil {
+			a.log.Warnf("error checking validator health: %v", err)
+		}
+		if !status.IsHealthy() {
+			a.log.Warnf("validator health is not healthy: %s", status.String())
```
Stefan-Ethernal:
```suggestion
			a.log.Warnf("validator status is not healthy: %s", status.String())
```

aggsender/types/interfaces.go
```diff
@@ -177,13 +177,44 @@ type VerifyIncomingRequest struct {
...
+		", Version: " + h.Version + "}"
+}
+
 type CertificateValidator interface {
 	ValidateCertificate(ctx context.Context, params VerifyIncomingRequest) error
 }
 
 // CertificateValidateAndSigner is an interface to attach a certificate validator and signer
 // to aggsender regular flow
 type CertificateValidateAndSigner interface {
```
Stefan-Ethernal, line 215: WDYT, more concise IMO?

```suggestion
type CertificateValidator interface {
```

aggsender/aggsender.go
```diff
@@ -172,6 +172,16 @@ func (a *AggSender) Start(ctx context.Context) {
...
+		}
+		a.log.Infof("Validator health check: %s", status.String())
```
Stefan-Ethernal: Because other logs have the low case as well, wdyt?

```suggestion
		a.log.Infof("validator health check: %s", status.String())
```

agglayer/grpc/conversion_to_agglayer.go
```diff
@@ -0,0 +1,341 @@
...
+	v1nodetypes "buf.build/gen/go/agglayer/agglayer/protocolbuffers/go/agglayer/node/types/v1"
+	v1types "buf.build/gen/go/agglayer/interop/protocolbuffers/go/agglayer/interop/types/v1"
+	agglayertypes "github.com/agglayer/aggkit/agglayer/types"
+	"github.com/agglayer/aggkit/bridgesync"
+	treetypes "github.com/agglayer/aggkit/tree/types"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+var (
+	ErrNotImplemented = errors.New("aggsender-validator not implemented")
```
Stefan-Ethernal: Unused, remove it?

```suggestion
```

agglayer/client.go
Stefan-Ethernal: +1 for removal, I don't see any benefit in having this comment

aggsender/aggsender.go
```diff
@@ -282,26 +292,26 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certifi
...
-		a.log.Warnf("rate limit reached , next cert %s can be submitted after %s so sleeping. Rate:%s",
-			certificate.ID(),
-			rateLimitSleepTime.String(), a.rateLimiter.String())
-		time.Sleep(*rateLimitSleepTime)
-	}
-	a.log.Infof("certificate ready to be sent to AggLayer: %s start: %s , end: %s",
-		certificate.Brief(), startEpochStatus.String(), a.epochNotifier.GetEpochStatus().String())
-	metrics.CertificateBuildTime(time.Since(start).Seconds())
-
 	validatorSignature, err := a.callValidator(ctx, certificate)
```
Stefan-Ethernal, line 295: WDYT to rename the `callValidator` to `validateCertificate`?
```suggestion
	validatorSignature, err := a.validateCertificate(ctx, certificate)
```

aggsender/aggsender.go
```diff
@@ -282,26 +292,26 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certifi
...
 	if err != nil {
 		a.saveNonAcceptedCert(ctx, certificate, certificateParams.CreatedAt, err)
 		return nil, fmt.Errorf("certificate validation failed: %w", err)
 	}
+	a.log.Infof("certificate ready to be sent to AggLayer: %s start: %s , end: %s",
```
Stefan-Ethernal: Nitpick: remove extra space?

```suggestion
	a.log.Infof("certificate ready to be sent to AggLayer: %s start: %s, end: %s",
```

aggsender/aggsender.go
```diff
@@ -360,7 +370,7 @@ func (a *AggSender) callValidator(
 	if a.validator == nil {
 		return nil, nil
 	}
-
+	a.log.Infof("certificate sending to Validator: %s", certificate.Brief())
```
Stefan-Ethernal: WDYT?
```suggestion
	a.log.Infof("delegating certificate validation: %s", certificate.Brief())
```

aggsender/validator/cert_hash.go
```diff
@@ -0,0 +1,34 @@
...
+// before returning result to the aggsender
+func HashCertificateToSign(cert *agglayertypes.Certificate) common.Hash {
+	globalIndexHashes := make([][]byte, len(cert.ImportedBridgeExits))
+	for i, importedBridgeExit := range cert.ImportedBridgeExits {
+		globalIndexHashes[i] = importedBridgeExit.GlobalIndex.Hash().Bytes()
+	}
+	networkID := make([]byte, u32Size)
+	binary.BigEndian.PutUint32(networkID, cert.NetworkID)
+	height := make([]byte, u64Size)
+	binary.BigEndian.PutUint64(height, cert.Height)
```
Stefan-Ethernal: WDYT? (helper function is implemented here: `aggkitcommon "github.com/agglayer/aggkit/common"`)
```suggestion
	height := aggkitcommon.Uint64ToBytes(cert.Height)
```

aggsender/validator/cert_hash.go
```diff
@@ -0,0 +1,34 @@
...
+
+// HashCertificateToSign is the hash of the certificate that the validator will sign
+// before returning result to the aggsender
+func HashCertificateToSign(cert *agglayertypes.Certificate) common.Hash {
+	globalIndexHashes := make([][]byte, len(cert.ImportedBridgeExits))
+	for i, importedBridgeExit := range cert.ImportedBridgeExits {
+		globalIndexHashes[i] = importedBridgeExit.GlobalIndex.Hash().Bytes()
+	}
+	networkID := make([]byte, u32Size)
+	binary.BigEndian.PutUint32(networkID, cert.NetworkID)
```
Stefan-Ethernal: WDYT?
```suggestion
	networkID := aggkitcommon.Uint32ToBytes(cert.NetworkID)
```

aggsender/validator/cert_hash.go
```diff
@@ -0,0 +1,34 @@
...
+
+	agglayertypes "github.com/agglayer/aggkit/agglayer/types"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+const (
+	u32Size = 4
+	u64Size = 8
+)
```
Stefan-Ethernal: This can be removed in case we opt to use the helper functions defined here:
- https://github.com/agglayer/aggkit/blob/02876ab3617544b0c162fd2f764ae8a59e027bea/common/common.go#L53-L59
- https://github.com/agglayer/aggkit/blob/02876ab3617544b0c162fd2f764ae8a59e027bea/common/common.go#L26-L32

aggsender/validator/cert_hash.go
Stefan-Ethernal: Would it make sense to move this function to the [Certificate](https://github.com/agglayer/aggkit/blob/02876ab3617544b0c162fd2f764ae8a59e027bea/agglayer/types/types.go#L251) implementation? It could be a method instead of helper function. WDYT?

```go
func (c *Certificate) HashToSign() common.Hash{
    // TODO: implementation
}
```

aggsender/flows/flow_pp_test.go
```diff
@@ -735,7 +738,8 @@ func generateTestProof(t *testing.T) treetypes.Proof {
 
 func Test_PPFlow_GetCertificateBuildParams(t *testing.T) {
 	t.Parallel()
-
+	// OVerride to have always the same value and not depend on real clock
```
Stefan-Ethernal:
```suggestion
	// override to have always the same value and not depend on real clock
```

aggsender/flows/flow_aggchain_prover_test.go
```diff
@@ -28,7 +28,8 @@ import (
 
 func Test_AggchainProverFlow_GetCertificateBuildParams(t *testing.T) {
 	t.Parallel()
-
+	// ovveride TimeNowFunc for testing
```
Stefan-Ethernal:
```suggestion
	// override TimeNowFunc for testing
```

aggsender/validator/remote_validator.go
```diff
@@ -39,6 +39,10 @@ func (v *RemoteValidator) String() string {
 	return "RemoteValidator"
 }
 
+func (v *RemoteValidator) HealthCheck(ctx context.Context) (*types.HealthCheckResponse, error) {
```
Stefan-Ethernal, line 43: Nitpick, add a comment?

aggsender/validator/validator_service.go
```diff
@@ -2,22 +2,131 @@ package validator
...
+	signer signertypes.Signer) *ValidatorService {
+	return &ValidatorService{
+		log:            logger,
+		validator:      validator,
+		agglayerClient: agglayerClient,
+		signer:         signer,
+	}
+}
+
+// Implementa el método Status
```
Stefan-Ethernal: Although I understand, maybe translate it to English (or remove the comment) :slightly_smiling_face:
```suggestion
// Implementa el método Status
```

Stefan-Ethernal (approved):
Generally LGTM :100:

grpc/server.go
joanestebanr: I agree, I change it!

config/default.go
joanestebanr: This is the same `Method` as more other fields like `AggSender.AggsenderPrivateKey`, `AggOracle.EVMSender.EthTxManager.PrivateKeys`.... This is already documented in [here](https://github.com/agglayer/aggkit/blob/develop/docs/common_config.md#signerconfig). 

 I would  add a comment refereing this doc.

config/default.go
joanestebanr: no idea, the biggest that I see was arround 25Mb, maybe it's too big this value... I'm going to set to 25Mb

aggsender/validator/local_validator.go
joanestebanr: The idea is to keep that to validate the internal validation code, as soon as the code is mature enough we can remove it

config/default.go
Stefan-Ethernal: You can add some extra margin just to be safe, for example 100 MB? WDYT?

agglayer/grpc/conversion_to_agglayer.go
joanestebanr: good catch, we use a long build the code but it's not longer in use!

aggsender/aggsender.go
joanestebanr: I think that this name cames from this [PR](https://github.com/agglayer/aggkit/pull/738) if @goran-ethernal agrees I can renamed it

aggsender/validator/cert_hash.go
joanestebanr: Thanks you, I was looking for that in the code! I have moved to your suggestion. 
Also I have created the function `Uint32ToBigEndianBytes` as an alias of `Uint32ToBytes`

aggsender/validator/validator_service.go
joanestebanr: you're right!!! 🤣🤣

aggsender/validator/config.go
joanestebanr: It disable the debug RPC service, what is missing is the code to do it! f873ff62b7338b8c51db151527cc5b659760a0d6

aggsender/types/interfaces.go
joanestebanr: `CertificateValidator` already exists, it just validate but don't sign the certificate

PR #753, opened 2025-07-16
feat: save settled cert on startup from scratch

goran-ethernal:
## Description

Because of the new `aggsender validator` call, when starting `aggsender` from scratch, especially in situations where `aggsender` `db` was removed for some reason, we need to sync the last settled certificate as well, not just the last pending, because, to call the `aggsender validator`, regular `aggsender` needs to send the `certificate id` of the last settled certificate as well.

Fixes # (issue)

### DIFF
--- a/aggsender/db/aggsender_db_storage.go
+++ b/aggsender/db/aggsender_db_storage.go
@@ -73,6 +73,8 @@ type AggSenderStorage interface {
 	GetNonAcceptedCertificate() (*NonAcceptedCertificate, error)
 	// SaveOrUpdateCertificate saves or updates a certificate in the storage
 	SaveOrUpdateCertificate(ctx context.Context, certificate types.Certificate) error
+	// GetLastSettledCertificate returns the last settled certificate from the storage
+	GetLastSettledCertificate() (*types.CertificateHeader, error)
 }
 
 var _ AggSenderStorage = (*AggSenderSQLStorage)(nil)
@@ -192,6 +194,20 @@ func (a *AggSenderSQLStorage) GetLastSentCertificateHeader() (*types.Certificate
 	if err := meddler.QueryRow(a.db, &certificateHeader,
 		fmt.Sprintf("%s ORDER BY height DESC LIMIT 1;", selectQueryCertificateHeader)); err != nil {
 		return nil, getSelectQueryError(0, err)
+	}
+	return &certificateHeader, nil
+}
+
+// GetLastSettledCertificate returns the last settled certificate from the storage
+func (a *AggSenderSQLStorage) GetLastSettledCertificate() (*types.CertificateHeader, error) {
+	var certificateHeader types.CertificateHeader
+	if err := meddler.QueryRow(a.db, &certificateHeader,
+		fmt.Sprintf("%s WHERE status = $1 ORDER BY height DESC LIMIT 1;", selectQueryCertificateHeader),
+		agglayertypes.Settled); err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, db.ErrNotFound
+		}
+		return nil, fmt.Errorf("error getting last settled certificate: %w", err)
 	}
 	return &certificateHeader, nil
 }
--- a/aggsender/db/aggsender_db_storage.go
+++ b/aggsender/db/aggsender_db_storage.go
@@ -71,6 +71,8 @@ type AggSenderStorage interface {
 	SaveNonAcceptedCertificate(ctx context.Context, nonAcceptedCert *NonAcceptedCertificate) error
 	// GetNonAcceptedCertificate returns the last non-accepted certificate
 	GetNonAcceptedCertificate() (*NonAcceptedCertificate, error)
+	// SaveOrUpdateCertificate saves or updates a certificate in the storage
+	SaveOrUpdateCertificate(ctx context.Context, certificate types.Certificate) error
 }
 
 var _ AggSenderStorage = (*AggSenderSQLStorage)(nil)
@@ -194,6 +196,57 @@ func (a *AggSenderSQLStorage) GetLastSentCertificateHeader() (*types.Certificate
 	return &certificateHeader, nil
 }
 
+// SaveOrUpdateCertificate saves the certificate in the storage
+// It will insert a new certificate or update the existing one if it has the same height and certificate ID
+func (a *AggSenderSQLStorage) SaveOrUpdateCertificate(ctx context.Context, certificate types.Certificate) error {
+	tx, err := newTxer(ctx, a.db)
+	if err != nil {
+		return fmt.Errorf("saveOrUpdateCertificate NewTx. Err: %w", err)
+	}
+	shouldRollback := true
+	defer func() {
+		if shouldRollback {
+			if errRllbck := tx.Rollback(); errRllbck != nil {
+				a.logger.Errorf(errWhileRollbackFormat, errRllbck)
+			}
+		}
+	}()
+
+	certInfo, err := convertCertificateToCertificateInfo(&certificate)
+	if err != nil {
+		return fmt.Errorf("error converting certificate to certificate info: %w", err)
+	}
+
+	var count int
+	err = tx.QueryRow(`SELECT COUNT(*) FROM certificate_info WHERE height = $1;`, certInfo.Height).Scan(&count)
+	if err != nil {
+		return fmt.Errorf("error checking if certificate exists: %w", err)
+	}
+
+	// meddler does not support upsert, so we need to do this manually by checking if the certificate exists
+	if count == 0 {
+		// insert new certificate if it does not exist
+		if err = meddler.Insert(tx, "certificate_info", certInfo); err != nil {
+			return fmt.Errorf("error inserting certificate info: %w", err)
+		}
+	} else {
+		// if the certificate exists, we need to update it
+		if err = updateCertStatus(tx, certInfo.CertificateID, certInfo.Status, certInfo.UpdatedAt); err != nil {
+			return fmt.Errorf("error updating certificate status: %w", err)
+		}
+	}
+
+	if err = tx.Commit(); err != nil {
+		return fmt.Errorf("saveOrUpdateCertificate commit. Err: %w", err)
+	}
+	shouldRollback = false
+
+	a.logger.Debugf("inserted certificate - Height: %d. Hash: %s",
+		certInfo.Height, certInfo.CertificateID)
+
+	return nil
+}
+
 // SaveLastSentCertificate saves the last certificate sent to the aggLayer
 func (a *AggSenderSQLStorage) SaveLastSentCertificate(ctx context.Context, certificate types.Certificate) error {
 	tx, err := db.NewTx(ctx, a.db)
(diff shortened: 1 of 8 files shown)


aggsender/statuschecker/initial_state.go
```diff
@@ -37,9 +37,10 @@ func (i initialStatusAction) String() string {
 }
 
 type initialStatusResult struct {
-	action  initialStatusAction
-	message string
-	cert    *agglayertypes.CertificateHeader
+	action      initialStatusAction
+	message     string
+	cert        *agglayertypes.CertificateHeader
+	settledCert *agglayertypes.CertificateHeader
```
joanestebanr: I suggest instead of adding the field `settledCert` to this struct, just returns a list of actions  in the `process()` call. 
There is a case, that you need to update certificate and insert last settled that is not possible to achieve with current struct

goran-ethernal: Which case do you mention? Is it `Case 3.2` or `Case 4` in the code? Or maybe `Case 5`?

goran-ethernal: I've added a list of actions now, instead of this: https://github.com/agglayer/aggkit/pull/753/commits/b62533e427e97447cec0a2804d407e25ff2a9dbd

aggsender/statuschecker/initial_state.go
```diff
@@ -15,6 +15,8 @@ import (
 var newInitialStatusFn = newInitialStatus
 
 const (
+	initialStatusResultsCapacity = 2
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/753/commits/030455d3200f824957ff66f9649c3e5c5360f882

aggsender/statuschecker/initial_state.go
```diff
@@ -207,3 +272,75 @@ func (i *initialStatus) getLatestAggLayerCert() *agglayertypes.CertificateHeader
...
+		if i.LocalSettledCert != nil {
+			return nil, fmt.Errorf("recovery: local settled certificate exists (%s)"+
+				"but agglayer has no settled certificate", i.LocalSettledCert.ID())
+		}
+
+		// CASE 2: Both local and agglayer have no settled certificate
+		return newInitialStatusResult(
+			InitialStatusActionNone,
+			"agglayer and local storage have no settled certificate",
+			i.SettledCert,
```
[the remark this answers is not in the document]
goran-ethernal: We do not use this, plus it is a pointer, and we only use it for logging, where we do handle the nil certificate, so I wouldn't fix this.

aggsender/statuschecker/initial_state.go
```diff
@@ -10,23 +10,30 @@ import (
...
 
 type initialStatus struct {
-	SettledCert *agglayertypes.CertificateHeader
-	PendingCert *agglayertypes.CertificateHeader
-	LocalCert   *types.CertificateHeader
-	log         common.Logger
+	SettledCert      *agglayertypes.CertificateHeader
+	PendingCert      *agglayertypes.CertificateHeader
+	LocalCert        *types.CertificateHeader
+	LocalSettledCert *types.CertificateHeader
```
joanestebanr: maybe it's a chance to improve this naming: 
- AgglayerLastSettledCert
- AgglayerLastPendingCert
- LocalLastCert
- LocalLastSettledCert

WDYT?

joanestebanr (approved):

aggsender/statuschecker/initial_state.go
goran-ethernal: Done: https://github.com/agglayer/aggkit/pull/753/commits/b124f07d7a0a028328b8a00ee5e68f500230e223

aggsender/db/aggsender_db_storage.go
```diff
@@ -194,6 +198,71 @@ func (a *AggSenderSQLStorage) GetLastSentCertificateHeader() (*types.Certificate
...
+		}
+	}
+
+	if err = tx.Commit(); err != nil {
+		return fmt.Errorf("saveOrUpdateCertificate commit. Err: %w", err)
+	}
+	shouldRollback = false
+
+	a.logger.Debugf("inserted certificate - Height: %d. Hash: %s",
+		certInfo.Height, certInfo.CertificateID)
```
Stefan-Ethernal: Vary the log message, based on the action (if it was insert or update), wdyt?

```go
if count == 0 {
    a.logger.Debugf("inserted certificate - Height: %d. Hash: %s",
        certInfo.Height, certInfo.CertificateID)
} else {
    a.logger.Debugf("updated certificate - Height: %d. Hash: %s",
        certInfo.Height, certInfo.CertificateID)
}
```

aggsender/db/aggsender_db_storage.go
```diff
@@ -194,6 +198,71 @@ func (a *AggSenderSQLStorage) GetLastSentCertificateHeader() (*types.Certificate
...
+			return nil, db.ErrNotFound
+		}
+		return nil, fmt.Errorf("error getting last settled certificate: %w", err)
+	}
+	return &certificateHeader, nil
+}
+
+// SaveOrUpdateCertificate saves the certificate in the storage
+// It will insert a new certificate or update the existing one if it has the same height and certificate ID
+func (a *AggSenderSQLStorage) SaveOrUpdateCertificate(ctx context.Context, certificate types.Certificate) error {
```
Stefan-Ethernal, line 217: WDYT about implementing an UPSERT statement? Providing the example snippet:

Just to say I don't have strong preferences, because I do understand that `CertificateInfo` has around 15 or so fields and it gets tedious enumerating them manually (much easier to rely on the `meddler.Insert`).

```go
func (a *AggSenderSQLStorage) SaveOrUpdateCertificate(ctx context.Context, certificate types.Certificate) error {
	tx, err := newTxer(ctx, a.db)
	if err != nil {
		return fmt.Errorf("saveOrUpdateCertificate NewTx: %w", err)
	}
	defer func() {
		if err != nil {
			if errRollback := tx.Rollback(); errRollback != nil {
				a.logger.Errorf("rollback failed: %v", errRollback)
			}
		}
	}()

	certInfo, err := convertCertificateToCertificateInfo(&certificate)
	if err != nil {
		return fmt.Errorf("error converting certificate: %w", err)
	}

    // THE IMPORTANT PART (there might be some fields missing here)
	_, err = tx.ExecContext(ctx, `
		INSERT INTO certificate_info (height, certificate_id, status, updated_at)
		VALUES ($1, $2, $3, $4)
		ON CONFLICT (height) DO UPDATE
		SET status = EXCLUDED.status,
		    updated_at = EXCLUDED.updated_at;
	`, certInfo.Height, certInfo.CertificateID, certInfo.Status, certInfo.UpdatedAt)
	if err != nil {
		return fmt.Errorf("upsert certificate_info: %w", err)
	}
    // THE END OF IMPORTANT PART

	if err = tx.Commit(); err != nil {
		return fmt.Errorf("commit failed: %w", err)
	}

	a.logger.Debugf("upserted certificate - Height: %d. Hash: %s", certInfo.Height, certInfo.CertificateID)
	return nil
}
```

aggsender/statuschecker/initial_state_test.go
```diff
@@ -190,57 +206,115 @@ func TestRegularCases(t *testing.T) {
...
+		{
+			name:             "14| LocalCert: ID3, h1, pending	| LocalSettled: ID2, h1	| AgglayerSettled: ID1, h2   | AgglayerPending: ID3, h1 | agglayer doesn't have settled cert",
+			localCert:        &certTestData{hash1, 3, agglayertypes.Pending},
+			localSettledCert: &certTestData{hash1, 2, agglayertypes.Settled},
+			agglayerSettled:  &certTestData{hash2, 1, agglayertypes.Settled},
+			agglayerPending:  &certTestData{hash1, 3, agglayertypes.Pending},
+			resultError:      true,
+		},
+		{
+			name:             "14| LocalCert: ID3, h1, pending | LocalSettled: ID2, h1 | AgglayerSettled: ID2, h1 | AgglayerPending: ID3, h1 | store(PENDING) & none(SETTLED)",
```
Stefan-Ethernal: Should this be test case #15 (and increment each subsequent by one), or is deliberately this way?

```suggestion
			name:             "15| LocalCert: ID3, h1, pending | LocalSettled: ID2, h1 | AgglayerSettled: ID2, h1 | AgglayerPending: ID3, h1 | store(PENDING) & none(SETTLED)",
```

aggsender/statuschecker/initial_state.go
```diff
@@ -163,47 +229,123 @@ func (i *initialStatus) process() (*initialStatusResult, error) {
...
-	if i.PendingCert == nil {
-		return i.SettledCert
+	if i.AgglayerLastPendingCert == nil {
+		return i.AgglayerLastSettledCert
+	}
+	return i.AgglayerLastPendingCert
+}
+
+// processSettledCert checks the last settled certificate from agglayer vs local storage
+func (i *initialStatus) processSettledCert() (*initialStatusResult, error) {
```
Stefan-Ethernal: WDYT?
```suggestion
func (i *initialStatus) processLastSettledCert() (*initialStatusResult, error) {
```

aggsender/statuschecker/initial_state.go
```diff
@@ -163,47 +229,123 @@ func (i *initialStatus) process() (*initialStatusResult, error) {
...
+
+		// CASE 5.2: the local settled certificate matches the agglayer settled certificate
+		return newInitialStatusResult(
+			InitialStatusActionNone,
+			"last settled certificate already in local storage with same height and ID",
+			i.AgglayerLastSettledCert,
+		), nil
+	}
+
+	// CASE 5: We have a settled certificate in local storage that is lower than the one in the agglayer
```
Stefan-Ethernal: WDYT?
```suggestion
	// CASE 6: We have a settled certificate in local storage that is lower than the one in the agglayer
```

Stefan-Ethernal (approved):
:100: LGTM!

aggsender/db/aggsender_db_storage.go
goran-ethernal: Nope, we can not go with UPSERT here, if we are going to insert the certificate, we need to insert all of its data. Current meddler we are using, is not really supporting `UPSERT`, and since certificates have complex data like nullable hashes, or aggchain proofs, etc., which require custom meddlers to be inserted in `db`, we can not write an easy query for `UPSERT` without adding code that already does the same as our custom meddlers.

Since we open one txn in this case, its easier to just ask if the certificate already exists, if it does, only update its status and when it was updated (no other fields should be updated here, since only `Status` and `UpdatedAt` are mutable fields, others should always be immutable). If it does not exist, we insert it using meddler.

aggsender/statuschecker/initial_state.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/753/commits/7ab140b26d79ace0bfe83b72ead1c4ed52dafcc8

aggsender/statuschecker/initial_state.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/753/commits/7ab140b26d79ace0bfe83b72ead1c4ed52dafcc8

aggsender/statuschecker/initial_state_test.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/753/commits/7ab140b26d79ace0bfe83b72ead1c4ed52dafcc8

aggsender/db/aggsender_db_storage.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/753/commits/7ab140b26d79ace0bfe83b72ead1c4ed52dafcc8

PR #760, opened 2025-07-18
feat: retry build/submit certificate in case of error

### LINKED ISSUE #727: aggsender: retry to send certificate if the error is a temporal one
# Feature Request

When aggsender reach the percentatge of the epoch try to send a certificate, if fails wait to next epoch. 
It make sense that if fails for a temporal reason (e.g. there are  no bridges for a PP certificate that require that) retries before waiting next epoch

joanestebanr:
## 🔄 Changes Summary
- Aggsender when reach the epoch percentatge try to submit a certificate to `agglayer` if something goes wrong must retry 
- Internaly have been added a way of define retry policies: 
  - You can define  a backoff function or a list os defailts
  - Set a field in config file of type `common.RetryPolicyGenericConfig` 
  - Call to this field `.NewRetryHandler()` to get the specific retryHandler class
  - Call to `aggkitcommon.Execute(retryHandler,...,`
- Fix flaky unit tests: 
   -  Agglayer_client_cache: removed go routines
   - Flow_pp and flow_aggchain_prover test: remove global variable to get `timeNow()`

## ⚠️ Breaking Changes
- 🛠️ **Config**: NA
- 🔌 **API/CLI**: NA
- 🗑️ **Deprecated Features**: NA

## 📋 Config Updates
- 🧾 **Diff/Config snippet**: 
- Add field `AggSender.RetriesToBuildAndSendCertificate`
```
[AggSender]
[AggSender.RetriesToBuildAndSendCertificate]
                Mode= "delays"
		Delays = [ "1m", "1m", "2m", "5m", "5m", "8m" ]
		MaxRetries = 6 # 1+6 attemps, around 22m 
```

## ✅ Testing
- 🤖 **Automatic**: E2E

## 🐞 Issues
- Closes #727


common/logger_silent.go
```diff
@@ -0,0 +1,23 @@
+package common
+
+// Logger is an interface that defines the methods to log messages.
+// but don't emit any log
+// If you want to use a nil logger, instead of checking all the time you
+// can use a object of this type
+type SilentLogger struct{}
+
+// NewSlientLogger creates a new SilentLogger instance.
+func NewSlientLogger() Logger {
```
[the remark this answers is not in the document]
joanestebanr: removed function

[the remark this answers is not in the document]
joanestebanr: removed function

aggsender/aggsender.go
```diff
@@ -271,6 +270,22 @@ func (a *AggSender) sendCertificates(ctx context.Context, returnAfterNIterations
...
+		"sendCertificateWithRetries",
+		func() (*agglayertypes.Certificate, error) {
+			cert, err := a.sendCertificate(ctx)
+			a.status.SetLastError(err)
+			if err != nil {
+				a.log.Error(err)
+			}
+			a.checkSendCertificateStopCondition(err)
+			return cert, err
+		})
```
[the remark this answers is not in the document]
joanestebanr: changed 241d57403ce7b8fc00fb2556425663791d8130ff

aggsender/config/config.go
```diff
@@ -89,6 +89,8 @@ type Config struct {
 	RequireValidatorCall bool `mapstructure:"RequireValidatorCall"`
 	// ValidatorClient is the configuration for the ValidatorClient
 	ValidatorClient *grpc.ClientConfig `mapstructure:"ValidatorClient"`
+	// RetriesToBuildAndSendCertificate is the configuration for the retries to build and send a certificate
+	RetriesToBuildAndSendCertificate common.RetryDelays `mapstructure:"RetriesToBuildAndSendCertificate"`
```
goran-ethernal: Maybe it makes sense to move the `RetryCertAfterInError` field to retry config, something like this:

```go
type CertificateRetry struct {
     common.RetryDelays `mapstructure:",squash"`
     // RetryCertAfterInError when a cert pass to 'InError'
     // state the AggSender will try to resend it immediately
     RetryCertAfterInError bool `mapstructure:"RetryCertAfterInError"`
}
```

```toml
[AggSender.CertificateRetry]
Delays = [ "30s", "1m", "2m", "5m", "5m", "10m" ]
MaxAttempts = 6 # around 23m
RetryCertAfterInError = true
```

I've just saw this in @Stefan-Ethernal PR, it looks like it supports embedded config params from a different struct.
Now we have multiple fields in regards to certificate retry, I think it would be good to move them all together in one struct.

WDYT?

common/retry_delays.go
```diff
@@ -0,0 +1,138 @@
...
+
+// RetryDelays is a struct that holds the retry delays and the maximum number of retries.
+// It implements the RetryDelayer interface, which allows executing a function with retry logic.
+// The delays are specified as a slice of types.Duration, and the maximum number of retries can be set.
+// If maxRetries is set to 0, it means infinite retries are allowed.
+//
+// You can also abort the retrying wrapping the error ErrAbort into the result
+// this is useful if there are some conditions that should not be retried
+type RetryDelays struct {
+	Delays []types.Duration
```
goran-ethernal: Just to be sure these are used in config:

```suggestion
	Delays []types.Duration `mapstructure:"Delays"`
	MaxAttempts int `mapstructure:"MaxAttempts"`
```

common/retry_delays.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	logFunc("fails to execute %s after %d retries",
+		name, retries)
+	return zero, fmt.Errorf("fails to execute %s after %d retries. %w",
+		name, retries, ErrExecutionFails)
+}
+
+// Execute executes the provided function with retry logic for a non return function
+// check func Execute[T any] for more details
+func (r *RetryDelays) Execute(ctx context.Context,
```
goran-ethernal: You are not using this function anywhere.

common/retry_delays.go
```diff
@@ -0,0 +1,138 @@
...
+	logFunc func(format string, args ...interface{}),
+	name string,
+	fn func() (T, error)) (T, error) {
+	var zero T
+	if err := retryDelaysConfig.Validate(); err != nil {
+		return zero, err
+	}
+	if logFunc == nil {
+		// if logger is nil, we create a silent logger
+		logFunc = func(format string, args ...interface{}) {}
```
goran-ethernal: Maybe have a global variable called `emptyLogFn`:

```go
var emptyLogFn = func(format string, args ...interface{}){}
```

and then:

```go
if logFunc == nil {
    logFunc = emptyLogFn
}
```

common/retry_delays.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	retries := 0
+	for {
```
goran-ethernal: I think you can move the first if to the for loop itself:

```suggestion
	for retryAttempt := 0; retryDelaysConfig.MaxAttempts == 0 || retryAttempt < retryDelaysConfig.MaxAttempts; retryAttempt++
```

common/retry_delays.go
```diff
@@ -0,0 +1,138 @@
...
+		if retryDelaysConfig.MaxAttempts > 0 && retries >= retryDelaysConfig.MaxAttempts {
+			break
+		}
+		var delay types.Duration
+		if retries < len(retryDelaysConfig.Delays) {
```
goran-ethernal: You can simplify this:

```go
delay := retryDelaysConfig.Delays[min(attempt, len(retryDelaysConfig.Delays)-1)]
```

common/retry_delays.go
```diff
@@ -0,0 +1,138 @@
...
+			delay = retryDelaysConfig.Delays[retries]
+		} else {
+			delay = retryDelaysConfig.Delays[len(retryDelaysConfig.Delays)-1]
+		}
+		logFunc("executing %s try %d/%d (next delay: %s)",
+			name, retries+1, retryDelaysConfig.MaxAttempts, delay.String())
+		result, err := fn()
+		if err != nil && errors.Is(err, ErrAbort) {
```
goran-ethernal: Hmm, I would maybe try to improve readability here:

```go
result, err := fn()
if err == nil {
	logFunc("successful run %s in try %d", name, attempt+1)
	return result, nil
}

if errors.Is(err, ErrAbort) {
	logFunc("aborting execution of %s due to error: %v", name, err)
	return zero, err
}

select {
case <-ctx.Done():
	logFunc("executing %s try %d/%d was canceled", name, retryAttempt+1, retryDelaysConfig.MaxAttempts)
	return zero, ctx.Err()
case <-time.After(delay.Duration):
	// continue
}

```

aggsender/config/config.go
joanestebanr: But are not related these fields... I don't see why to put together. Maybe we can create a new section add config param related to retries and refactor the names to be more explicit: 
```
[Retry]
RetryCertAfterInError = true
DelayBetweenRetries = "30s"
MaxRetriesStoreCertificate = 3
[AggSender.CertificateRetry]
Delays = [ "30s", "1m", "2m", "5m", "5m", "10m" ]
MaxAttempts = 6 # around 23m
```

BTW: Which @Stefan-Ethernal PR support embedded config params?

common/retry_delays.go
joanestebanr: deleted 8cb298851398ee0ad0e248e9d8aadf6b037be2ff

common/retry_delays.go
joanestebanr: It's only used in 1 place, and externally to set it you have to just pass a `nil`, I don't see any improvement there

aggsender/config/config.go
Stefan-Ethernal: For the reference:
- how to define the field, the `squash` annotation (https://github.com/agglayer/aggkit/pull/761/files#diff-32f4f366dad3cee33f9165dbf93cf213721a7dd734bf3e99cd295e35ba122e4dR116)
- how it is used in the config (https://github.com/agglayer/aggkit/pull/761/files#diff-61cfb7e001b72e449356ab5369f659461155cdd3c1b162d0ada0529a836d8374R44-R49)

common/retry_delays.go
joanestebanr: Yes, good suggestion! 340f317f474bc188d3969c7366f6d0d69851cbfa

common/retry_delays.go
joanestebanr: good simplification!

common/retry_delays_test.go
```diff
@@ -0,0 +1,125 @@
...
+		MaxAttempts: 1,
+	}
+	ctx := context.Background()
+	fn := func() (string, error) {
+		return "ok", nil
+	}
+	result, err := Execute(r, ctx, nil, "test-nil-logger", fn)
+	require.NoError(t, err)
+	require.Equal(t, "ok", result)
+}
```
Stefan-Ethernal: Nitpick, empty line missing:

```suggestion
}

```

aggsender/config/config.go
goran-ethernal: @joanestebanr I was only talking about `RetryCertAfterInError` since only this field has anything to do with resending certificate after some error happens, so it looked like it would make sense to group that with fields you added, since they all are used for retry after an error.

common/retry_delays.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	for _, delay := range r.Delays {
+		if delay.Duration <= 0 {
+			return fmt.Errorf("retry delay must be greater than zero, got %s. %w",
+				delay.Duration, ErrInvalidConfig)
+		}
+	}
+	if r.MaxAttempts < 0 {
+		return fmt.Errorf("max retries cannot be negative, got %d. %w",
+			r.MaxAttempts, ErrInvalidConfig)
```
Stefan-Ethernal: Does it make sense to first report the generic error and then fill in the details?

```suggestion
		return fmt.Errorf("%w: retry delays cannot be nil.", ErrInvalidConfig)
	}
	if len(r.Delays) == 0 {
		return fmt.Errorf("%w: retry delays cannot be empty.", ErrInvalidConfig)
	}
	for _, delay := range r.Delays {
		if delay.Duration <= 0 {
			return fmt.Errorf("%w: retry delay must be greater than zero, got %s.",
				delay.Duration)
		}
	}
	if r.MaxAttempts < 0 {
		return fmt.Errorf("%w: max retries cannot be negative, got %d",
			ErrInvalidConfig, r.MaxAttempts)
```

common/retry_delays.go
goran-ethernal: I was just thinking in the terms of not instantiating anything if in the case of a nil `logFunc`, just use some already made global variable.

Define a global variable in the file and just use it in this case, and not instantiating a new object each time.

But its just a small thing, I don't have a real preference.

common/retry_delays.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	logFunc("fails to execute %s after %d retries",
+		name, attempt)
+	return zero, fmt.Errorf("fails to execute %s after %d retries. %w",
+		name, attempt, ErrExecutionFails)
+}
+
+func executionCheckResult() {
+
+}
```
Stefan-Ethernal: Remove, unused?

```suggestion
```

common/retry_delays.go
```diff
@@ -0,0 +1,134 @@
...
+// the execution will be aborted and no more retries will be attempted.
+func Execute[T any](retryDelaysConfig *RetryDelays,
+	ctx context.Context,
+	logFunc func(format string, args ...interface{}),
+	name string,
+	payloadFunc func() (T, error)) (T, error) {
+	var zero T
+	if err := retryDelaysConfig.Validate(); err != nil {
+		return zero, err
+	}
```
Stefan-Ethernal: Would it make sense to validate the retry configuration on startup (in `run.go`) and catch the invalid situations as soon as possible?

common/retry_delays.go
```diff
@@ -0,0 +1,134 @@
...
+	if logFunc == nil {
+		// if logger is nil, we create a silent logger
+		logFunc = func(format string, args ...interface{}) {}
+	}
+
+	attempt := 0
+	for attempt := 0; retryDelaysConfig.InfiniteRetries() || attempt < retryDelaysConfig.MaxAttempts; attempt++ {
+		delay := retryDelaysConfig.Delay(attempt)
+		logFunc("executing %s try %d/%d (next delay: %s)",
+			name, attempt+1, retryDelaysConfig.MaxAttempts, delay.String())
```
Stefan-Ethernal: Should we tweak the log message in case we have infinite retries? In particular, it probably does not make sense to output `%d/%d` for the attempts, but maybe just output the current attempt. WDYT?

common/retry_delays.go
goran-ethernal: We already do that: https://github.com/agglayer/aggkit/blob/3f326c55e3e6bb878cf1ad3eb6a529facd8181d6/cmd/run.go#L272.
We just need to add the call to validate this config as well in that `aggsendercfg.Validate` function.

common/retry_delays.go
joanestebanr: No strong opinion about that, I apply it!

common/retry_delays.go
joanestebanr: good catch!

common/retry_delays.go
joanestebanr: thanks you!, good catch!

common/retry_delays.go
joanestebanr: I agree, I have moved the code to `RetryDelays.StringAttemp`

aggsender/config/config.go
joanestebanr: But are not related, it's a bit missleading because if I see  grouping that lead to think that if `RetryCertAfterInError` is false is not going to be retries at all

common/retry_delays.go
```diff
@@ -0,0 +1,176 @@
...
+	var zero T
+	if err := retryDelaysConfig.Validate(); err != nil {
+		return zero, err
+	}
+	if logFunc == nil {
+		// if logger is nil, we create a silent logger
+		logFunc = silentLog
+	}
+	var err error
+	var attempt int
```
Stefan-Ethernal: Nitpick:

```suggestion
	var (
	    err error
            attempt int
        )
```

joanestebanr:
PR Converted  because we are going to unify the retry config with PR https://github.com/agglayer/aggkit/pull/761/files#

common/retry_config.go
```diff
@@ -0,0 +1,115 @@
...
+// it's a merge of struct RetryBackoffConfig and RetryDelaysConfig in order of
+// simplify reading from config file (check types)
+type RetryPolicyGenericConfig struct {
+	Mode              RetryConfigMode // "delays", "backoff", or "" for no retries
+	MaxRetries        int
+	Delays            []types.Duration
+	InitialBackoff    types.Duration
+	MaxBackoff        types.Duration
+	BackoffMultiplier float64
+	// internal field to cache the underlaying config
```
Stefan-Ethernal:
```suggestion
	// internal field to cache the underlying config
```

common/retry_config.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	return fmt.Sprintf("RetryPolicyConfig{Mode: %s, Config: %s}", r.Mode, cfg.String())
+}
+
+// CleanCache clears the internal cache of the RetryPolicyGenericConfig.
+// useful is you modify any params and want to rebuild the underlaying object
+func (r *RetryPolicyGenericConfig) CleanCache() {
+	// Clean the cache to force a new Factory call next time
+	r.cache = nil
+}
```
Stefan-Ethernal: Unused (other than in the unit test), remove it, wdyt?

```suggestion
```

common/retry_config.go
```diff
@@ -0,0 +1,115 @@
...
+	return cfg.Validate()
+}
+
+func (r *RetryPolicyGenericConfig) NewRetryHandler() *RetryHandler {
+	cfg, err := r.Factory()
+	if err != nil {
+		return NewRetryHandler(nil, 0) // return a no-op handler
+	}
+	return cfg.NewRetryHandler()
+}
```
Stefan-Ethernal, line 51: Nitpick
```suggestion
}

```

common/retry_config.go
```diff
@@ -0,0 +1,115 @@
...
+func (r *RetryPolicyGenericConfig) Brief() string {
+	cfg, err := r.Factory()
+	if err != nil {
+		return fmt.Sprintf("RetryPolicyConfig{Error: %s}", err)
+	}
+	return fmt.Sprintf("%s/%s", r.Mode, cfg.Brief())
+}
```
Stefan-Ethernal, line 59:
```suggestion
}

```

common/retry_config.go
```diff
@@ -0,0 +1,115 @@
...
+		return err
+	}
+	return cfg.Validate()
+}
+
+func (r *RetryPolicyGenericConfig) NewRetryHandler() *RetryHandler {
+	cfg, err := r.Factory()
+	if err != nil {
+		return NewRetryHandler(nil, 0) // return a no-op handler
+	}
```
Stefan-Ethernal: Let's not silence this error, we should at least log it (if not interrupt the booting of the aggkit altogether). WDYT?

common/retry_backoff_config.go
```diff
@@ -0,0 +1,66 @@
...
+	"math"
+	"time"
+
+	commontypes "github.com/agglayer/aggkit/common/types"
+	"github.com/agglayer/aggkit/config/types"
+)
+
+var _ commontypes.RetryPolicyConfigurer = (*RetryBackoffConfig)(nil)
+
+type RetryBackoffConfig struct {
```
Stefan-Ethernal, line 14: I'd suggest the rename, in order not to confuse it with the struct that is used to unmarshal the TOML config. WDYT?

```suggestion
type RetryBackoffPolicy struct {
```

common/retry_delays_config.go
```diff
@@ -0,0 +1,73 @@
...
+
+// RetryDelaysConfig defines the configuration for retry delays.
+type RetryDelaysConfig struct {
+	// MaxRetries is the maximum number of retries to attempt.
+	// if MaxRetries is -1, it means infinite retries.
+	// if MaxRetries is 0, it means no retries will be attempted.
+	MaxRetries int `mapstructure:"MaxRetries"`
+	// Delays is a list of durations to wait before each retry.
+	// if there are more attempts that item o the list is used the last one
+	Delays []types.Duration `mapstructure:"Delays"`
```
Stefan-Ethernal: Since it is not unmarshalled from the TOML, I beleive the annotations can be removed?
```suggestion
	MaxRetries int
	// Delays is a list of durations to wait before each retry.
	// if there are more attempts that item o the list is used the last one
	Delays []types.Duration
```

common/retry_delays_config.go
```diff
@@ -0,0 +1,73 @@
...
+	"fmt"
+
+	commontypes "github.com/agglayer/aggkit/common/types"
+	"github.com/agglayer/aggkit/config/types"
+)
+
+var _ commontypes.RetryPolicyConfigurer = (*RetryDelaysConfig)(nil)
+
+// RetryDelaysConfig defines the configuration for retry delays.
+type RetryDelaysConfig struct {
```
Stefan-Ethernal, line 13: WDYT (in order not to be considered that it is unmarshalled from the TOML config)
```suggestion
type RetryDelaysPolicy struct {
```

common/retry_config.go
```diff
@@ -0,0 +1,103 @@
...
+func (r *RetryPolicyGenericConfig) SetCache(instance commontypes.RetryPolicyConfigurer) {
+	r.cache = instance
+}
+
+func (r *RetryPolicyGenericConfig) Factory() (commontypes.RetryPolicyConfigurer, error) {
+	if r.cache != nil {
+		return r.cache, nil
+	}
+	var cache commontypes.RetryPolicyConfigurer
+	var err error
```
Stefan-Ethernal, line 88: Nitpick
```suggestion
	var (
        cache commontypes.RetryPolicyConfigurer
	    err error
    )
```

common/retry_config.go
```diff
@@ -0,0 +1,103 @@
...
+// simplify reading from config file (check types)
+type RetryPolicyGenericConfig struct {
+	Mode              RetryConfigMode // "delays", "backoff", or "" for no retries
+	MaxRetries        int
+	Delays            []types.Duration
+	InitialBackoff    types.Duration
+	MaxBackoff        types.Duration
+	BackoffMultiplier float64
+	// internal field to cache the underlaying config
+	cache commontypes.RetryPolicyConfigurer `mapstructure:"-"`
```
Stefan-Ethernal, line 34: Do we have a better place to keep this field, as this seems a bit odd to me (we are abusing the struct that should be used only to model the TOML configuration for retries).

Unsure about this, but just thinking out loud... WDYT about having another struct, that would keep the retryPolicy and know everything about creating the certain `RetryPolicy`?

```go
type RetryPolicyManager struct {
    retryPolicy commontypes.RetryPolicyConfigurer
}

func (rpm *RetryPolicyManager) Factory(cfg *RetryPolicyGenericConfig) commontypes.RetryPolicyConfigurer, error {
    if r.retryPolicy != nil {
		return r.retryPolicy, nil
	}
	var (
        retryPolicy commontypes.RetryPolicyConfigurer
	    err error
    )
	switch cfg.Mode {
	case RetryConfigModeDelays:
		retryPolicy, err = NewRetryDelaysConfig(cfg)
	case RetryConfigModeBackoff:
		retryPolicy, err = NewRetryBackoffConfig(cfg)
	case RetryConfigModeNoRetries:
		retryPolicy = &RetryDelaysConfig{MaxRetries: 0}

	default:
		return nil, fmt.Errorf("%w: unknown mode %s", ErrInvalidRetryConfigMode, r.Mode)
	}
	if err != nil {
		return nil, fmt.Errorf("failed to create retry policy config: %w", err)
	}
	r.SetCache(retryPolicy)
	return cache, nil
}
```

common/retry_delays.go
joanestebanr: It increase the number of lines in 2, do you think it's worth it?

common/retry_config.go
joanestebanr: That function is required by unittest, can't be removed

common/retry_backoff_config.go
joanestebanr: In fact is a config, not a Policy. The logic is in another object... maybe we can came up with a better name...

common/retry_config.go
joanestebanr: Same case, it increase the number of lines in 2. WDYT?

common/retry_delays.go
Stefan-Ethernal: I'm not insisting to change it, but generally speaking if the variables are logically related/used in the same logical scope, the grouping of vars makes sense.

common/retry_config.go
joanestebanr: I'm on the fence, I agree that current implementation abuse of the  config struct, but also think that adding another object is overcomplicating things...maybe we can huddle and discuss this point

Stefan-Ethernal: My feeling is that we need better separation of concerns here. This is a dirty approach and we should come up with something better, although I don't have a clear idea...

Stefan-Ethernal (approved):

common/retry_config.go
joanestebanr: It's important to notice that to be able to do `RetryPolicyGenericConfig.Validate()` we need the underlying object... so splitting this object is not so obvious, for that reason, after a huddle, we agree to left in that way until we came up with a better idea.

common/retry_config.go
joanestebanr: we agree that right now it's better left as is

common/retry_delays_config.go
joanestebanr: We discussed it, and both are in the fence, that have a part of config and a part of logic, so there are no clear answer so we agree to left as `...Config`

common/retry_backoff_config.go
joanestebanr: We discuss it, and both are in the fence, that have a part of config and a part of logic, so there are no clear answer so we agree to left as `...Config`

aggsender/flows/flow_base.go
```diff
@@ -103,8 +89,12 @@ func NewBaseFlow(
 		l1InfoTreeDataQuerier: l1InfoTreeDataQuerier,
 		lerQuerier:            lerQuerier,
 		cfg:                   cfg,
+		timeNowFunc:           TimeNowUTC,
 	}
 }
+func (f *baseFlow) SetTimeNowFunc(timeNow func() uint32) {
```
goran-ethernal: Since this is only used in the `flows` package, can we make this private or remove it, and just set the `baseFlow.timeNowFunc = testTimeNowFn` directly. I do not think it is good to expose this only for tests.

common/retry_backoff_config.go
```diff
@@ -0,0 +1,77 @@
...
+		InitialBackoff:    cfg.InitialBackoff,
+		MaxBackoff:        cfg.MaxBackoff,
+		BackoffMultiplier: cfg.BackoffMultiplier,
+		MaxRetries:        cfg.MaxRetries,
+	}
+	return res, res.Validate()
+}
+
+func (r *RetryBackoffConfig) NewRetryHandler() (commontypes.RetryHandler, error) {
+	// TODO: check that implementation
```
goran-ethernal, line 36: Should we remove this TODO?

common/retry_handler_delays.go
```diff
@@ -0,0 +1,166 @@
...
+	return fmt.Sprintf("RetryHandlerDelays{%s}", r.RetryDelaysConfig.String())
+}
+
+// InfiniteRetries return true if the configuration allows infinite retries.
+func (r *RetryHandlerDelays) InfiniteRetries() bool {
+	// Infinite retries are allowed if MaxAttempts is 0.
+	return r != nil && r.MaxRetries == MaxAttemptsInfinite
+}
+
+func (r *RetryHandlerDelays) NoRetries() bool {
```
goran-ethernal: This explains a bit more precise what this returns. `NoRetries` to me seems like you want it to have no retries, and not asking it if it is configured to have 0 retries.

```suggestion
func (r *RetryHandlerDelays) IsNoRetriesConfigured() bool {
```

goran-ethernal (approved):

common/retry_backoff_config.go
joanestebanr: Not yet, that is for @Stefan-Ethernal that must check that the conversion is the expected

common/retry_handler_delays.go
joanestebanr: Also renamed  `InfiniteRetries()` to `IsInfiniteRetriesConfigured()`

PR #767, opened 2025-07-22
chore: remove reorg from l1infotreesync and bridgsesyncl1

### LINKED ISSUE #737: Consider removing L1 reorg detector in certain syncers
Through #720 we have changed to rely on `finalized` blocks for L1 networks (namely L1 info tree syncer and bridge L1 syncer). We probably can consider if we would like to remove the reorg detector for the aforementioned syncers that are relying on hard commitment for no reorgs.

temaniarpit27:
## 🔄 Changes Summary
- removes reorg detector from l1infotreesync and bridgsesyncl1

## 🐞 Issues
- Closes #737


bridgesync/bridgesync.go
```diff
@@ -253,6 +257,11 @@ func newBridgeSync(
 		return nil, err
 	}
 
+	rdString := ""
```
temaniarpit27, line 260: check if a better way

PR #773, opened 2025-07-23
feat: validator: new hash for validator

joanestebanr:
## 🔄 Changes Summary
- It update the way of calculating hash to be signed by `validator` 
```
        let certificate_id = self.hash();

        let commitment = match self.l1_info_tree_leaf_count {
            Some(leaf_count) => keccak256_combine([
                certificate_id.as_digest().as_slice(),
                leaf_count.to_le_bytes().as_slice(),
            ]),
            None => *certificate_id,
        };
```
- Also introduce a `Validate()` function for `agglayer.Certificate`
- 
## ⚠️ Breaking Changes
NA

## 📋 Config Updates
NA

### DIFF
--- a/agglayer/types/types.go
+++ b/agglayer/types/types.go
@@ -259,6 +259,24 @@ type Certificate struct {
 	L1InfoTreeLeafCount uint32                `json:"l1_info_tree_leaf_count,omitempty"`
 }
 
+// Validate certificate
+func (c *Certificate) Validate() error {
+	if c == nil {
+		return errors.New("certificate is nil")
+	}
+	for i, bridgeExit := range c.BridgeExits {
+		if err := bridgeExit.Validate(); err != nil {
+			return fmt.Errorf("bridge exit %d is invalid: %w", i, err)
+		}
+	}
+	for i, importedBridgeExit := range c.ImportedBridgeExits {
+		if err := importedBridgeExit.Validate(); err != nil {
+			return fmt.Errorf("imported bridge exit %d is invalid: %w", i, err)
+		}
+	}
+	return nil
+}
+
 // UnmarshalJSON is the implementation of the json.Unmarshaler interface
 func (c *Certificate) UnmarshalJSON(data []byte) error {
 	aux := &struct {
@@ -311,8 +329,9 @@ func (c *Certificate) Brief() string {
 	return res
 }
 
-// Hash returns a hash that uniquely identifies the certificate
-func (c *Certificate) Hash() common.Hash {
+// CertificateID returns a certificateID that identifies the certificate
+// next fields are not included: CustomChainData, AggchainData, L1InfoTreeLeafCount
+func (c *Certificate) CertificateID() common.Hash {
 	bridgeExitsHashes := make([][]byte, len(c.BridgeExits))
 	for i, bridgeExit := range c.BridgeExits {
 		bridgeExitsHashes[i] = bridgeExit.Hash().Bytes()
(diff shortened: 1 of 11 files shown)


goran-ethernal (approved):

agglayer/types/types.go
```diff
@@ -259,6 +259,24 @@ type Certificate struct {
 	L1InfoTreeLeafCount uint32                `json:"l1_info_tree_leaf_count,omitempty"`
 }
 
+// Validate certificate
+func (c *Certificate) Validate() error {
+	if c == nil {
+		return errors.New("certificate is nil")
+	}
+	for i, bridgeExit := range c.BridgeExits {
```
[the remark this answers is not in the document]
joanestebanr: Validate() function take care of the nil case

agglayer/types/types.go
```diff
@@ -490,6 +524,19 @@ type BridgeExit struct {
 	Metadata           []byte         `json:"metadata"`
 }
 
+func (b *BridgeExit) Validate() error {
```
Stefan-Ethernal, line 527: Maybe a dummy question, but...
Would it make sense to validate that the amount is not less than zero?

agglayer/types/types.go
```diff
@@ -680,6 +730,16 @@ func (l *L1InfoTreeLeaf) String() string {
...
 
+func (l *L1InfoTreeLeaf) Validate() error {
+	if l == nil {
+		return errors.New("L1InfoTreeLeaf is nil")
+	}
+	if l.Inner == nil {
+		return errors.New("L1InfoTreeLeaf inner is nil")
+	}
+	return nil
+}
```
Stefan-Ethernal, line 741: Is this an oversight, or this function is invoked only in the tests? If we don't need it, I'm for removing it. WDYT?

agglayer/types/types.go
```diff
@@ -836,6 +916,19 @@ func (c *ClaimFromRollup) Hash() common.Hash {
...
+func (c *ClaimFromRollup) Validate() error {
+	if c == nil {
+		return errors.New("ClaimFromRollup is nil")
+	}
+	if c.ProofLeafLER == nil || c.ProofLERToRER == nil || c.ProofGERToL1Root == nil {
+		return errors.New("ClaimFromRollup has nil proofs")
+	}
+	if c.L1Leaf == nil {
+		return errors.New("ClaimFromRollup has nil L1Leaf")
+	}
```
Stefan-Ethernal, line 928: Should we invoke `L1Leaf.Validate()` here (if it is not nil)?

aggsender/validator/validator_service_test.go
```diff
@@ -119,6 +119,27 @@ func TestValidatorService_ValidateCertificate(t *testing.T) {
 		_, err := testData.sut.ValidateCertificate(t.Context(), req)
 		require.ErrorContains(t, err, "Invalid certificate conversion")
 	})
+	t.Run("converted certificate with bridgeExits", func(t *testing.T) {
+		testData := newValidatorServiceTestData(t)
+		cert := testCertificate1
```
Stefan-Ethernal, line 124: There is a warning for this, because the certificate instance contains the Mutex

```go
assignment copies lock value to cert: buf.build/gen/go/agglayer/agglayer/protocolbuffers/go/agglayer/node/types/v1.Certificate contains google.golang.org/protobuf/runtime/protoimpl.MessageState contains sync.Mutex
```

Can we create a new instance/deep copy out of the `testCertificate1`?

Stefan-Ethernal (approved):

agglayer/types/types.go
joanestebanr: No strong opinion about that... a thought is that from our side that doesn't invalidate a certificate because we are not doing anything with the value (just hashing). Maybe agglayer need to check that before updating a tree

agglayer/types/types.go
joanestebanr: You're right, I'm going to raise another PR with that! thanks you!

joanestebanr: https://github.com/agglayer/aggkit/pull/777

aggsender/validator/validator_service_test.go
joanestebanr: https://github.com/agglayer/aggkit/pull/777

agglayer/types/types.go
joanestebanr: I implement the check in new PR #777 , just in case

PR #778, opened 2025-07-24
fix: contention in evm driver due to retry handler

Stefan-Ethernal:
## 🔄 Changes Summary
- Spawn a different goroutine for reorgs processing, as it needs to happen independently from the main block processing loop. If the database ends up with inconsistent or invalid data (e.g., due to a reorg and the sequence of events on channels for reorg and new downloaded block being handled), the retry mechanism can get stuck — it keeps retrying the same failing operation (of inserting invalid data) and prevents the reorg from being handled, because both `handleNewBlock` and `handleReorg` were executed from the same goroutine. By moving reorg handling to its own goroutine, we ensure that reorgs can still be processed even if the main processing loop is stuck in a retry state (indefinetly).
- Implemented `TestSync_ReorgCancelsRetryHandlerInHandleNewBlock` unit test which proves that when reorg is triggered, the `handleNewBlock` retrying gets cancelled.
- Refactored the `evmdriver` unit test to be table tests.

The analysis of thie specific problem was made [here](https://github.com/agglayer/pm/issues/20#issuecomment-3113640027).

## ⚠️ Breaking Changes
- N/A

## 📋 Config Updates
- N/A

## ✅ Testing
- 🤖 **Automatic**: CI checks (`TestSync_ReorgCancelsRetryHandlerInHandleNewBlock` unit test)

## 🐞 Issues
- Closes https://github.com/agglayer/pm/issues/20

## 📝 Notes
- The deliberate decision was made not to add mutexes to the `ProcessBlock` and `Reorg` functions of the `processors`, since they do not rely on any shared state and it seems that the SQLite driver handles the synchronization on a db level (transactions)


sync/evmdriver.go
```diff
@@ -145,43 +148,62 @@ reset:
...
-		case <-ctx.Done():
-			d.log.Info("sync stopped due to context done")
-			cancel()
-			return
-		case b, ok := <-downloadCh:
-			if ok {
-				// when channel is closing, it is sending an empty block with num = 0, and empty hash
-				// because it is not passing object by reference, but by value, so do not handle that since it is closing
+	// Block processing goroutine
+	blockProcessingDone := make(chan struct{})
```
vcastellm, line 152: Where is this channel filled?

Stefan-Ethernal: In the `evmdownloader`:
- https://github.com/agglayer/aggkit/blob/0b6e54d41aec4ed731b5de84bddf9f5a1bcf54b3/sync/evmdownloader.go#L212
- https://github.com/agglayer/aggkit/blob/0b6e54d41aec4ed731b5de84bddf9f5a1bcf54b3/sync/evmdownloader.go#L225-L228

sync/evmdriver.go
```diff
@@ -145,106 +148,130 @@ reset:
...
-		if succeed {
-			break
-		}
 	}
 }
 
-func (d *EVMDriver) handleReorg(ctx context.Context, cancel context.CancelFunc, firstReorgedBlock uint64) {
-	// stop downloader
-	cancel()
+type permanentError struct {
```
goran-ethernal: Maybe call it something like `stopProcessError`? It makes more sense to me. Permanent error seems like you want to stop the processing until you restart `aggkit`. WDYT?

Stefan-Ethernal: I've renamed it to `stopRetryError`, wdyt? https://github.com/agglayer/aggkit/pull/778/commits/bd2a5b6d4bc29145e5de8bc3e0a78880a42d6de0

goran-ethernal (approved):

arnaubennassar (approved):

PR #808, opened 2025-08-01
feat: store certificates in file

vcastellm:
## 🔄 Changes Summary

This pull request introduces functionality to handle signed certificate files by saving their content to disk and updating the database to store file paths instead of the content. Additionally, it includes logic to read the content back from files when necessary. The changes improve storage efficiency and maintainability for signed certificates.

### File Handling for Signed Certificates:
* Added a method `saveSignedCertificateToFile` in `aggsender/db/aggsender_db_storage.go` to save signed certificate content to files and return the file path. This method determines the file storage location based on the database directory and handles file writing.
* Updated `SaveOrUpdateCertificate` and `SaveLastSentCertificate` methods in `aggsender/db/aggsender_db_storage.go` to use `saveSignedCertificateToFile` for storing signed certificate content as file paths before database operations. [[1]](diffhunk://#diff-e381a7a01497b1c89fcd51acf63d7f51713a9b25c747ed857fb0ba12b2bfca98R221-R230) [[2]](diffhunk://#diff-e381a7a01497b1c89fcd51acf63d7f51713a9b25c747ed857fb0ba12b2bfca98R284-R315)

### Reading Signed Certificate Content:
* Modified the `toCertificate` method in `aggsender/db/types.go` to check if a signed certificate is stored as a file path. If so, it reads the content from the file and updates the certificate object accordingly. Errors during file reading are logged but do not interrupt the process. [[1]](diffhunk://#diff-3056b58b1356a8f22db2887de7583ab74837d85e75c955da54dde1004d3926e4R42-R54) [[2]](diffhunk://#diff-3056b58b1356a8f22db2887de7583ab74837d85e75c955da54dde1004d3926e4L55-R85)
* Added a helper method `isFilePath` in `aggsender/db/types.go` to determine if a string is likely a file path based on path separators and file extensions.

### Dependency Updates:
* Added imports for `os`, `path/filepath`, and `strings` to support file handling in both `aggsender/db/aggsender_db_storage.go` and `aggsender/db/types.go`. [[1]](diffhunk://#diff-e381a7a01497b1c89fcd51acf63d7f51713a9b25c747ed857fb0ba12b2bfca98R6-R11) [[2]](diffhunk://#diff-3056b58b1356a8f22db2887de7583ab74837d85e75c955da54dde1004d3926e4R6-R12)


vcastellm:
# Signed Certificate File Storage Example

This document demonstrates how the modified `SaveOrUpdateCertificate` function now stores signed certificates as files instead of storing the content directly in the database.

## Overview

The implementation has been modified to:
1. Save the `SignedCertificate` content to a local file in the same directory as the database
2. Store the file path in the `SignedCertificate` field instead of the content
3. Automatically read the file content when retrieving certificates from the database

## File Storage Details

### File Location
Files are stored in the same directory as the aggsender database (specified by `DBPath` in the configuration).

### File Naming Convention
Files are named using the certificate ID in hexadecimal format:
```
signed_cert_{certificateID_hex}.pem
```

For example, if the certificate ID is `0x1234567890abcdef...`, the file would be named:
```
signed_cert_1234567890abcdef....pem
```

## Example Usage

```go
package main

import (
    "context"
    "fmt"
    "path/filepath"

    "github.com/agglayer/aggkit/aggsender/db"
    "github.com/agglayer/aggkit/aggsender/types"
    "github.com/agglayer/aggkit/log"
)

func main() {
    // Setup storage configuration
    dbPath := "/path/to/aggsender.db"
    cfg := db.AggSenderSQLStorageConfig{
        DBPath:                  dbPath,
        KeepCertificatesHistory: false,
    }

    storage, err := db.NewAggSenderSQLStorage(log.WithFields("example"), cfg)
    if err != nil {
        panic(err)
    }

    // Create a certificate with signed certificate content
    signedCertContent := `-----BEGIN CERTIFICATE-----
MIICXjCCAcSgAwIBAgIJAL5KxQFKQKxjMA0GCSqGSIb3DQEBCwUAMEUxCzAJBgNV
BAYTAkFVMRMwEQYDVQQIDApTb21lLVN0YXRlMSEwHwYDVQQKDBhJbnRlcm5ldCBX
aWRnaXRzIFB0eSBMdGQwHhcNMjMwODIwMDgwMDAwWhcNMjQwODIwMDgwMDAwWjBF
MQswCQYDVQQGEwJBVTETMBEGA1UECAwKU29tZS1TdGF0ZTEhMB8GA1UECgwYSW50
ZXJuZXQgV2lkZ2l0cyBQdHkgTHRkMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIB
CgKCAQEAuLHZnN5N4Q0rJmF8X4Dz8sQHbwj9L0RhNhZnQg5XmHGV2Uw3N3PtJkR5
-----END CERTIFICATE-----`

    certificate := types.Certificate{
        Header: &types.CertificateHeader{
            Height:        1,
            CertificateID: common.HexToHash("0x1234567890abcdef1234567890abcdef12345678"),
        },
        SignedCertificate: &signedCertContent,
    }

    // Save the certificate - this will create a file
    ctx := context.Background()
    err = storage.SaveOrUpdateCertificate(ctx, certificate)
    if err != nil {
        panic(err)
    }

    fmt.Println("Certificate saved successfully!")

    // The file should now exist at:
    expectedFilePath := filepath.Join(filepath.Dir(dbPath), "signed_cert_1234567890abcdef1234567890abcdef12345678.pem")
    fmt.Printf("Signed certificate file should be at: %s\n", expectedFilePath)

    // Retrieve the certificate - content will be read from file
    retrievedCert, err := storage.GetCertificateByHeight(1)
    if err != nil {
        panic(err)
    }

    // The retrieved certificate will contain the original content, not the file path
    if retrievedCert.SignedCertificate != nil {
        fmt.Println("Retrieved certificate content:")
        fmt.Println(*retrievedCert.SignedCertificate)
    }
}
```

## What Happens Behind the Scenes

### When Saving a Certificate:

1. **File Creation**: The `saveSignedCertificateToFile` method creates a new file in the database directory
2. **Content Writing**: The signed certificate content is written to the file with 0644 permissions
3. **Path Storage**: The file path replaces the content in the `SignedCertificate` field before storing in the database
4. **Database Storage**: The certificate info (with file path) is saved to the database

### When Retrieving a Certificate:

1. **Database Query**: The certificate info is retrieved from the database (containing the file path)
2. **File Detection**: The `isFilePath` method detects if the `SignedCertificate` field contains a file path
3. **Content Reading**: If it's a file path, the content is read from the file
4. **Content Return**: The original certificate content is returned to the caller

## Benefits

1. **Storage Efficiency**: Large certificate content is not stored directly in the database
2. **File System Benefits**: Can leverage file system features like compression, backup, etc.
3. **Transparency**: The API remains the same - callers don't need to know about the file storage
4. **Scalability**: Better performance for large certificates

## Error Handling

- If file writing fails during save, an error is returned
- If file reading fails during retrieval, an error is logged but the function continues with the file path
- Empty or nil certificate content is handled gracefully (no file is created)

## File Management

Files are created automatically when certificates are saved. Currently, files are not automatically cleaned up when certificates are deleted - this could be added as a future enhancement.

aggsender/db/aggsender_db_storage.go
```diff
@@ -268,8 +281,40 @@ func (a *AggSenderSQLStorage) SaveOrUpdateCertificate(ctx context.Context, certi
 	return nil
 }
 
+// saveSignedCertificateToFile saves the signed certificate content to a file in the same directory as the database
+// and returns the file path
+func (a *AggSenderSQLStorage) saveSignedCertificateToFile(
```
goran-ethernal, line 279: Doesn't this shift us to always save the signed certificate in a file? Could we maybe have some code that will utilize this but only on extremelly large certificates that are > 1GB?

I think saving this in the db has its advantages, because I don't expect a very large number of certificates to be this big, plus we have in our backlog a task to build a retain policy for certificates in db, so we keep the db in some limits when it comes to size.

Or we can return that config field called `SaveCertificateInFile` and let users decide what they want.

WDYT?

aggsender/db/aggsender_db_storage.go
```diff
@@ -268,8 +281,40 @@ func (a *AggSenderSQLStorage) SaveOrUpdateCertificate(ctx context.Context, certi
...
+	certificateID common.Hash,
+	signedCertContent string) (string, error) {
+	// Get the directory where the database is stored
+	dbDir := filepath.Dir(a.cfg.DBPath)
+
+	// Create a filename using the certificate ID
+	fileName := fmt.Sprintf("signed_cert_%s.pem", hex.EncodeToString(certificateID[:]))
```
Stefan-Ethernal: Slightly shorter, WDYT?

```suggestion
	fileName := fmt.Sprintf("signed_cert_%s.pem", certificateID.Hex())
```

aggsender/db/aggsender_db_storage.go
```diff
@@ -215,6 +218,16 @@ func (a *AggSenderSQLStorage) GetLastSettledCertificate() (*types.CertificateHea
 // SaveOrUpdateCertificate saves the certificate in the storage
 // It will insert a new certificate or update the existing one if it has the same height and certificate ID
 func (a *AggSenderSQLStorage) SaveOrUpdateCertificate(ctx context.Context, certificate types.Certificate) error {
```
Stefan-Ethernal, line 219: Shouldn't we delete the file in case we are [removing](https://github.com/agglayer/aggkit/blob/c60e1b885c9b94d9b49e1b2a7b98ad75cc83c374/aggsender/db/aggsender_db_storage.go#L409-L416) the certificate from the db, WDYT?

aggsender/db/aggsender_db_storage.go
Stefan-Ethernal: Why is the extension of this file `.pem`? It is not an SSL certificate...
IMO it is more consistent to be either: `.bin`, `.dat` etc. WDYT?

aggsender/db/aggsender_db_storage.go
Stefan-Ethernal: Also we should figure out if we would like to store the hex encoded binary data in the file, or store it as JSON as it is in the db...

goran-ethernal: I think JSON because it is human readable, and this is only used for debugging and analysis, so JSON seems more appropriate.

vcastellm: I don't like the idea to have a different behaviour for them, the reference is the DB filed, the fact that the blob is stored on separate files or in a single file in the db should be abstracted for the developer, so everything goes to the same place.

> plus we have in our backlog a task to build a retain policy for certificates in db, so we keep the db in some limits when it comes to size.

This should apply also to both pruning records and the files that goes along with them.

> Or we can return that config field called SaveCertificateInFile and let users decide what they want.

If this is transparent for the operator, I don't find it necessary to make it configurable, all files in the DB dir should be backup, they're considered part of the DB. Similar to other DB storages.

> Also we should figure out if we would like to store the hex encoded binary data in the file, or store it as JSON as it is in the db...

As this is for debugging purposes and disk is super cheap + we'll have pruning IMO let's go with JSON, same as @goran-ethernal said

aggsender/db/aggsender_db_storage.go
vcastellm: Yeah, forgot to change to json, done

aggsender/db/aggsender_db_storage.go
vcastellm: Done

aggsender/db/aggsender_db_storage.go
goran-ethernal: I agree we shouldn't have a different behaviour, but I would leave that to the operator. Having the choice to save them or not save them in the file, in my opinion, is the best way to go. We already had that config param in early versions of aggsender, so we could just bring it back.

But, having said this, I would keep the saving of the signed certificate in that field in db, but have a check in code that if the string is more then 1 GB, just leave it empty, or put N/A string.

vcastellm: @goran-ethernal I disagree with this approach, they've already many options and we need to reduce not increment them, I'm big fun of opinionated soft to simplify UX, if you leave the option to the operator we'll need to add the:

> but have a check in code that if the string is more then 1 GB, just leave it empty, or put N/A string.

And having to explain all those conditions vs just saying "the certificates are saved to disk", I prefer the later and no options, just my opinion, wdyt?

goran-ethernal: I agree we should do some automatic checks and stuff where we can, but I do not agree this is one of them. If someone does not want to generate files in its file system, there is no way he can turn this off. 

Our config is already to big, I agree, but, this is a debug option, and it should not be automatic in my opinion, but let the operator decide if he wants to turn debug options or not.

Also, introducing a check if signed certificate string is too big is basically couple of lines of code. This is a quick solution by copilot:
```go
func IsStringLargerThan1GB(s string) bool {
	const oneGB = 1 << 30 // 1GB = 2^30 bytes
	return len(s) > oneGB
}
```

Stefan-Ethernal:
https://github.com/agglayer/aggkit/pull/808#issuecomment-3144648123

This comment is misleading, because AI was thinking it is about the SSL certificate (the example code section is about the SSL certificate), can we update it accordingly?

aggsender/db/aggsender_db_storage.go
```diff
@@ -268,8 +275,34 @@ func (a *AggSenderSQLStorage) SaveOrUpdateCertificate(ctx context.Context, certi
...
+// saveSignedCertificateToFile saves the signed certificate content to a file in the same directory as the database
+// and returns the file path
+func (a *AggSenderSQLStorage) saveSignedCertificateToFile(
+	certificateID common.Hash,
+	signedCertContent string) (string, error) {
+	// Get the directory where the database is stored
+	dbDir := filepath.Dir(a.cfg.DBPath)
+
+	// Create a filename using the certificate ID
+	fileName := fmt.Sprintf("signed_cert_%s.json", hex.EncodeToString(certificateID[:]))
```
Stefan-Ethernal: Nitpick: it should be enough to simplify it this way and let the `Stringer` implementation do it's stuff. WDYT?

```suggestion
	fileName := fmt.Sprintf("signed_cert_%s.json", certificateID)
```

Stefan-Ethernal (approved):
LGTM, although I'm concerned about cleaning this up and preserving data consistency, because portion of data is in the db, and another part is in the file.

vcastellm:
> LGTM, although I'm concerned about cleaning this up and preserving data consistency, because portion of data is in the db, and another part is in the file.

Valid concern but in theory, both saving and removing points are centralized and implemented in the code, it can happen that the user deletes files from disk, but even in this case it's controlled case, but in any case this will be equivalent to deleting files from a KV store folder, basically you cannot expect that it works, right? in our case is even more harmless. I'm quite happy with this TBH

aggsender/db/aggsender_db_storage.go
vcastellm: As mentioned, we have the configuration param to store historical certificates in the DB, so when that is not enabled only a single record and file will be stored in the disk, and that should be the normal operation. So I consider that as the configuration option for creating these multiple debugging files.

goran-ethernal (approved):

PR #809, opened 2025-08-01
fix: local_debug: expose aggkit ports

joanestebanr:
## 🔄 Changes Summary
-Exposing the local ports of `aggkit` produce a conflict if you try to run at the same time.
      - Now you can choose to expose it of use the default ones using flag ` -p: expose same ports as service in kurtosis`
      - Useful if you need to run bridge locally, because you can stop `aggkit-001` and debug it
- Refactor a bit the code to avoid duplications

```
Usage: ./scripts/local_config_fep
  -h: help
  -e: kurtosis enclave name (default op)
  -p: expose same ports as service in kurtosis (RPC and REST)
```

## How to debug `aggsender` (example FEP): 
- Edit file `.github/test_e2e_single_chain_fork12_op_succinct_args.json` and add `"aggkit_components": "aggoracle,bridge",`
- Deploy kurtosis: example
```
./test/run-local-e2e.sh single-l2-network-fork12-op-succinct ~/kurtosis-cdk/kurtosis-cdk-0ca6e29d64aa8ee4f13f8dfe8a1ef36c1225317d -
```
- Now configure locally your aggsender:
```
./scripts/local_config_fep
```
Notice that you don't use the flag `-p` because you don't want to conflict with RPC and REST ports exposed by the running container
- Add json to your local `launch.json` and you can launch in debug `aggsender`

- You can run e2e test to produce certificates and now is going to work because the `bridge` is running inside `aggkit-001` as the e2e expected
```
./test/run-local-e2e.sh single-l2-network-fork12-op-succinct - ~/e2e/e2e-afb105b9b74aec0885d15add406e2a365d7e49b1
```


temaniarpit27 (approved):

vcastellm (approved):
LGTM 💯

PR #824, opened 2025-08-05
feat: move rate limit to agglayer client

### LINKED ISSUE #756: aggsender: integrate RateLimit to agglayer client directly
# Feature Request
Currently `aggsender` have rate limit component to prevent spamming `agglayer`. The idea is to move this object to `agglayer` client so the `validator` can also take advantage of that

temaniarpit27:
## 🔄 Changes Summary
- move rate limiter from aggsender to agglayer client

## ⚠️ Breaking Changes
- 🛠️ **Config**: Removed `AggSender.MaxSubmitCertificateRate`. Migrated this to `AggSender.AgglayerClient.APIRateLimits`

## 🐞 Issues
- Closes #756

## Dependent PRs
- https://github.com/0xPolygon/kurtosis-cdk/pull/761


agglayer/rate_limit_wrapper.go
```diff
@@ -0,0 +1,101 @@
...
+	rateLimiter, exists := r.rateLimiters[methodName]
+	r.mu.RUnlock()
+
+	if !exists {
+		return
+	}
+
+	if sleepTime := rateLimiter.Call(methodName, true); sleepTime != nil {
+		if r.logger != nil {
+			r.logger.Debugf("Rate limit applied for method '%s', slept for %s", methodName, sleepTime.String())
```
joanestebanr: The log must be done before sleeping, also the level of this log must be configurable (warn, info, debug)

agglayer/rate_limit_wrapper.go
```diff
@@ -0,0 +1,101 @@
...
+// RateLimitWrapper wraps an AgglayerClientInterface and applies rate limiting based on configuration
+type RateLimitWrapper struct {
+	client       AgglayerClientInterface
+	rateLimiters map[string]*aggkitcommon.RateLimit
+	logger       *log.Logger
+	mu           sync.RWMutex
+}
+
+// NewRateLimitWrapper creates a new rate limiting wrapper around an agglayer client
+func NewRateLimitWrapper(client AgglayerClientInterface, config ClientConfig, logger *log.Logger) *RateLimitWrapper {
```
joanestebanr: could logger  be an interface `aggkitcommon.Logger`?

agglayer/rate_limit_wrapper.go
```diff
@@ -0,0 +1,101 @@
...
+	aggkitcommon "github.com/agglayer/aggkit/common"
+	"github.com/agglayer/aggkit/log"
+	ethCommon "github.com/ethereum/go-ethereum/common"
+)
+
+// RateLimitWrapper wraps an AgglayerClientInterface and applies rate limiting based on configuration
+type RateLimitWrapper struct {
+	client       AgglayerClientInterface
+	rateLimiters map[string]*aggkitcommon.RateLimit
+	logger       *log.Logger
```
joanestebanr: could logger  be an interface `aggkitcommon.Logger`?

agglayer/rate_limit_wrapper.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/agglayer/aggkit/agglayer/types"
+	aggkitcommon "github.com/agglayer/aggkit/common"
+	"github.com/agglayer/aggkit/log"
+	ethCommon "github.com/ethereum/go-ethereum/common"
+)
+
+// RateLimitWrapper wraps an AgglayerClientInterface and applies rate limiting based on configuration
+type RateLimitWrapper struct {
+	client       AgglayerClientInterface
+	rateLimiters map[string]*aggkitcommon.RateLimit
```
joanestebanr, line 17: Why a pointer (`*aggkitcommon.RateLimit`) instead of a object?

joanestebanr (commented):
Other comments: 
- Set default value for `SendCertificate` as previous default (20 tries in 1h)
- Remove `AggSender.MaxSubmitCertificateRate` from default.go
- Set removed fields as deprecated

agglayer/rate_limit_wrapper.go
temaniarpit27: because we are modifying rate limits in `Call` func

agglayer/rate_limit_wrapper.go
temaniarpit27: modified it

agglayer/rate_limit_wrapper.go
temaniarpit27: done

agglayer/rate_limit_wrapper.go
temaniarpit27: done

aggsender/aggsender.go
```diff
@@ -330,12 +326,6 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certifi
 		a.log.Warn("dry run mode enabled, skipping sending certificate")
 		return certificate, nil
 	}
-	if rateLimitSleepTime := a.rateLimiter.Call("sendCertificate", false); rateLimitSleepTime != nil {
```
temaniarpit27, line 333: @joanestebanr were we actually not limiting the user? do we want to keep the same behaviour as well in current calls?

config/default.go
```diff
@@ -209,6 +209,11 @@ RequireValidatorCall = false
 		MaxRetries = 6 # 1+6 attempts, around 22m
 	[AggSender.AgglayerClient]
 		Cached = false
+		[[AggSender.AgglayerClient.APIRateLimits]]
```
joanestebanr, line 213: Why the   key is `[[`?

config/config.go
```diff
@@ -65,13 +65,15 @@ const (
...
-	l1NetworkConfigURLDeprecatedHint      = "L1NetworkConfig.URL is deprecated, use L1NetworkConfig.RPC.URL instead"
-	reorgDetectorL1DeprecatedHint         = "ReorgDetectorL1 is deprecated, remove it from configuration"
+	aggOracleBlockFinalityDeprecated       = "AggOracle.BlockFinality is deprecated, remove it from configuration"
+	l1InfoTreeSyncBlockFinalityDeprecated  = "L1InfoTreeSync.BlockFinality is deprecated, remove it from configuration"
+	bridgeL1SyncBlockFinalityDeprecated    = "BridgeL1Sync.BlockFinality is deprecated, remove it from configuration"
+	lastGERSyncDeprecatedHint              = "LastGERSync is deprecated, use L2GERSync instead"
+	lastGERSyncSyncModeDeprecatedHint      = "LastGERSync.SyncMode is deprecated, remove it from configuration"
+	l1NetworkConfigURLDeprecatedHint       = "L1NetworkConfig.URL is deprecated, use L1NetworkConfig.RPC.URL instead"
+	reorgDetectorL1DeprecatedHint          = "ReorgDetectorL1 is deprecated, remove it from configuration"
+	maxSubmitCertificateRateDeprecatedHint = "AggSender.MaxSubmitCertificateRate is deprecated, " +
```
joanestebanr, line 75: In fact have been moved to another place, maybe make sense to explain how to configure it

common/rate_limit.go
```diff
@@ -62,37 +62,36 @@ func (r *RateLimit) String() string {
 }
 
 // Call is used before making a call, it will sleep if the rate limit is reached if param allowToSleep is true
-func (r *RateLimit) Call(msg string, allowToSleep bool) *time.Duration {
```
joanestebanr: Why do you remove `*time.Duration`? it make sense that if you set `allowToSleep` to false it returns this value. For sure the way that have been integrated this object `allowToSleep`is always true, and ignored the returned value

agglayer/rate_limit_wrapper.go
```diff
@@ -0,0 +1,111 @@
...
+func (r *RateLimitWrapper) applyRateLimit(methodName string) {
+	r.mu.RLock()
+	rateLimiter, exists := r.rateLimiters[methodName]
+	r.mu.RUnlock()
+
+	if !exists {
+		return
+	}
+
+	rateLimiter.Call(methodName, true)
```
joanestebanr: You are using `rateLimiter` with the mutex **unlocked**

agglayer/rate_limit_wrapper_test.go
```diff
@@ -0,0 +1,595 @@
...
+
+	wrapper := NewRateLimitWrapper(mockClient, config, mockLogger)
+	str := wrapper.String()
+	require.Contains(t, str, "RateLimitWrapper")
+	require.Contains(t, str, "SendCertificate")
+
+	mockLogger.AssertExpectations(t)
+}
+
+func TestNewRateLimitWrapper_WithNilLogger_Panics(t *testing.T) {
```
vcastellm: too verbose?

vcastellm (approved):
LGTM

config/default.go
temaniarpit27: Because we can have different methods and their rate limits.
so if we want to add a new method name, we can just add similarly here.
this is how toml files accept arrays

agglayer/rate_limit_wrapper.go
temaniarpit27: fixed

common/rate_limit.go
temaniarpit27: fixed

config/config.go
temaniarpit27: fixed

agglayer/rate_limit_wrapper.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	return wrapper
+}
+
+// applyRateLimit applies rate limiting for the given method name
+func (r *RateLimitWrapper) applyRateLimit(methodName string) {
+	r.mu.RLock()
+	rateLimiter, exists := r.rateLimiters[methodName]
+	r.mu.RUnlock()
```
joanestebanr, line 49: maybe you can defer the unlock(), WDTY?

temaniarpit27: but defer will then include sleep time as well which can increase the lock for a longer period of time

temaniarpit27: read and write can be in 1 lock but i dont think time.sleep should be included in lock

joanestebanr (approved):

PR #826, opened 2025-08-05
fix: when bridge l1 syncer is stopped and l1 info tree syncer works

temaniarpit27:
## 🔄 Changes Summary
- bug fix related to l1 info tree index not found

Closes - #827


bridgeservice/bridge.go
```diff
@@ -968,7 +968,12 @@ func (b *BridgeService) getFirstL1InfoTreeIndexForL1Bridge(ctx context.Context,
 
 	root, err := b.bridgeL1.GetRootByLER(ctx, lastInfo.MainnetExitRoot)
 	if err != nil {
-		return 0, err
+		b.logger.Infof("failed to get root by LER for L1: %v, lastInfo MainnetExitRoot: %v", err, lastInfo.MainnetExitRoot)
```
Stefan-Ethernal: Should we indicate in the log that we are trying with the fallback mechanism?

bridgeservice/bridge.go
```diff
@@ -968,7 +968,12 @@ func (b *BridgeService) getFirstL1InfoTreeIndexForL1Bridge(ctx context.Context,
...
+		root, err = b.bridgeL1.GetLastRoot(ctx)
+		if err != nil {
+			b.logger.Errorf("failed to get last root for L1: %v", err)
```
Stefan-Ethernal: It should be logged at the top lvl, so maybe we can remove this log from here, but just provide enough context in the err msg itself. WDYT?

PR #831, opened 2025-08-06
fix: bridge service db issues

rachit77:
## 🔄 Changes Summary
- This PR fixes the error we were noticing on bridge service `cdk22`

## 🐞 Issues
- Closes #825

PR #838, opened 2025-08-07
feat: mechanism to calculate last settled block by not using `metadata`

goran-ethernal:
## 🔄 Changes Summary
This PR introduces a new interface called `CertificateQuerier` which, for now, has only one function called `GetLastSettledCertificateToBlock` which calculates the last settled block from a previously settled certificate.

This will be used on the validator, to calculate what was the last settled block, so we can figure out from which block a new certificate that is being verified is built.

It will also be used on startup from scratch, when `aggsender` syncs its state with `agglayer`, so we can save the `ToBlock` of the last settled certificate in the `db`.

The motivation behind this PR is to not use the `Metadata` field on the certificate, since it will not be signed by the proposer and validators, hence, it can be faked, but deduce the block range from other certificate data.

## How the last settled block is calculated

The last settled block in the `CertificateQuerier` is calculated like this:
- First it will check the `NewLocalExitRoot` field on the certificate, if it is not the first `empty` LER, meaning, there were some bridge exits in the certificate, or in certificates before it, it will get the block num on which the `NewLocalExitRoot` hash was added on `L2` by calling the `L2 bridge syncer`.
- Secondly, it will call the `agglayer` endpoint to get the latest settled imported bridge exit. Note that this PR added only the endpoint sceleton, since this is not yet implemented on `agglayer`. This endpoint will return the `GlobalIndex` of the last settled imported bridge exit (claim), and we will use that to get the claim from the `L2 bridge syncer` since `GlobalIndex` is unique per claim.
- Thirdly, it will ping the `AggchainFEP` rollup contract to get the last settled `L2` block from it (which is gotten from the `aggchain proof` in the `FEP` certificates). If the network is not `FEP` this call will just return 0.
- At the end, it will compare the 3 blocks gotten in previous steps, and choose the max value from it.

We need all three values because, PP certificates can either contain only bridge exits, or only imported bridge exits, and FEP certificates can be empty.

**IMPORTANT NOTE**: This code is not used anywhere. It will be used in subsequent PRs.
 

## ⚠️ Breaking Changes
NA

## 📋 Config Updates
NA

## ✅ Testing
- 🤖 **Automatic**: aggkit CI


aggsender/query/certificate_query.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// TODO - this might need to be changed once agglayer gives support for this
+	// 2. Get the latest settled imported bridge exit block number
+	latestSettledIbe, err := c.agglayerClient.GetLatestSettledImportedBridgeExit(ctx)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get latest settled imported bridge exit from agglayer: %w", err)
+	}
+
+	if latestSettledIbe != nil {
+		bigGlobalIndex := latestSettledIbe.ToBigInt()
```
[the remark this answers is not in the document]
goran-ethernal: Fixed: https://github.com/agglayer/aggkit/pull/838/commits/47f9a452a30967b18b3eaf2c4e12e44bb929b620

bridgesync/processor.go
```diff
@@ -430,6 +430,30 @@ func (p *processor) GetClaims(ctx context.Context, fromBlock, toBlock uint64) ([
...
 }
 
+func (p *processor) GetClaimByGlobalIndex(ctx context.Context, globalIndex *big.Int) (Claim, error) {
+	tx, err := p.startTransaction(ctx, true)
+	if err != nil {
+		return Claim{}, err
+	}
+	defer p.rollbackTransaction(tx)
+
+	var claim Claim
```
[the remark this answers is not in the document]
goran-ethernal: Fixed: https://github.com/agglayer/aggkit/pull/838/commits/47f9a452a30967b18b3eaf2c4e12e44bb929b620

agglayer/types/types.go
```diff
@@ -471,6 +471,10 @@ type GlobalIndex struct {
 	LeafIndex   uint32 `json:"leaf_index"`
 }
 
+func (g *GlobalIndex) ToBigInt() *big.Int {
```
Stefan-Ethernal, line 475: Nitpick: consider adding the comment

agglayer/mock_agglayer_client.go
```diff
@@ -260,6 +260,64 @@ func (_c *AgglayerClientMock_GetLatestSettledCertificateHeader_Call) RunAndRetur
 	return _c
 }
```
Stefan-Ethernal, line 262: Is it feasible to move this mock to the `mocks` pkg alongside other mocks (because of potential import cycles etc.)?

bridgesync/processor.go
```diff
@@ -430,6 +430,34 @@ func (p *processor) GetClaims(ctx context.Context, fromBlock, toBlock uint64) ([
...
+	}
+
+	tx, err := p.startTransaction(ctx, true)
+	if err != nil {
+		return Claim{}, err
+	}
+	defer p.rollbackTransaction(tx)
+
+	var claim Claim
+	if err = meddler.QueryRow(tx, &claim, fmt.Sprintf(`
```
Stefan-Ethernal: We don't need the transactions in case of reading (refer to the #831)
```suggestion
	var claim Claim
	if err = meddler.QueryRow(p.db, &claim, fmt.Sprintf(`
```

bridgesync/bridgesync_test.go
```diff
@@ -132,7 +132,7 @@ func TestGetLastProcessedBlock(t *testing.T) {
 
 func TestGetBridgeRootByHash(t *testing.T) {
```
Stefan-Ethernal:
```suggestion
func TestGetExitRootByHash(t *testing.T) {
```

bridgesync/bridgesync.go
```diff
@@ -326,13 +326,21 @@ func (s *BridgeSync) GetLastProcessedBlock(ctx context.Context) (uint64, error)
...
+func (s *BridgeSync) GetExitRootByHash(ctx context.Context, root common.Hash) (*tree.Root, error) {
 	if s.processor.isHalted() {
 		return nil, sync.ErrInconsistentState
 	}
 	return s.processor.exitTree.GetRootByHash(ctx, root)
 }
 
+func (s *BridgeSync) GetClaimByGlobalIndex(ctx context.Context, globalIndex *big.Int) (Claim, error) {
+	if s.processor.isHalted() {
+		s.processor.log.Error("processor is halted, cannot get claim by global index")
```
Stefan-Ethernal: Suggesting to remove this, because of consistency. WDYT?

```suggestion
```

aggsender/query/certificate_query.go
```diff
@@ -0,0 +1,107 @@
...
+	agglayertypes "github.com/agglayer/aggkit/agglayer/types"
+	"github.com/agglayer/aggkit/aggsender/types"
+)
+
+var _ types.CertificateQuerier = (*certificateQuerier)(nil)
+
+// certificateQuerier handles querying for certificate ranges
+// settled and pending certificates
+type certificateQuerier struct {
+	bridgeSyncer       types.L2BridgeSyncer
```
Stefan-Ethernal: WDYT?

```suggestion
	l2BridgeSyncer       types.L2BridgeSyncer
```

Stefan-Ethernal (approved):

aggsender/query/certificate_query.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/838/commits/a0a491d0e1a9299e71c55ce098802a68184555fe

bridgesync/bridgesync.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/838/commits/a0a491d0e1a9299e71c55ce098802a68184555fe

bridgesync/bridgesync_test.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/838/commits/a0a491d0e1a9299e71c55ce098802a68184555fe

bridgesync/processor.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/838/commits/a0a491d0e1a9299e71c55ce098802a68184555fe

agglayer/mock_agglayer_client.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/838/commits/a0a491d0e1a9299e71c55ce098802a68184555fe

agglayer/types/types.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/838/commits/a0a491d0e1a9299e71c55ce098802a68184555fe

PR #846, opened 2025-08-11
feat: decouple `aggsender-validator` from `metadata` field

goran-ethernal:
## 🔄 Changes Summary
This PR decouples `aggsender-validator` from using the `Metadata` field in certificate to figure out block range.

The validator will now use the `certificate querier` interface to get the `ToBlock` from the last settled certificate in order to figure out if the new certificate is valid or not.

## ⚠️ Breaking Changes
NA

## 📋 Config Updates
NA

## ✅ Testing

## 🔗 Related PRs
- #838
- #839

### DIFF
--- a/aggsender/flows/factory.go
+++ b/aggsender/flows/factory.go
@@ -37,27 +37,20 @@ func NewFlow(
 ) (types.AggsenderFlow, error) {
 	switch types.AggsenderMode(cfg.Mode) {
 	case types.PessimisticProofMode:
-		signer, err := initializeSigner(ctx, cfg.AggsenderPrivateKey, logger)
-		if err != nil {
-			return nil, err
-		}
-		logger.Infof("Initializing RollupManager contract at address: %s. Genesis block: %d",
-			cfg.RollupManagerAddr, cfg.RollupCreationBlockL1)
-		lerQuerier := query.NewLERDataQuerier(cfg.RollupCreationBlockL1, rollupDataQuerier)
-		l2BridgeQuerier := query.NewBridgeDataQuerier(logger, l2Syncer, cfg.DelayBetweenRetries.Duration)
-		l1InfoTreeQuerier := query.NewL1InfoTreeDataQuerier(l1Client, l1InfoTreeSyncer)
-		logger.Infof("Aggsender signer address: %s", signer.PublicAddress().Hex())
-		baseFlow := NewBaseFlow(
-			logger, l2BridgeQuerier, storage, l1InfoTreeQuerier, lerQuerier,
-			NewBaseFlowConfig(cfg.MaxCertSize, 0, false),
+		commonFlowComponents, err := createCommonComponents(
+			ctx, cfg, logger, storage, l1Client, l1InfoTreeSyncer, l2Syncer, rollupDataQuerier, 0, false,
 		)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create common flow components: %w", err)
+		}
+
 		return NewPPFlow(
 			logger,
-			baseFlow,
+			commonFlowComponents.baseFlow,
 			storage,
-			l1InfoTreeQuerier,
-			l2BridgeQuerier,
-			signer,
+			commonFlowComponents.l1InfoTreeDataQuerier,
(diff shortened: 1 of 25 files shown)


aggsender/query/ler_query.go
```diff
@@ -52,3 +50,17 @@ func (l *lerDataQuerier) GetLastLocalExitRoot() (common.Hash, error) {
...
+	if err != nil {
+		return common.Hash{}, fmt.Errorf("failed to get last local exit root: %w", err)
+	}
+
+	if ler == aggkitcommon.ZeroHash {
+		return types.EmptyLER, nil
+	}
+
+	return ler, nil
+}
```
Stefan-Ethernal: Why don't we merge this function into `GetLastLocalExitRoot` (https://github.com/agglayer/aggkit/blob/416c39c080e8e3df9b1511430c038ffc97a6e545/aggsender/query/ler_query.go#L45-L52)? Especially because I find the `GetStartLER` name misleading (it is more convenient to name it `GetLastLocalExitRoot`, and if we merge the logic into that one it is a win win situation IMO).

aggsender/flows/factory.go
```diff
@@ -43,12 +43,7 @@ func NewFlow(
 		}
 		logger.Infof("Initializing RollupManager contract at address: %s. Genesis block: %d",
 			cfg.RollupManagerAddr, cfg.RollupCreationBlockL1)
-		lerQuerier, err := query.NewLERDataQuerier(
-			cfg.RollupManagerAddr, cfg.RollupCreationBlockL1, rollupDataQuerier)
-		if err != nil {
-			return nil, fmt.Errorf("error creating LER data querier: %w", err)
-		}
-
+		lerQuerier := query.NewLERDataQuerier(cfg.RollupCreationBlockL1, rollupDataQuerier)
```
Stefan-Ethernal: IMO you could extract the instatiation of the common dependencies (e.g. lerQuerier, l2BridgeQuerier, l1InfoTreeQuerier etc.), before the switch-case and reduce duplication (at least slightly).

If you are concerned about instantiating before validating the aggsender mode, we can first validate if the mode is within the valid range of values and if not, return early. WDYT?

Stefan-Ethernal (commented):
Aren't we yet ready to remove the certificate metadata (https://github.com/agglayer/aggkit/blob/416c39c080e8e3df9b1511430c038ffc97a6e545/aggsender/types/certificate_metadata.go#L19-L38)? AFAIK it is now used only in some tests.

aggsender/validator/compare_certs.go
```diff
@@ -41,22 +40,8 @@ func DiffsCertificate(
 	}
 
 	if validatingCertificate.Metadata != expectedCertificate.Metadata {
```
Stefan-Ethernal, line 42: Are we going to remove the metadata field? Or we have agreed to keep it relaying between the aggsender proposer and validators?

goran-ethernal:
> Aren't we yet ready to remove the certificate metadata 

@Stefan-Ethernal  That will be done in a subsequent PR. I wanted to make these PRs as small and concise as possible, so that they are more easier to review.

aggsender/query/ler_query.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/846/commits/95a6b77ae12cb374017c4a58f4ca51f7fa1f8b5c

aggsender/validator/compare_certs.go
goran-ethernal: No, the metadata field will remain, because `agglayer` will not remove it because of backwards compatibility, and because they use it for calculating `CertificateID`, but it will be enforced as `ZeroHash`.

goran-ethernal: But `aggsender` doesn't need it for anything anymore.

aggsender/flows/factory.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/846/commits/095099ed65b60a611b24a94e6bdfdf864ca58e7d

Stefan-Ethernal (approved):

Issue #803, opened 2025-07-31
feat: aggsender-validator: support PP or FEP by configuration


joanestebanr:
# Feature Request

You must be able to configure the validator mode:
```
[Validator]
Mode = "PessimisticProof" # PessimisticProof | AggchainProof
```
where `AggchainProof` is for FEP and `PessimisticProof` for PP

For PP we have the struct `Validator.PPConfig` to specific configuration, we need to create another struct for FEP parameters.
Note: Let the default.go values for the new fields point to `aggsender` section

PR #863, opened 2025-08-12
feat: last L2 block in certificate sanity check on `aggsender-validator`

### LINKED ISSUE #805: feat: aggsender-validator: sanity check LastL2Block
# Feature Request

Must check that lastL2block passed as grpc param is >= to the last included block in certificate (based on NewLER + imported bridges exits)

goran-ethernal:
## 🔄 Changes Summary
This PR adds a sanity check on the `aggsender-validator` for the `LastL2BlockInCert` field which represents the last L2 block in the proposed certificate.

- `LastL2BlockInCert` can not be lower or equal to the last settled block in network.
- `LastL2BlockInCert` can not be lower than the deduced `ToBlock` in the proposed certificate gotten by checking the `NewLocalExitRoot` and imported bridge exits in the certificate.

## ⚠️ Breaking Changes
N/A

## 📋 Config Updates
N/A

## ✅ Testing

## 🐞 Issues
- Closes #805


aggsender/validator/validate_certificate.go
```diff
@@ -225,3 +235,27 @@ func (a *CertificateValidator) getCertificatePreBuildParams(
...
 	}, nil
 }
+
+// validateLastL2BlockInCert checks that the provided last L2 block in the certificate by the proposer
+// is greater or equal to the blocks we see in the new certificate
+func (a *CertificateValidator) validateLastL2BlockInCert(
+	ctx context.Context,
+	req types.VerifyIncomingRequest,
+	lastSettledBlock uint64) error {
+	if req.LastL2BlockInCert <= lastSettledBlock {
```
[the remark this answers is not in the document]
goran-ethernal: Not true. If `lastSettledBlock = 0` all positive values of `LastL2BlockInCert` are accepted. If `LastL2BlockInCert = 0` it is an invalid value and should return an error. Certificate has to have some blocks in it.

aggsender/validator/validate_certificate.go
```diff
@@ -68,6 +68,18 @@ func (a *CertificateValidator) ValidateCertificate(ctx context.Context, params t
...
+			return fmt.Errorf("failed to get last settled certificate block: %w", err)
+		}
+	}
+
+	// Validate last L2 block in certificate
+	if err := a.validateLastL2BlockInCert(ctx, params, previousCertificateToBlock); err != nil {
+		return fmt.Errorf("failed to validate last L2 block in new certificate: %w", err)
+	}
+
 	// Between cert must be no gap because if there are could be a attack vector
```
Stefan-Ethernal:
```suggestion
	// Between cert must be no gap because if there are could be an attack vector
```

aggsender/query/certificate_query.go
```diff
@@ -106,15 +103,83 @@ func (c *certificateQuerier) GetLastSettledCertificateToBlock(
...
+	cert *agglayertypes.Certificate, certToBlock uint64) types.CertificateType {
+	if cert.AggchainData != nil {
+		// if AggchainData is present, we can use it to determine the certificate type
+		_, ok := cert.AggchainData.(*agglayertypes.AggchainDataProof)
+		if !ok {
+			return types.CertificateTypeFEP
+		}
+
 		return types.CertificateTypePP
 	}
```
Stefan-Ethernal, line 148: Wasn't it enough to determine the certificate type based solely on the previous logic (`CalculateCertificateTypeFromToBlock`)? It doesn't look like an expensive checks in the `CalculateCertificateTypeFromToBlock` (it is just a question whether it is completely reliable). WDYT?

IMO it would be preferably to rely on the same logic for determining the certificate type on all occasions.

aggsender/query/certificate_query.go
```diff
@@ -106,15 +103,83 @@ func (c *certificateQuerier) GetLastSettledCertificateToBlock(
...
+	if c.aggchainFEPQuerier.IsFEP() {
+		// if we are in a FEP network, we can determine the type based on the start of the FEP
+		if certToBlock < c.aggchainFEPQuerier.StartL2Block() {
+			// if the certificate is for a block before the start of the FEP, it is a PP certificate
+			return types.CertificateTypePP
+		}
+		// otherwise, it is a FEP certificate
+		return types.CertificateTypeFEP
+	}
+
```
Stefan-Ethernal, line 165: We have also the `CertificateTypeOptimistic` (namely check if the optimistic mode is enabled and return it in case it is a FEP certificate). Do we need to cover that case here as well?

aggsender/validator/validate_certificate.go
```diff
@@ -225,3 +235,27 @@ func (a *CertificateValidator) getCertificatePreBuildParams(
...
+		return fmt.Errorf("last L2 block in certificate %d must be greater than last settled block %d",
+			req.LastL2BlockInCert, lastSettledBlock)
+	}
+
+	newCertToBlock, err := a.certQuerier.GetNewCertificateToBlock(ctx, req.Certificate)
+	if err != nil {
+		return fmt.Errorf("failed to get new certificate to block: %w", err)
+	}
+
+	if newCertToBlock > req.LastL2BlockInCert {
```
Stefan-Ethernal, line 255: This is unclear to me why it is ok to be less (or equal). I thought that this needs to be strictly equal?

aggsender/validator/validate_certificate.go
```diff
@@ -225,3 +235,27 @@ func (a *CertificateValidator) getCertificatePreBuildParams(
...
 }
+
+// validateLastL2BlockInCert checks that the provided last L2 block in the certificate by the proposer
+// is greater or equal to the blocks we see in the new certificate
+func (a *CertificateValidator) validateLastL2BlockInCert(
+	ctx context.Context,
+	req types.VerifyIncomingRequest,
+	lastSettledBlock uint64) error {
+	if req.LastL2BlockInCert <= lastSettledBlock {
+		return fmt.Errorf("last L2 block in certificate %d must be greater than last settled block %d",
```
Stefan-Ethernal: Nitpick:

```suggestion
		return fmt.Errorf("the last L2 block in the certificate (%d) must be greater than the last settled block (%d)",
```

aggsender/query/certificate_query.go
```diff
@@ -65,17 +66,13 @@ func (c *certificateQuerier) GetLastSettledCertificateToBlock(
...
-				cert.NewLocalExitRoot.String(), err)
-		}
-
-		lastBridgeExitBlock = newLER.BlockNum
+	// if NewLER is not the first empty LER, it means that the certificate
+	// or certificate before it had bridge exits, so we can use it to
+	// to determine the last bridge exit block
+	lastBridgeExitBlock, err = c.getBlockNumFromLER(ctx, cert.NewLocalExitRoot)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get exit root by hash using NewLocalExitRoot %s: %w",
```
Stefan-Ethernal: This error message needs to be adapted, something like


```suggestion
		return 0, fmt.Errorf("failed to resolve the bridge exit block number for NewLocalExitRoot %s: %w",
```

aggsender/query/certificate_query.go
```diff
@@ -106,15 +103,83 @@ func (c *certificateQuerier) GetLastSettledCertificateToBlock(
...
+		// if there are imported bridge exits, we can use the last one to determine the new certificate to block
+		lastImportedBridgeExit := cert.ImportedBridgeExits[len(cert.ImportedBridgeExits)-1]
+		bigGlobalIndex := lastImportedBridgeExit.GlobalIndex.ToBigInt()
+		claim, err := c.l2BridgeSyncer.GetClaimByGlobalIndex(ctx, bigGlobalIndex)
+		if err != nil {
+			return 0, fmt.Errorf("failed to get claim by global index %s: %w", bigGlobalIndex.String(), err)
+		}
+
+		lastImportedBridgeExitBlock = claim.BlockNum
 	}
```
Stefan-Ethernal: Can we extract this to a common function and reuse it in the `GetLastSettledCertificateToBlock` (https://github.com/agglayer/aggkit/blob/aa01492a9119b51b56a0c0f09a00e8018c0a170c/aggsender/query/certificate_query.go#L54) as well. WDYT?

aggsender/query/certificate_query.go
goran-ethernal: No it is not. In the case when someone is sending a new full certificate, it comes with the proof (in case of FEP certificate), or with a signature (in case of PP), so `ToBlock` is not really accurate enough, because what if someone called a prover to generate the proof, or what if he didn't called it but it should have. This can be used for additional checks.

When syncing with agglayer from scratch, we do not know this info, so we rely only on the ToBlock.

aggsender/query/certificate_query.go
goran-ethernal: There is no way to know if the certificate is Optimistic or not, since it contains the same data as a FEP certificate. Only proposer knows this because it checks the contract if the optimistic mode was turned on or not.  This certificate type is essentially only a construct of `aggsender`. It is not something `agglayer` saves or takes care of, so we can only guess it as best as we can.

aggsender/validator/validate_certificate.go
goran-ethernal: No it doesn't need to be strictly equal. A proposer could have created a certificate from blocks 10-20, but in last two blocks, 19-20, it didn't have any bridge transactions. So, when trying to guess the `ToBlock` of a certificate we rely on `NewLocalExitRoot` and imported bridge exits in the certificate, so if the last two blocks didn't have any, we will determine a lower block as `ToBlock`, but the validator will use the `LastL2BlockInCert` to build a certificate. If there were any bridge transactions in those two blocks, but certificate tried to fake it, we will know, and return an error.

aggsender/query/certificate_query.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/863/commits/c26ec0fc78d2e5b08ad3c9667e4a78f3af6b5668

aggsender/query/certificate_query.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/863/commits/c26ec0fc78d2e5b08ad3c9667e4a78f3af6b5668

aggsender/validator/validate_certificate.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/863/commits/c26ec0fc78d2e5b08ad3c9667e4a78f3af6b5668

aggsender/validator/validate_certificate.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/863/commits/24fc74782ed72fd4c1903444265abb81cefb910f

aggsender/query/certificate_query.go
goran-ethernal: This will be addressed in: https://github.com/agglayer/aggkit/issues/803

Stefan-Ethernal (approved):

Issue #835, opened 2025-08-07
[BUG] Bridge service context deadline exceeded


Stefan-Ethernal:
There are some occasions where the context times out. In this particular case, we were reading the db content and the thing is we are reusing the `ctx` that was created with the timeout for the request. It is worth considering whether we can decouple these timeouts or just increase the [REST.ReadTimeout](https://github.com/agglayer/aggkit/blob/82bb0be308ae9bdc8f6bdd194c6f1afab69c03f7/config/default.go#L138) by default...

```go
failed to convert token mappings to the object model: context deadline exceeded
```

```go
github.com/agglayer/aggkit/bridgesync.(*processor).fetchTokenMappings
	/app/bridgesync/processor.go:883
github.com/agglayer/aggkit/bridgesync.(*processor).GetTokenMappings
	/app/bridgesync/processor.go:845
github.com/agglayer/aggkit/bridgesync.(*BridgeSync).GetTokenMappings
	/app/bridgesync/bridgesync.go:341
github.com/agglayer/aggkit/bridgeservice.(*BridgeService).GetTokenMappingsHandler
	/app/bridgeservice/bridge.go:493
github.com/gin-gonic/gin.(*Context).Next
	/go/pkg/mod/github.com/gin-gonic/gin@v1.10.1/context.go:185
github.com/agglayer/aggkit/bridgeservice.New.LoggerHandler.func1
	/app/bridgeservice/bridge.go:146
github.com/gin-gonic/gin.(*Context).Next
	/go/pkg/mod/github.com/gin-gonic/gin@v1.10.1/context.go:185
github.com/gin-gonic/gin.CustomRecoveryWithWriter.func1
	/go/pkg/mod/github.com/gin-gonic/gin@v1.10.1/recovery.go:102
github.com/gin-gonic/gin.(*Context).Next
	/go/pkg/mod/github.com/gin-gonic/gin@v1.10.1/context.go:185
github.com/gin-gonic/gin.(*Engine).handleHTTPRequest
	/go/pkg/mod/github.com/gin-gonic/gin@v1.10.1/gin.go:644
github.com/gin-gonic/gin.(*Engine).ServeHTTP
	/go/pkg/mod/github.com/gin-gonic/gin@v1.10.1/gin.go:600
net/http.serverHandler.ServeHTTP
	/usr/local/go/src/net/http/server.go:3301
net/http.(*conn).serve
	/usr/local/go/src/net/http/server.go:2102
```

PR #874, opened 2025-08-13
feat: use bridge db query timeout

rachit77:
## 🔄 Changes Summary
- This PR adds db query timeout for db tx

### Config Changes
- Added `DBQueryTimeout`  to `[BridgeL1Sync]` section
- Added `DBQueryTimeout`  to `[BridgeL2Sync]` section

### Purpose
Enables configurable database query timeouts for bridge db operations.

Issue: #835

### DIFF
--- a/bridgesync/processor.go
+++ b/bridgesync/processor.go
@@ -344,13 +344,13 @@ func (b BridgeSyncRuntimeData) IsCompatible(storage BridgeSyncRuntimeData) error
 }
 
 type processor struct {
-	db                   *sql.DB
-	exitTree             *tree.AppendOnlyTree
-	log                  *log.Logger
-	mu                   mutex.RWMutex
-	halted               bool
-	haltedReason         string
-	databaseQueryTimeout time.Duration
+	db             *sql.DB
+	exitTree       *tree.AppendOnlyTree
+	log            *log.Logger
+	mu             mutex.RWMutex
+	halted         bool
+	haltedReason   string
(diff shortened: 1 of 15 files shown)


temaniarpit27 (approved):
lgtm
can we have a testcase to check if db query timeout is working as expected?
maybe decrease the query timeout to nanosec and what err it returns

rachit77:
> lgtm can we have a testcase to check if db query timeout is working as expected? maybe decrease the query timeout to nanosec and what err it returns

done: bbbe8c54b42cb809c8b85bbb36444e8f5408c1de
@temaniarpit27

bridgesync/processor.go
```diff
@@ -1005,3 +1043,7 @@ func (p *processor) unhalt() {
 	p.haltedReason = ""
 	p.log.Info("processor unhalted")
 }
+
+func (p *processor) withDatabaseTimeout(ctx context.Context) (context.Context, context.CancelFunc) {
+	return context.WithTimeout(context.WithoutCancel(ctx), p.dbQueryTimeout)
```
goran-ethernal: Do we want to detach it from the parent context with `WithoutCancel`? I think we should take into consideration if the parent context cancels before this one expires as well.

rachit77: fixed: ff393029a3b74fb5af17942ac6a87308a381a59a

goran-ethernal (approved):

PR #877, opened 2025-08-14
feat: poll committee for signatures

### LINKED ISSUE #793: feat: aggsender-proposer: ask to multi aggsender-validator for a signature
# Feature Request
- Configuration of validator #796 
- Threshold
- Call to validators (parallel?)
- Add signature to certificate
- Submit to agglayer depends on #794

Stefan-Ethernal:
## 🔄 Changes Summary
Intregrate multisig committee in the aggsender proposer. On each certificate submission, the aggsender proposer queries the smart contract for the actual committee and delegates the certificate validation to the validators and aggregates the signatures before sending them alongside with the certificate to the `agglayer`.

## ✅ Testing
- 🤖 **Automatic**: CI
- 🖱️ **Manual**: N/A

## 🐞 Issues
- Closes #793

## 📝 Notes
- N/A

## TODOs:

- [x] Remove empty url [validation](https://github.com/agglayer/aggkit/blob/0c31ef8ff11fe183682ddd79b5fdc92ac6a00963/aggsender/config/config.go#L129-L131) as they are retrieved from the contract


cmd/run.go
```diff
@@ -94,6 +94,13 @@ func start(cliCtx *cli.Context) error {
 	l2GERSync := runL2GERSyncIfNeeded(
 		cliCtx.Context, components, cfg.L2GERSync, reorgDetectorL2, l2Client, l1InfoTreeSync,
 	)
+
+	committeeQuerier, err := query.NewECDSAMultisigCommitteeQuery(cfg.L1NetworkConfig.RollupAddr, l1Client)
```
goran-ethernal: Maybe use the `runIfNedded` pattern for this as well?

aggsender/aggsender.go
```diff
@@ -416,24 +428,120 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certifi
...
+
+	for _, v := range validators {
+		g.Go(func() error {
+			status, err := v.HealthCheck(ctx)
+			if err != nil {
+				a.log.Warnf("error checking validator health (URL=%s): %v", v.URL(), err)
+			}
+
+			if !status.IsHealthy() {
+				a.log.Warnf("validator status (URL=%s) is not healthy: %s", v.URL(), status.String())
```
goran-ethernal: Maybe return an error here?

aggsender/aggsender.go
```diff
@@ -416,24 +428,120 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certifi
...
+
+	// Wait for all goroutines to finish in background
+	go func() {
+		_ = g.Wait()
+		close(sigCh)
+		close(errCh)
+	}()
+
+	signatures := make([][]byte, 0, len(validators))
+	for sig := range sigCh {
```
goran-ethernal: Hmm, the logic with multiple channels is really complex to me. Maybe we can just add a `mutex` and use it to write to signature and errors arrays. It will also decrease the lines of code in the function. WDYT?

aggsender/aggsender.go
```diff
@@ -416,24 +428,120 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certifi
...
+	// Collect all errors from errCh
+	errs := make([]error, 0)
+	for err := range errCh {
+		errs = append(errs, err)
+	}
+
+	// If we didn’t reach threshold, return combined error
+	if uint32(len(signatures)) < signaturesThreshold {
+		if len(errs) > 0 {
+			return signatures, errors.Join(errs...)
```
goran-ethernal, line 519: I would perhaps log the errors returned by validators even if we have a threshold. WDYT?

aggsender/aggsender.go
Stefan-Ethernal: The idea was to just log and skip the unhealthy validators. Seems fine by me, but unsure if you are concerned about something particular?

cmd/run.go
Stefan-Ethernal: I've moved the instantiation of `commiteeQuerier` next to the `aggsender` creation https://github.com/agglayer/aggkit/pull/877/commits/80268b2d8956aa6b376701c8b795ebd9964da0d7

aggsender/aggsender.go
Stefan-Ethernal: Kept the channel, but simplified a bit, by introducing the struct that holds both signature and error. Let me know if you still think we should simplify it https://github.com/agglayer/aggkit/pull/877/commits/80268b2d8956aa6b376701c8b795ebd9964da0d7

aggsender/aggsender.go
Stefan-Ethernal: We are logging the validation `errors` on the WARNING lvl, IMO it should be enough, because we don't want to skip the certificate submission if we have a threshold of signatures, in the [sendCertificate](https://github.com/agglayer/aggkit/blob/80268b2d8956aa6b376701c8b795ebd9964da0d7/aggsender/aggsender.go#L382) func . Do you agree?

goran-ethernal: I see you are logging them in the go routine.

aggsender/aggsender.go
goran-ethernal: Ok, I think we are fine to log that it can't be reached, and it won't have a signature, so it doesn't matter if we return an error. Let's leave it like this.

cmd/run.go
```diff
@@ -113,6 +114,12 @@ func start(cliCtx *cli.Context) error {
 
 			go b.Start(cliCtx.Context)
 		case aggkitcommon.AGGSENDER:
+			committeeQuerier, err := query.NewECDSAMultisigCommitteeQuery(cfg.L1NetworkConfig.RollupAddr, l1Client)
```
goran-ethernal, line 118: Still not sure if this is going to be the `RollupAddr` or some new address. We need to check this with the protocol team.

aggsender/aggsender.go
```diff
@@ -197,16 +184,21 @@ func (a *AggSender) Start(ctx context.Context) {
...
-		if err != nil {
-			a.log.Warnf("error checking validator health: %v", err)
-		}
-		if !status.IsHealthy() {
-			a.log.Warnf("validator status is not healthy: %s", status.String())
-		}
-		a.log.Infof("validator health check: %s", status.String())
+
+	if !a.cfg.RequireValidatorCall {
+		a.localValidator = validator.NewLocalValidator(
```
goran-ethernal, line 191: Add a comment that this is only used for testing and that it should be removed after we test the multisig.

goran-ethernal (approved):

aggsender/aggsender.go
goran-ethernal: Yeah, I do not like to complicate things with channels. In this case, we can leave it, but please add this check as well to the result channel loop:

```go
for {
    val, ok := <-results
    if !ok {
        break // channel closed
    }
    // process val
}
```

Because, when the channel closes, it sends two things, a default zero value of the channel, and a boolean if it is closed or not. In current code, when the channel closes, we will receive a `sigResult{}` with nil error and nil signature, and we will mistakenly add a nil signature to the list.

cmd/run.go
Stefan-Ethernal: I'll add a TODO there as well.

aggsender/aggsender.go
Stefan-Ethernal: Fixed https://github.com/agglayer/aggkit/pull/877/commits/bbff0fe0370caf6c6ad6dc24c282fd56561f1d24

cmd/run.go
Stefan-Ethernal: Fixed https://github.com/agglayer/aggkit/pull/877/commits/bbff0fe0370caf6c6ad6dc24c282fd56561f1d24

aggsender/aggsender.go
Stefan-Ethernal: I have double checked, but it seems that it is functionally equivalent. This is the explanation, in a nutshell:

When you use range over a channel, Go will not give you an extra “fake” element — it will simply stop iteration the moment the channel is closed and the buffer is drained.

That means with your current code:
- You cannot accidentally append a zero signResult{} from the channel being closed unless you explicitly wrote it to the channel.
- range results already has the ok check built in.

PR #881, opened 2025-08-15
feat: support multiple `claims` by `global index`

### LINKED ISSUE #878: feat: support multiple claims with same global index when querying `L2 bridge syncer`
Because of `Remove GER` feature, there can be rare cases where two claims can have same `global index`. The `L2 bridge syncer` does not have constraints on its `db` in regards to `global index`, which is good, but we need to filter them by `bridge exit hash` as well, because `agglayer` endpoint for `last settled imported bridge exit` will return its `bridge exit hash` also.

The best way to go is just query all the `claims` from `L2 bridge syncer` `db` that have a certain `global index`, and just recalculate their `bridge exit hash` to see which one matches the one gotten from `agglayer`.

goran-ethernal:
## 🔄 Changes Summary
Because of `Remove Injected GER` feature, the `L2 Bridge Syncer` can have (in rare cases) two `claims` with the same `global index`, where one of them gets unclaimed.

Because of this, and since syncer's `db` has no constraints on `global index` field, the `aggsender-validator` needs to support this as well, when querying the `claim` based on `global index`.

The `agglayer` will return a `bridge exit hash` alongside `global index` for the last settled `imported bridge exit` (`claim`), which we can use to filter out the desired `claim`. 

The `aggsender-validator` will query all the `claims` from the syncer based on the last settled `global index` and then filter them out by their `bridge exit hash` to find a correct one.

## ⚠️ Breaking Changes
NA

## 📋 Config Updates
NA

## ✅ Testing

## 🐞 Issues
- Closes #878

### DIFF
--- a/aggsender/converters/imported_bridge_exit_converter_test.go
+++ b/aggsender/converters/imported_bridge_exit_converter_test.go
@@ -17,6 +17,25 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestBre(t *testing.T) {
+	ibe := &agglayertypes.ImportedBridgeExit{
+		GlobalIndex: &agglayertypes.GlobalIndex{MainnetFlag: true, RollupIndex: 0, LeafIndex: 1},
+		BridgeExit: &agglayertypes.BridgeExit{
+			LeafType: agglayertypes.LeafTypeAsset,
+			TokenInfo: &agglayertypes.TokenInfo{
+				OriginNetwork:      1,
+				OriginTokenAddress: common.HexToAddress("0x123"),
+			},
+			DestinationNetwork: 2,
+			DestinationAddress: common.HexToAddress("0x456"),
+			Amount:             big.NewInt(100),
+			Metadata:           crypto.Keccak256([]byte("metadata")),
+		},
+	}
+
+	fmt.Println(ibe.BridgeExit.Hash().String())
+}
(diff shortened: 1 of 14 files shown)


aggsender/converters/imported_bridge_exit_converter_test.go
```diff
@@ -17,6 +17,25 @@ import (
...
+			},
+			DestinationNetwork: 2,
+			DestinationAddress: common.HexToAddress("0x456"),
+			Amount:             big.NewInt(100),
+			Metadata:           crypto.Keccak256([]byte("metadata")),
+		},
+	}
+
+	fmt.Println(ibe.BridgeExit.Hash().String())
+}
```
Stefan-Ethernal: Remove? :rofl:
```suggestion
```

Stefan-Ethernal (approved):

aggsender/converters/imported_bridge_exit_converter_test.go
goran-ethernal: Indeed 🤣

goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/881/commits/40651625d56fdb1dbe90a2b75f3f3869d7d7c2eb

PR #898, opened 2025-08-21
feat: don't use `debug` endpoint on `aggsender-validator`, instead verify proofs on `claims`

### LINKED ISSUE #892: AggSender validator verifies MPs instead of using debug endpoint
In order to make Agssender Validator lighter, it couls verify the MerkleProofs received by the proposer instead of syncing them (and checking that they are the same as synced), which requires the debug endpoint

goran-ethernal:
## 🔄 Changes Summary
This PR removes the dependency on the `debug` endpoint on an L2 rpc on the `aggsender-validator`, meaning, when `aggsender-validator` is run by itself on a separate container, without `bridge`, `aggsender` and `aggchain-proof-gen` components, it will not use sync the full claims.

What the `aggsender-validator` will do when verifying the new certificate gotten from the `aggsender-proposer` is that, it will verify the proofs for each `claim` (`imported bridge exit`) provided by the proposer by using the calculated finalized `L1 info tree root` on the validator.

This decouples the validator from the debug endpoints, which, depending on the rpc provider, can be supported, or in some cases, even not provided. Plus, checking the proofs on each `claim` by using the `L1 info tree root` the validator provided, makes an addition security check of validity of a proposed ceritifacte.

## ⚠️ Breaking Changes
NA

## 📋 Config Updates
NA

## ✅ Testing
- 🤖 **Automatic**: `aggkti` CI

## 🐞 Issues
- Closes #892

### DIFF
--- a/agglayer/types/types.go
+++ b/agglayer/types/types.go
@@ -787,20 +787,20 @@ type Claim interface {
 	Validate() error
 }
 
-// ClaimFromMainnnet represents a claim originating from the mainnet
-type ClaimFromMainnnet struct {
+// ClaimFromMainnet represents a claim originating from the mainnet
+type ClaimFromMainnet struct {
 	ProofLeafMER     *MerkleProof    `json:"proof_leaf_mer"`
 	ProofGERToL1Root *MerkleProof    `json:"proof_ger_l1root"`
 	L1Leaf           *L1InfoTreeLeaf `json:"l1_leaf"`
 }
 
 // Type is the implementation of Claim interface
-func (c ClaimFromMainnnet) Type() string {
+func (c ClaimFromMainnet) Type() string {
 	return "Mainnet"
 }
 
 // MarshalJSON is the implementation of Claim interface
-func (c *ClaimFromMainnnet) MarshalJSON() ([]byte, error) {
+func (c *ClaimFromMainnet) MarshalJSON() ([]byte, error) {
 	return json.Marshal(&struct {
 		Child map[string]interface{} `json:"Mainnet"`
 	}{
@@ -812,7 +812,7 @@ func (c *ClaimFromMainnnet) MarshalJSON() ([]byte, error) {
 	})
 }
 
-func (c *ClaimFromMainnnet) UnmarshalJSON(data []byte) error {
+func (c *ClaimFromMainnet) UnmarshalJSON(data []byte) error {
 	if string(data) == nullStr {
 		return nil
 	}
@@ -835,7 +835,7 @@ func (c *ClaimFromMainnnet) UnmarshalJSON(data []byte) error {
 }
 
 // Hash is the implementation of Claim interface
-func (c *ClaimFromMainnnet) Hash() common.Hash {
+func (c *ClaimFromMainnet) Hash() common.Hash {
 	if c.Validate() != nil {
 		return common.Hash{}
 	}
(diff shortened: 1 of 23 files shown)


agglayer/types/types.go
```diff
@@ -579,9 +580,9 @@ func (b *BridgeExit) Hash() common.Hash {
 
 	return crypto.Keccak256Hash(
 		[]byte{b.LeafType.Uint8()},
-		aggkitcommon.Uint32ToBytes(b.TokenInfo.OriginNetwork),
+		aggkitcommon.Uint32ToBigEndianBytes(b.TokenInfo.OriginNetwork),
 		b.TokenInfo.OriginTokenAddress.Bytes(),
-		aggkitcommon.Uint32ToBytes(b.DestinationNetwork),
+		aggkitcommon.Uint32ToBigEndianBytes(b.DestinationNetwork),
```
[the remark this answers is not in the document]
goran-ethernal: It is not, the `UInt32ToBytes` converted it to big endian, but it was not clear from the name.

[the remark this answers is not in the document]
goran-ethernal: It is not breaking anything, the `UInt32ToBytes` converted it to `big endian`, but it was not clear from the name.

tree/tree.go
```diff
@@ -262,7 +262,7 @@ func CalculateRoot(leafHash common.Hash, proof [types.DefaultHeight]common.Hash,
 	node := leafHash
 
 	// Compute the Merkle root
-	for height := uint8(0); height < types.DefaultHeight; height++ {
+	for height := range types.DefaultHeight {
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/898/commits/8c4bf6e340dadf938f7863e4af19d0f0107c48e1

agglayer/types/types.go
```diff
@@ -863,6 +864,37 @@ func (c *ClaimFromMainnnet) Validate() error {
 	return nil
 }
 
+// VerifyProofs verifies the inclusion proofs for the given mainnet claim
+func (c *ClaimFromMainnnet) VerifyProofs(l1RootFromWhichToProof, leafHash common.Hash, leafIndex uint32) error {
```
rachit77:
```suggestion
func (c *ClaimFromMainnet) VerifyProofs(l1RootFromWhichToProof, leafHash common.Hash, leafIndex uint32) error {
```

goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/898/commits/29557c4afe8270497c30a08cc905738ca8c33959

agglayer/types/types.go
```diff
@@ -946,6 +978,43 @@ func (c *ClaimFromRollup) String() string {
 		c.ProofLeafLER.String(), c.ProofLERToRER.String(), c.ProofGERToL1Root.String(), c.L1Leaf.String())
 }
 
+func (c *ClaimFromRollup) VerifyProofs(
```
temaniarpit27, line 981: is it possible to extract common part from above added func?

temaniarpit27 (approved):

agglayer/types/types.go
goran-ethernal: Not really. It would mean adding another base struct, which just adds more complexity. The common part is only in the `ProofGERToL1Root`, so I would not complicate the code because of one common proof field.

PR #902, opened 2025-08-22
feat: filter logs recursively

### LINKED ISSUE #812: Failed to filter logs, because the RPC yields too many results
When indexing events in the `evmdownloader`, in particular the l1 info tree syncer, there are occasions that it is unable to index the events, because there are too many log entries for a specific block range yielded. This is in particular the case with the Tenderly RPC nodes.

So we should try to come up with some solution, maybe to narrow the block range, until it successfully yield the results.

```go
github.com/agglayer/aggkit/sync.(*EVMDownloaderImplementation).GetLogs
	/app/sync/evmdownloader.go:419
github.com/agglayer/aggkit/sync.(*EVMDownloaderImplementation).getEventsByBlockRangeWithRetry
	/app/sync/evmdownloader.go:335
github.com/agglayer/aggkit/sync.(*EVMDownloaderImplementation).GetEventsByBlockRange
	/app/sync/evmdownloader.go:324
github.com/agglayer/aggkit/sync.(*EVMDownloader).Download
	/app/sync/evmdownloader.go:169
```

```go
error calling FilterLogs to eth client: filter: FromBlock: 8820233, ToBlock: 8821233, Addresses: [0x2968D6d736178f8FE7393CC33C87f29D9C287e78 0xE2EF6215aDc132Df6913C8DD16487aBF118d1764], Topics: [] err:%!w(*rpc.jsonError=&{-32602 invalid params Query returned more than 20000 results. Try with this block range [0x869609, 0x869773].}) 
```

Image

rachit77:
## 🔄 Changes Summary
- This PR introduces change that makes `evmdownloader` fetch logs recursively using smaller block ranges when the original range contains too many logs.

## 🐞 Issues
- Closes #812

### DIFF
--- a/sync/evmdownloader.go
+++ b/sync/evmdownloader.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math/big"
 	"slices"
+	"strings"
 	"time"
 
 	"github.com/agglayer/aggkit/log"
@@ -325,6 +326,11 @@ func (d *EVMDownloaderImplementation) GetEventsByBlockRange(ctx context.Context,
 	return d.getEventsByBlockRangeWithRetry(ctx, fromBlock, toBlock, 0)
 }
 
+func (d *EVMDownloaderImplementation) GetLogs(ctx context.Context, fromBlock, toBlock uint64) []types.Log {
+	unfilteredLogs := d.getUnfilteredLogs(ctx, fromBlock, toBlock)
+	return d.filterLogs(unfilteredLogs)
+}
+
 func (d *EVMDownloaderImplementation) getEventsByBlockRangeWithRetry(
 	ctx context.Context,
 	fromBlock, toBlock uint64, retryCount int,
@@ -395,7 +401,7 @@ func filterQueryToString(query ethereum.FilterQuery) string {
 		query.FromBlock.String(), query.ToBlock.String(), query.Addresses, query.Topics)
 }
 
-func (d *EVMDownloaderImplementation) GetLogs(ctx context.Context, fromBlock, toBlock uint64) []types.Log {
+func (d *EVMDownloaderImplementation) getUnfilteredLogs(ctx context.Context, fromBlock, toBlock uint64) []types.Log {
 	var (
 		attempts       = 0
 		unfilteredLogs []types.Log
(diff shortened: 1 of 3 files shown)


sync/evmdownloader_test.go
```diff
@@ -595,3 +595,55 @@ func runSteps(t *testing.T, fromBlock uint64, steps []evmTestStep) {
...
+			Address:     contractAddr,
+			BlockNumber: 175,
+			Topics:      []common.Hash{eventSignature},
+			BlockHash:   common.HexToHash("0x456"),
+		},
+	}
+	mockEthClient.EXPECT().FilterLogs(ctx, mock.Anything).Return(secondHalfLogs, nil).Once()
+
+	result := sut.getUnfilteredLogs(ctx, fromBlock, toBlock)
+	fmt.Println("result", result)
```
temaniarpit27: dont need this

sync/evmdownloader.go
```diff
@@ -416,28 +422,49 @@ func (d *EVMDownloaderImplementation) GetLogs(ctx context.Context, fromBlock, to
...
 			}
 
+			errStr := err.Error()
+			if strings.Contains(errStr, "Query returned more than") && fromBlock < toBlock {
+				midBlock := (fromBlock + toBlock) / 2 //nolint:mnd
+				if midBlock >= fromBlock {
+					d.log.Warnf("too many results for block range [%d, %d], splitting into [%d, %d] and [%d, %d]",
+						fromBlock, toBlock, fromBlock, midBlock, midBlock+1, toBlock)
+
+					firstHalf := d.getUnfilteredLogs(ctx, fromBlock, midBlock)
```
goran-ethernal: Overall, looks nice. I only have once concern here. If we have huge block range, with a lot of logs, splitting this with recursion can make it go very deep where there is always a risk of stack overflow. But, I do not expect cases like this.

Maybe one possible approach would be lowering the batch size in a loop before calling the `rpc` node in case there is this error, something like this:

```go
initialBatchSize := toBlock - fromBlock + 1 // start with full range

var (
    results []types.Log
    batchSize = initialBatchSize
)


for start := fromBlock; start <= toBlock; {
  end := start + batchSize - 1
  if end > toBlock {
	  end = toBlock
  }
  
  for {
      logs, err = d.ethClient.FilterLogs(ctx, start, end)
      if err = nil {
         // append to results and break
      }
      
      if errors.Is(err, context.Canceled) {
			return nil
	  }

      if strings.Contains(err.Error(), "Query returned more than") {
           if batchSize == 1 {
				return nil, fmt.Errorf("too many logs even in single block %d", start)
		   }
		   
		   batchSize /= 2 // we lower the batchSize
		   d.log.Warnf("too many logs in range [%d,%d], reducing batch size to %d", start, end, batchSize)
           
           // rest of the logic
      }
}
```

This is just a suggestion, where we do not use recursion, and which also covers the case when one block has too many logs of interest (which should never happen, but still, you never know).

Like I said, I am fine with you solution as well. If you think we are good with what you provided, I will approve the PR. This comment is just a food for thought.

temaniarpit27 (approved):

sync/evmdownloader.go
rachit77: I’ll accommodate the changes you suggested. Although the likelihood of the original logic causing a stack overflow is very low, avoiding recursion is still a better approach.

rachit77: done: [bc8182a](https://github.com/agglayer/aggkit/pull/902/commits/bc8182ab95ed8814f6fbe40cc738e6281fbf8213)

goran-ethernal (approved):

PR #916, opened 2025-08-27
feat: confgure block finality based on blockReference and amount of blocks

### LINKED ISSUE #901: configure L2 block finality based on amount of blocks
Aggsender should be able to select block range based on `toBlock = last sequenced L2 block - X`

joanestebanr:
## 🔄 Changes Summary
- Finalty Block can be configured using a block reference + a offset.
- To set aggsender toBlock range must define the finality of  `BridgeL2Sync`: 
```
[BridgeL2Sync]
BlockFinality = "LatestBlock/-30"
```
- Disabled l2gersync E2E unittest because is broken: 
    -  The e2e test is flacky because it depend on time, adding this new feature increase the execution time and make it fails very ofter. Related issue about this point: #924 and #925

## ⚠️ Breaking Changes
- Removed from config `Aggsender.BlockFinality`: this field was a missleading because the finality of aggsender is defined by `BridgeL2Sync.BlockFinality`. It got the last block processed by l2bridgesyncer as a final block.

## 📋 Config Updates
- Removed from config `Aggsender.BlockFinality`
- The finality fields in config file can specify an offset: 
     - This change is retrocompatible
     - Apply to all fields related to finality: 
         - `BridgeL2Sync.BlockFinality`
         - `L2GERSync.BlockFinality`
         - `ReorgDetectorL2.FinalizedBlock`
     - These fields accept a suffix that is the offset to apply. e.g: 
```
FinalizedBlock = "LatestBlock/-10"
```
Another examples: 
-  `SafeBlock/10`
- `FinalizedBlock/-10`

## 🐞 Issues
- Closes #901


types/block_finality.go
```diff
@@ -67,24 +97,115 @@ func (BlockNumberFinality) JSONSchema() *jsonschema.Schema {
...
+			blockNumber += uint64(offset)
+		}
+	} else {
+		blockNumber += uint64(offset)
+	}
+	// Can't return a block number biggest than Latest, so Latest+10 is the same as Latest+0
+	if b == Latest {
+		return min(blockNumber, originalBlockNumber)
+	}
+	return blockNumber
```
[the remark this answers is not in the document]
joanestebanr: The idea is allow to blockRange have the full range of values of uint64, so the code is for dealing with that situation

types/block_finality.go
```diff
@@ -1,60 +1,92 @@
...
 	"github.com/invopop/jsonschema"
 )
 
+// BlockNumberFinality represents a block finality with an optional offset
 type BlockNumberFinality struct {
-	string `validate:"required"`
+	Block  BlockNumber
+	Offset int64
+	// IfNotFoundReturnsZeroFlag indicates whether to return zero if the block is not found
+	IfNotFoundReturnsZeroFlag bool
```
goran-ethernal: I don't see this being used in any of the components. It looks like its always false.

types/block_finality.go
```diff
@@ -1,60 +1,92 @@
...
 }
 
-func NewBlockNumberFinality(s string) BlockNumberFinality {
-	return BlockNumberFinality{s}
+// NewBlockNumberFinality creates a new BlockNumberFinality from a string
+// format: <blockName>[/<offset>] e.g: "SafeBlock", "FinalizedBlock/-5", "LatestBlock/+10"
+func NewBlockNumberFinality(s string) (BlockNumberFinality, error) {
+	result := BlockNumberFinality{}
+	splitted := strings.Split(s, blockNameAndOffsetSeparator)
+	if len(splitted) == 0 || len(splitted) > 2 {
```
goran-ethernal, line 27: Maybe move the length of split string to a constant. Something like `const maxBlockFinalityItems = 2`.

types/block_finality.go
```diff
@@ -1,60 +1,92 @@
...
 		return fmt.Errorf("failed to parse BlockNumberFinality %s: %w", string(data), err)
 	}
-	d.string = res.string
+	b.Block = res.Block
+	b.Offset = res.Offset
 	return nil
 }
 
+// JSONSchema returns the JSON schema for BlockNumberFinality
 func (BlockNumberFinality) JSONSchema() *jsonschema.Schema {
```
goran-ethernal, line 88: Should we update the `JSONSchem` with new examples like `Finalized/30`?

types/block_finality.go
```diff
@@ -67,24 +99,120 @@ func (BlockNumberFinality) JSONSchema() *jsonschema.Schema {
...
+func (b BlockNumber) ApplyOffset(blockNumber uint64, offset int64) uint64 {
+	originalBlockNumber := blockNumber
+	if offset < 0 {
+		if blockNumber < uint64(-offset) {
+			blockNumber = 0
+		} else {
+			blockNumber += uint64(offset)
+		}
+	} else {
+		blockNumber += uint64(offset)
```
goran-ethernal, line 182: Thinking out loud, what happens if someone configures: `Finalized/40`, and we have an example where:
- finalized block = 10
- latest block = 30

When we apply the offset, we will return a block that does not exist.

To me this offset only makes sense with the `Latest` and `Pending` blocks and it has to be negative.

goran-ethernal (approved):

types/block_finality.go
joanestebanr: you're right, I'm going to remove it!

types/block_finality.go
joanestebanr: I don't see the benefits, you need to check a constant for something that is really straightfoward

goran-ethernal: Its because of `linter` to not use the `nolint` directives. I agree with you, its straightforward. Maybe we can turn this rule off in the `linter`. Its anoying. But, we don't have to do that now.

types/block_finality.go
joanestebanr: You're right, it can be misconfigured... For `latest` the positive values are forbidden, for the rest you can configure whatever you want. This value are not intended to be modified without knowledge of what are you doing, so I hope that if somebody set `Finalized/40` understand that there are more that 40 block up to `latest`.. WDTY?

goran-ethernal: I agree with you, although, in my opinion, having offset with `Finalized` and `Safe` blocks doesn't bring any value, but let's leave it if someone wants to play around, although I think they only need this feature when an RPC node does not support the `Finalized` and `Safe` blocks.

goran-ethernal (approved):

types/block_finality.go
joanestebanr: But I don't see that the `linter` is   complaining about this line... but I agree that rule is a bit anyoing!  😆

rachit77 (approved):

PR #923, opened 2025-08-28
feat: add support for individual components

### LINKED ISSUE #880: [FEATURE]  Support running all syncers independently
Devtools team has cases where they would benefit running all syncers independently to the major components like `Aggsender` or `Bridge Service`, so we need to support this.

Basically, add a component for each syncer:
- `L1INFOTREESYNC`
- `L2BRIDGESYNC`
- `L1BRIDGESYNC`
- `L2GERSYNC`

rachit77:
## 🔄 Changes Summary
- This PR adds support for running following syncers in a standalone mode independently:
1. `L1INFOTREESYNC `
2. `L1BRIDGESYNC`
3. `L2BRIDGESYNC`
4. `L2GERSYNC`

## ✅ Testing
- 🖱️ **Manual**: Tested it Locally

## 🐞 Issues
- Closes #880


cmd/run.go
```diff
@@ -509,7 +509,9 @@ func runL1ClientIfNeeded(ctx context.Context,
 		aggkitcommon.AGGSENDERVALIDATOR,
 		aggkitcommon.BRIDGE,
 		aggkitcommon.L1INFOTREESYNC,
+		aggkitcommon.L2GERSYNC,
```
temaniarpit27, line 512: do we have some usecase for syncing L2gers?

temaniarpit27 (approved):

cmd/run.go
rachit77: We wanted to make all syncers independent so they can be used as needed. 
Thinking out loud, someone might want to spin up an additional service to track injected GERs, or in case of data corruption in the L2GER, they might only want to resync the L2GER syncer database.

Stefan-Ethernal (approved):

PR #926, opened 2025-09-01
feat: send `multisig` with certificate

goran-ethernal:
## 🔄 Changes Summary
This PR updates the `agglayer` `SubmitCertificate` API which now supports sending `multisig` with the certificate.

`multisig` is sent in the `AggchainData` field in the `Certificate` which supports four types:
- `AggchainDataProof` - in the case of `FEP` network with `multisig` turned off.
- `AggchainDataSignature` - in the case of `PP` network with `multisig` turned off.
- `AggchainDataMultisigWithProof` - in the case of `FEP` network with `multisig` turned on.
- `AggchainDataMultisig` - in the case of `PP` network with `multisig` turned on.

## ⚠️ Breaking Changes
NA

## 📋 Config Updates
NA

## ✅ Testing


aggsender/aggsender.go
```diff
@@ -337,31 +337,30 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certifi
 		return nil, fmt.Errorf("error building certificate: %w", err)
 	}
 
-	var (
-		validators          []types.CertificateValidateAndSigner
-		signaturesThreshold uint32
-	)
-
+	var multisig *agglayertypes.Multisig
 	if a.localValidator != nil {
```
joanestebanr, line 341: I known that this is not introduced in that PR but if there are a `localValidator` also must execute the `else` branch (`getValidators(ctx)`)

aggsender/flows/flow_pp.go
```diff
@@ -130,6 +130,23 @@ func (p *PPFlow) BuildCertificate(ctx context.Context,
 	return signedCert, nil
 }
 
+// UpdateAggchainData updates the AggchainData field in certificate with the multisig if needed
+func (p *PPFlow) UpdateAggchainData(
```
joanestebanr, line 134: why it returns `*agglayertypes.Certificate`? the param `cert` is always modified in all the implementation, and this function signature suggest that the `cert` param is not modified

aggsender/flows/flow_pp.go
```diff
@@ -130,6 +130,23 @@ func (p *PPFlow) BuildCertificate(ctx context.Context,
...
+	cert *agglayertypes.Certificate,
+	multisig *agglayertypes.Multisig) (*agglayertypes.Certificate, error) {
+	if multisig == nil {
+		// multisig not turned on, we don't need to update the certificate
```
joanestebanr, line 138: It's possible this situation? when is going to be `nil` multisig?

goran-ethernal: Yes, when we do not have  multisig turned on (`RequireCallValidator` is `false` in the config).

aggsender/flows/flow_pp.go
goran-ethernal: I will remove the return of the `*agglayertypes.Certificate` since that certificate is directly modified, and just return an error.

aggsender/aggsender.go
goran-ethernal: Local validator is only instantiated when `RequireCallValidator` is `false`, meaning, `multisig` is turned off. In that case, we instantiate it straight away in the `Start` function, and there is no need to call `getValidators`, since local validator is not a committee. `getValidators` reads the `multisig committee` from the contract.

agglayer/types/types.go
```diff
@@ -160,6 +160,10 @@ func (a *AggchainDataSelector) UnmarshalJSON(data []byte) error {
 		a.obj = &AggchainDataProof{}
 	} else if _, ok = obj["signature"]; ok {
 		a.obj = &AggchainDataSignature{}
+	} else if _, ok = obj["aggchain_proof"]; ok {
```
joanestebanr: I think that `aggchain_proof` correspond to `AggchainDataProof`  and `aggchain_proof_with_multisig` to `AggchainDataMultisigWithProof` ([proto ref](https://buf.build/agglayer/interop/docs/main:agglayer.interop.types.v1))

aggsender/aggsender.go
joanestebanr: The is going to be `nil`, but setting it must not prevent from running the remote validators

aggsender/aggsender.go
```diff
@@ -337,31 +337,30 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certifi
...
+			a.log.Warnf("local validation of certificate failed: %v. Cert: %s", err, certificate.Brief())
+		}
 	} else {
-		validators, signaturesThreshold, err = a.getValidators(ctx)
+		validators, signaturesThreshold, err := a.getValidators(ctx)
 		if err != nil {
 			return nil, fmt.Errorf("failed to get validators: %w", err)
 		}
+
+		multisig, err = a.pollValidatorCommittee(ctx, validators, signaturesThreshold,
```
joanestebanr, line 352: `pollValidatorCommittee` must check if some of the member of the comittee is himself? WDTY?

aggsender/flows/flow_pp.go
joanestebanr: I'm not able to find this param, where is?

goran-ethernal: Its `RequireValidatorCall`, sorry. We talked about this on slack, so we came to an agreement.

aggsender/aggsender.go
goran-ethernal: Its `RequireValidatorCall`, sorry. We talked about this on slack, so we came to an agreement.

aggsender/flows/flow_pp.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/926/commits/852bfd48c654015da4bce91af8ab8a5338cbe9b4

aggsender/aggsender.go
goran-ethernal: I don't see the point of running the local validator when we have multisig setup.

agglayer/types/types.go
goran-ethernal: This is not really one on one with agglayer, we are just assuming how the `json` format of the certificate looks like for the FEP certificates. Since we moved to `gRPC`, this is for sure not the same as `agglayer` for `json` format, but I think, we do not need to worry about how it looks like, since `agglayer` also moved to `gRPC`.

agglayer/types/types_test.go
```diff
@@ -1139,7 +1139,7 @@ func TestCertificate_ID(t *testing.T) {
 	require.Equal(t, "cert{height:2, networkID:1}", cert.ID())
 }
 
-func TestAggchainDataSignature_MarshalJUnmarshalJSON(t *testing.T) {
+func TestAggchainDataSignature_MarshalUnmarshalJSON(t *testing.T) {
```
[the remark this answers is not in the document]
goran-ethernal: Its already fixed.

aggsender/aggsender.go
```diff
@@ -378,11 +376,7 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certifi
 	}
 
 	// TODO: Update once agglayer endpoint changes to accept the multisig
```
Stefan-Ethernal: Remove TODO, it is obsoleted?

```suggestion
```

aggsender/aggsender.go
Stefan-Ethernal: If my understading is correct, we should remove the local validator implementation any time soon? It was just a way to test the validation of certificate, before we have the proper support for integration testing...?

agglayer/types/types.go
Stefan-Ethernal: Sorry if this is a noob question, but in this [definition](https://buf.build/agglayer/interop/docs/main:agglayer.interop.types.v1#agglayer.interop.types.v1.AggchainData), we don't have the field named `aggchain_proof`, but there is one called `generic`, should it be updated here?

Am I looking to the correct proto ref?

aggsender/aggsender.go
```diff
@@ -433,14 +427,14 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certifi
 	return certificate, nil
 }
 
-// pollValidators committee signers to validate the certificate and gather signatures
-func (a *AggSender) pollValidators(
+// pollValidatorCommittee calls all validator committee members to validate and sign the certificate
```
Stefan-Ethernal: WDYT?
```suggestion
// pollValidatorCommittee invokes validator committee members to validate and sign the certificate
```

aggsender/aggsender.go
```diff
@@ -526,26 +522,30 @@ func (a *AggSender) pollValidators(
...
 		}
 	}
 
-	if uint32(len(signatures)) < signaturesThreshold {
+	if uint32(len(multisig.Signatures)) < signaturesThreshold {
 		metrics.MultiSigThresholdNotReached()
 
 		if len(errs) > 0 {
-			return signatures, errors.Join(errs...)
+			return multisig, errors.Join(errs...)
```
Stefan-Ethernal: I'm having some second thoughts here, should we return a nil for the multisig?
```suggestion
			return nil, errors.Join(errs...)
```

Stefan-Ethernal (approved):
Generally LGTM, but leaving some comments to consider.

agglayer/types/types.go
goran-ethernal: I can not explain this easily, but they do not have json definition of certificates anymore. This marshal, unmarshal we have is no longer one on one with agglayer, and it is very different. We use this so we can save the certificate to a `.json` file on file system.

I will try to update this to be as close as it can.

aggsender/aggsender.go
goran-ethernal: Yes, you are right.

aggsender/aggsender.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/926/commits/8ba71cbfe84adf2ef596d7080f2893b4c9a1a564

aggsender/aggsender.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/926/commits/8ba71cbfe84adf2ef596d7080f2893b4c9a1a564

aggsender/aggsender.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/926/commits/8ba71cbfe84adf2ef596d7080f2893b4c9a1a564

agglayer/types/types.go
goran-ethernal: I've tried to make it more precise and similar to the `grpc` struct marshaling here https://github.com/agglayer/aggkit/pull/926/commits/2e6311be7156aa51cc21f9ada3e42dd4616f74ee, but like we discussed, this should be removed in `v0.8.0`.

aggsender/aggsender.go
```diff
@@ -433,14 +426,14 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certifi
 	return certificate, nil
 }
 
-// pollValidators committee signers to validate the certificate and gather signatures
-func (a *AggSender) pollValidators(
+// pollValidatorCommittee invokes validator committee members to validate and sign the certificate
+func (a *AggSender) pollValidatorCommittee(
```
joanestebanr, line 430: THis function is too big 109 LoC, may be make sense to split it, also is missing the aggsender-proposer signature. 
maybe could be moved and splitted into another file? WDTY?

aggsender/aggsender.go
```diff
@@ -475,21 +468,21 @@ func (a *AggSender) pollValidators(
 
 			status, err := v.HealthCheck(ctx)
```
joanestebanr: I don't see any reason to do this call... calling `ValidateAndSignCertificate` must be enough, WDTY?

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -333,6 +333,30 @@ func (a *AggchainProverFlow) BuildCertificate(ctx context.Context,
...
+func (a *AggchainProverFlow) UpdateAggchainData(
+	cert *agglayertypes.Certificate,
+	multisig *agglayertypes.Multisig,
+) error {
+	if multisig == nil {
+		// multisig not turned on, we don't need to update the certificate
+		return nil
+	}
+
+	proof, ok := cert.AggchainData.(*agglayertypes.AggchainDataProof)
```
joanestebanr, line 346: I would check if it's `AggchainDataMultisigWithProof` , because may be it's already updated, WDTY?

aggsender/validator/compare_certs.go
```diff
@@ -80,3 +84,18 @@ func DiffsBridgeExits(
...
 	return diffs
 }
+
+// DiffsImportedBridgeExits compares two slices of ImportedBridgeExit and returns a slice of strings
+// containing the differences between them.
+func DiffsImportedBridgeExits(
+	expected []*agglayertypes.ImportedBridgeExit,
+	validating []*agglayertypes.ImportedBridgeExit) []string {
+	diffs := make([]string, 0)
+	if len(expected) != len(validating) {
```
joanestebanr, line 94: are there other data that can we check? just the length?

aggsender/aggsender.go
goran-ethernal: I was thinking to do this through this issue: https://github.com/agglayer/aggkit/issues/938 which will also check the proposer signature. I don't want to include too much changes in this PR which is already too big.

WDYT?

aggsender/validator/compare_certs.go
goran-ethernal: I think this is enough, because what we sync in non full claims is not really clearly determined, and if it is, its determined in another package, so we can easily forget something.

aggsender/aggsender.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/926/commits/2db88f3fec06e034436f9c5a21b929d67a0c5891

aggsender/flows/flow_aggchain_prover.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/926/commits/2db88f3fec06e034436f9c5a21b929d67a0c5891

aggsender/aggsender.go
joanestebanr: ok, I have added to the issue the refactor required

aggsender/validator/compare_certs.go
goran-ethernal: I've added a check for global index as well: https://github.com/agglayer/aggkit/pull/926/commits/fa4a13953e3ddcfddf8a5cdc555b0139299f7fd3

The rest of the fields we can not check because they are not synced when we are not using the `debug` endpoint.

joanestebanr (approved):

Issue #924, opened 2025-08-28
`L2GERSync` misses to index certain events


rachit77:
Currently, `L2GERSync` skips intermediate blocks and only indexes the latest block in sovereign mode. This behavior needs to be updated so it matches legacy mode, where all blocks are processed and stored in the syncer database.

Also as of now we don't store empty processed block in db in case of sovereign mode which is again inconsistent with legacy mode. 

In legacy mode, however, there is an edge case during setup: when the restart block number is set to `last processed block + 1`, that block gets skipped and is never processed.

https://github.com/agglayer/aggkit/blob/2caf95bfe00c4a9cc5b536b481121bb143520835/sync/evmdownloader.go#L318-L320

PR #927, opened 2025-09-01
fix: `L2GERSync` syncer events indexing

rachit77:
## 🔄 Changes Summary
- This PR fixes an issue where `L2GERSync` was indexing only the latest block events in sovereign mode, causing some `insertGER` and `removeGER` events to be missed. This change ensures the syncer indexes all insertion and deletion events for GER and not just one corresponding to latest block.
- This PR also introduces the change which make the Aggkit node to stop in case synced L2 GER is not found in `L1infotreeSync` in sovereign mode.

## 📋 Config Updates
- 🧾 Added `SyncBlockChunkSize` in `[L2GERSync]`

## 🐞 Issues
- #924 
- #935


goran-ethernal (approved):

temaniarpit27 (approved):

l2gersync/e2e_test.go
```diff
@@ -43,11 +43,12 @@ func TestL2GERSyncE2E(t *testing.T) {
 		l2Setup.SimBackend.Client(),
 		l2Setup.GERAddr,
 		l1Setup.InfoTreeSync,
+		syncBlockChunkSize,
 		retryAfterErrorPeriod,
 		maxRetryAttemptsAfterError,
 		aggkittypes.LatestBlock,
 		waitForNewBlocksPeriod,
-		syncBlockChunkSize,
+		10, // downloadBufferSize
```
Stefan-Ethernal, line 51: WDYT, can we reuse the `syncBlockChunkSize` even for the `downloadBufferSize`? IMO it would be nice if we can reduce the number of parameters for the constructor function...

l2gersync/evm_downloader_sovereign_test.go
```diff
@@ -0,0 +1,113 @@
...
+			Data:        []byte{},
+			BlockNumber: fromBlock,
+			TxHash:      common.HexToHash("0x111"),
+			TxIndex:     0,
+			BlockHash:   testBlockHash,
+			Index:       0,
+		},
+	}
+
+	mockL2Client.On("ChainID", mock.Anything).Return(big.NewInt(1), nil).Maybe()
```
Stefan-Ethernal: The approach with using expected is better, because it would produce a compile time error, in case function gets renamed, removed etc.

I would suggest changing it on the other places within this test as well.

```suggestion
	mockL2Client.EXPECT().ChainID(mock.Anything).Return(big.NewInt(1), nil).Maybe()
```

l2gersync/evm_downloader_sovereign.go
```diff
@@ -80,16 +83,24 @@ func (d *downloaderSovereign) Download(ctx context.Context, fromBlock uint64, do
...
 
-		// Wait for new blocks before processing
-		fromBlock = d.WaitForNewBlocks(ctx, fromBlock)
-		for _, block := range d.GetEventsByBlockRange(ctx, fromBlock, fromBlock) {
+		// Wait for new blocks and get current head
+		latestBlock := d.WaitForNewBlocks(ctx, fromBlock)
+		toBlock := fromBlock + d.syncBlockChunkSize - 1
+		if toBlock > latestBlock {
+			toBlock = latestBlock
+		}
```
Stefan-Ethernal: This is a recommendation from the IDE, to simplify this part, like this
```suggestion
		toBlock := min(fromBlock + d.syncBlockChunkSize - 1, latestBlock)
```

l2gersync/evm_downloader_sovereign.go
```diff
@@ -125,7 +136,7 @@ func (d *downloaderSovereign) buildAppender(
 
 		l1InfoTreeLeaf, err := d.l1InfoTreeSync.GetInfoByGlobalExitRoot(insertGEREvent.NewGlobalExitRoot)
 		if err != nil {
-			return fmt.Errorf("failed to fetch l1 info tree for global exit root %s: %w",
+			log.Fatalf("CRITICAL: GER %s received from L2 is not present in L1InfoTreeSync: %v",
```
Stefan-Ethernal: I'd consider removing the CRITICAL prefix, as it will be obvious, since this log message will be accompanied by the panic anyway. WDYT?

```suggestion
			log.Fatalf("GER %s received from L2 is not present in L1InfoTreeSync: %v",
```

Stefan-Ethernal (approved):

l2gersync/evm_downloader_sovereign.go
rachit77: I’d like to keep this prefix, since other teams monitoring the log in production may not fully grasp the severity of the situation and might attempt to restart aggkit although it will fail again after a brief amount of time.
wdyt?

l2gersync/evm_downloader_sovereign.go
rachit77: nice one

l2gersync/e2e_test.go
rachit77: I’d prefer to keep it separate for clarity. Also, based on the current implementation, syncBlockChunkSize should ideally be greater than downloadBufferSize, since only blocks containing events related to GER are sent to the channel.

l2gersync/evm_downloader_sovereign.go
Stefan-Ethernal: Don't have the strong opinion, but just for the sake of consistency, it makes sense for each `Fatalf` invocation to either have or not the prefix.

rachit77: In that case i will remove the prefix

PR #944, opened 2025-09-04
feat: add config example

### LINKED ISSUE #941: Create `config.toml.example`
Create `config.toml.example` to be used by infra teams and operators as an example, and keep it updated. Probably we could split the defaults.go into this new file and load it from there

temaniarpit27:
## 🔄 Changes Summary
- add a config.toml.example for validators and infra operators

## 📋 Config Updates
- 🧾 **Diff/Config snippet**: Renamed `SequencerPrivateKeyPath` to `AggSenderPrivateKeyPath`, renamed `SequencerPrivateKeyPassword` to `AggSenderPrivateKeyPassword`

## 🐞 Issues
- Closes #941


config.toml.example
```diff
@@ -0,0 +1,633 @@
...
+L2URL = "<URL>"
+
+AggLayerURL = "<URL>"
+
+AggchainProofURL= "<URL>"
+
+NetworkID = <NETWORK_ID>
+
+SequencerPrivateKeyPath = "<PATH>"
+SequencerPrivateKeyPassword  = "<PASSWORD>"
```
Stefan-Ethernal: As a side note, unrelated to this PR, but probably it would make sense to rename this on some occasion

```suggestion
AggSenderPrivateKeyPath = "<PATH>"
AggSenderPrivateKeyPassword  = "<PASSWORD>"
```

config.toml.example
```diff
@@ -0,0 +1,633 @@
...
+# the AggSender should send the certificate
+# 0 -> Begin
+# 50 -> Middle
+# ------------------------------------------------------------------------------
+# EpochNotificationPercentage = 50
+
+# ------------------------------------------------------------------------------
+# AggSender will save the certificates to a file in this path
+# ------------------------------------------------------------------------------
+SaveCertificatesToFilesPath = "<PATH>"
```
Stefan-Ethernal: It seems like this config parameter is obsoleted in a meantime, as this is the only leftover [occassion](https://github.com/agglayer/aggkit/blob/47e3ab20368dedf466d5b8cebc2ba8aca8df3842/test/config/pp-config.toml.template#L203).

config.toml.example
```diff
@@ -0,0 +1,633 @@
...
+BlockFinality = "LatestBlock"
+
+# ==============================================================================
+#                        _           _                              __
+#  __ _  __ _  __ _  ___| |__   __ _(_)_ __  _ __  _ __ ___   ___  / _| __ _  ___ _ __
+# / _` |/ _` |/ _` |/ __| '_ \ / _` | | '_ \| '_ \| '__/ _ \ / _ \| |_ / _` |/ _ \ '_ \
+#| (_| | (_| | (_| | (__| | | | (_| | | | | | |_) | | | (_) | (_) |  _| (_| |  __/ | | |
+# \__,_|\__, |\__, |\___|_| |_|\__,_|_|_| |_| .__/|_|  \___/ \___/|_|  \__, |\___|_| |_|
+#       |___/ |___/                         |_|                        |___/
+# ------------------------------------------------------------------------------
```
Stefan-Ethernal: Optional nitpick (worth considering changing it in the Kurtosis CDK, although not important at all, due to my OCD I'm noticing such things :slightly_smiling_face:)

```suggestion
# ==============================================================================
#     _                ____ _           _       ____                   __  ____            
#    / \   __ _  __ _ / ___| |__   __ _(_)_ __ |  _ \ _ __ ___   ___  / _|/ ___| ___ _ __  
#   / _ \ / _` |/ _` | |   | '_ \ / _` | | '_ \| |_) | '__/ _ \ / _ \| |_| |  _ / _ \ '_ \ 
#  / ___ \ (_| | (_| | |___| | | | (_| | | | | |  __/| | | (_) | (_) |  _| |_| |  __/ | | |
# /_/   \_\__, |\__, |\____|_| |_|\__,_|_|_| |_|_|   |_|  \___/ \___/|_|  \____|\___|_| |_|
#         |___/ |___/                                                                                  
# ------------------------------------------------------------------------------
```

Stefan-Ethernal (approved):
Looking good! :100:

Stefan-Ethernal:
@temaniarpit27 double check if it is necessary to update the Kurtosis CDK, because of `SequencerPrivateKey*` renames.

As a side note, it would be interesting if we can provide a way to generate/update the toml example based on the `default.go` and a template, but probably through a separate PR. WDYT?

temaniarpit27:
> @temaniarpit27 double check if it is necessary to update the Kurtosis CDK, because of `SequencerPrivateKey*` renames.
> 
> As a side note, it would be interesting if we can provide a way to generate/update the toml example based on the `default.go` and a template, but probably through a separate PR. WDYT?

Already done that.
We need to first finalize what params we want people to configure and what we want to keep abstracted but provide in case of emergency. If we provide everything the same as default.go then it would be confusing for people. I have currently provided whatever is there in kurtosis-cdk

config.toml.example
```diff
@@ -0,0 +1,628 @@
...
+
+AggSenderPrivateKeyPath = "<PATH>"
+AggSenderPrivateKeyPassword  = "<PASSWORD>"
+
+polygonBridgeAddr = "<ADDR>"
+
+# ------------------------------------------------------------------------------
+# URL is the L1 network url used in case of OP networks
+# ------------------------------------------------------------------------------
+# RPCURL = <URL>
```
joanestebanr: This var doesn't exists

config.toml.example
```diff
@@ -0,0 +1,628 @@
...
+[L1Config]
+# ------------------------------------------------------------------------------
+# URL is the L1 network url
+# ------------------------------------------------------------------------------
+URL = "<URL>"
+
+# ------------------------------------------------------------------------------
+# L1 chain id
+# ------------------------------------------------------------------------------
+chainId = "<CHAIN_ID>"
```
joanestebanr: Don't need to be configure, can be 0 and is requests to RPC

config.toml.example
```diff
@@ -0,0 +1,628 @@
...
+# ------------------------------------------------------------------------------
+polygonRollupManagerAddress = "<ADDR>"
+
+# ------------------------------------------------------------------------------
+# Address of the pol token address on L1
+# ------------------------------------------------------------------------------
+polTokenAddress = "<ADDR>"
+
+# ------------------------------------------------------------------------------
+# Address of the sovereign rollup contract on L2
```
joanestebanr: are you sure that is in L2? maybe L1 and L2? WDTY?

config.toml.example
```diff
@@ -0,0 +1,628 @@
...
+# https://github.com/uber-go/zap/blob/a55bdc32f526699c3b4cc51a2cc97e944d02fbbf/config.go#L161
+# ------------------------------------------------------------------------------
+Environment = "production" # "production" or "development"
+
+# ------------------------------------------------------------------------------
+# Level determines the log level that will be written to the
+# log. Generally we'll switch to debug if we want to troubleshoot
+# something specifically otherwise we leave it at info
+# ------------------------------------------------------------------------------
+# Level = "info"
```
joanestebanr: why you don't define `Level` but you do define  `Outputs`? in both cases are you setting the default values

joanestebanr (commented):
I think that the `config.toml.example` require to be reviewed and check the default values to avoid misleading, also there are template conditional like ` {{- if .use_agg_sender_validator }}` that I don't known if that is expected. 
Maybe the values in the config file must be real (`Port = "<PORT>"`) it's not better to set `Port = 1234`, because `Port` is not an string is numerical

config.toml.example
temaniarpit27: https://github.com/0xPolygon/kurtosis-cdk/blob/c93b16b8c083ba5b88c17b1ca4e914231c237d8d/templates/aggkit/aggkit-config.toml#L88

Thats what it says here

config.toml.example
temaniarpit27: it exists in kurtosis-cdk
https://github.com/0xPolygon/kurtosis-cdk/blob/c93b16b8c083ba5b88c17b1ca4e914231c237d8d/templates/aggkit/aggkit-config.toml#L44
but doesnt exist in our repo. will remove

config.toml.example
temaniarpit27: https://github.com/agglayer/aggkit/blob/a8ff6c39eede8892abb712c07a7b2b47070a89cb/config/default.go#L30
https://github.com/0xPolygon/kurtosis-cdk/blob/c93b16b8c083ba5b88c17b1ca4e914231c237d8d/templates/aggkit/aggkit-config.toml#L70
exists here so i kept it configurable
should i remove it?

config.toml.example
temaniarpit27: can you explain more?
https://github.com/0xPolygon/kurtosis-cdk/blob/c93b16b8c083ba5b88c17b1ca4e914231c237d8d/templates/aggkit/aggkit-config.toml#L133
I picked up from here

temaniarpit27:
> I think that the `config.toml.example` require to be reviewed and check the default values to avoid misleading, also there are template conditional like ` {{- if .use_agg_sender_validator }}` that I don't known if that is expected. Maybe the values in the config file must be real (`Port = "<PORT>"`) it's not better to set `Port = 1234`, because `Port` is not an string is numerical

I have picked the file from kurtosis-cdk. for ports, i see that it is configured to be string in kurtosis-cdk. https://github.com/0xPolygon/kurtosis-cdk/blob/c93b16b8c083ba5b88c17b1ca4e914231c237d8d/templates/aggkit/aggkit-config.toml#L160. i have put something like <> to make sure people configure them.

config.toml.example
joanestebanr: Check it, but I think that the best value is `0` and so be removed from configuration

config.toml.example
joanestebanr: I suggest to check the aggkit code to be sure

config.toml.example
joanestebanr: The default for `Log` section is: 
```
[Log]
Environment = "development" # "production" or "development"
Level = "info"
Outputs = ["stderr"]
```
So defining `Outputs` with same values is redundant

config.toml.example
joanestebanr: I suggest to check it, but I think that the best value is `0` and so be removed from configuration

joanestebanr:
> > I think that the `config.toml.example` require to be reviewed and check the default values to avoid misleading, also there are template conditional like ` {{- if .use_agg_sender_validator }}` that I don't known if that is expected. Maybe the values in the config file must be real (`Port = "<PORT>"`) it's not better to set `Port = 1234`, because `Port` is not an string is numerical
> 
> I have picked the file from kurtosis-cdk. for ports, i see that it is configured to be string in kurtosis-cdk. https://github.com/0xPolygon/kurtosis-cdk/blob/c93b16b8c083ba5b88c17b1ca4e914231c237d8d/templates/aggkit/aggkit-config.toml#L160. i have put something like <> to make sure people configure them.

I suggest to use `aggkit` as source of true instead of `kurtosis-cdk`, I'm not sure that having some templating in a example config file is a good idea. Maybe you can create a example configfile for `aggsender`, another for `bridge` and so on... WDTY?

temaniarpit27:
> > > I think that the `config.toml.example` require to be reviewed and check the default values to avoid misleading, also there are template conditional like ` {{- if .use_agg_sender_validator }}` that I don't known if that is expected. Maybe the values in the config file must be real (`Port = "<PORT>"`) it's not better to set `Port = 1234`, because `Port` is not an string is numerical
> > 
> > 
> > I have picked the file from kurtosis-cdk. for ports, i see that it is configured to be string in kurtosis-cdk. https://github.com/0xPolygon/kurtosis-cdk/blob/c93b16b8c083ba5b88c17b1ca4e914231c237d8d/templates/aggkit/aggkit-config.toml#L160. i have put something like <> to make sure people configure them.
> 
> I suggest to use `aggkit` as source of true instead of `kurtosis-cdk`, I'm not sure that having some templating in a example config file is a good idea. Maybe you can create a example configfile for `aggsender`, another for `bridge` and so on... WDTY?

I have removed templating from the PR.
I am not sure if creating different config file for each component is a good idea. It will only confuse deployers as there are config parts which will be shared. We can in future think of this but at this point of time it would make more sense to keep the config example similar to default.go

config.toml.example
temaniarpit27: removed it

config.toml.example
temaniarpit27: removed it

config.toml.example
temaniarpit27: it should be L1
checked code https://github.com/agglayer/aggkit/blob/18729e2c3eae838fc4d6593879de5d76150c5db6/config/network.go#L33

config.toml.example
```diff
@@ -0,0 +1,189 @@
...
+
+polygonBridgeAddr = "0x0000000000000000000000000000000000000000"
+
+# This values can be override directly from genesis.json
+rollupCreationBlockNumber = 0
+rollupManagerCreationBlockNumber = 0
+genesisBlockNumber = 0
+
+# Default database query timeout
+defaultDBQueryTimeout = "30s"
```
joanestebanr: I think that this parameter must be moved from `DefaultMandatoryVars` to `DefaultVars` because it's not mandatory to set it, and must be removed from here

config.toml.example
```diff
@@ -0,0 +1,189 @@
...
+	polygonRollupManagerAddress = "0x0000000000000000000000000000000000000000"
+    # ------------------------------------------------------------------------------
+    # Address of the pol token address on L1
+    # ------------------------------------------------------------------------------
+	polTokenAddress = "0x0000000000000000000000000000000000000000"
+    # ------------------------------------------------------------------------------
+    # Address of the sovereign rollup contract on L1
+    # ------------------------------------------------------------------------------
+	polygonZkEVMAddress = "0x0000000000000000000000000000000000000000"
+	AggchainFEPAddr = "0x0000000000000000000000000000000000000000"
```
joanestebanr: Maybe it require a comment as the rest of fields

config.toml.example
```diff
@@ -0,0 +1,189 @@
...
+
+# Default database query timeout
+defaultDBQueryTimeout = "30s"
+
+[L1Config]
+    # ------------------------------------------------------------------------------
+    # URL is the L1 network url
+    # ------------------------------------------------------------------------------
+	URL = "{{L1URL}}"
+	chainId = 0
```
joanestebanr: Maybe it require a comment as the rest of fields, also I suggest to check in the code if 0 means auto-discover

config.toml.example
```diff
@@ -0,0 +1,189 @@
...
+# ==============================================================================
+# The below configs are the default parameters to be used.
+# The user can modify it based on requirements
+# ==============================================================================
+
+# PathRWData = "/tmp/aggkit"
+# RequireStorageContentCompatibility = true
+# GenerateAggchainProofTimeout = "1h"
+# [L2RPC]
+# 	Mode = "basic"
```
joanestebanr, line 85: comments possible values

config.toml.example
```diff
@@ -0,0 +1,189 @@
...
+# 	URL = "{{L2URL}}"
+# 	RetryMode = "backoff"
+# 	MaxRetries = 5
+# 	InitialBackoff = "2s"
+# 	MaxBackoff = "10s"
+# 	BackoffMultiplier = 2.0
+#
+# [Log]
+# Environment = "production"
+# Level = "info"
```
joanestebanr, line 101: comments possible values for `Environment` and `Level`

config.toml.example
```diff
@@ -0,0 +1,189 @@
...
+#
+# [ReorgDetectorL2]
+# DBPath = "{{PathRWData}}/reorgdetectorl2.sqlite"
+# FinalizedBlock = "LatestBlock"
```
joanestebanr: I don’t think we should encourage changing any of the values in `ReorgDetectorL2`

config.toml.example
```diff
@@ -0,0 +1,189 @@
...
+#
+# [L1InfoTreeSync]
+# DBPath = "{{PathRWData}}/L1InfoTreeSync.sqlite"
```
joanestebanr: I don’t think we should encourage changing `DBPath` but the field `SyncBlockChunkSize` maybe make sense to be configured

config.toml.example
```diff
@@ -0,0 +1,189 @@
...
+#
+# [AggOracle]
+# ------------------------------------------------------------------------------
+# Set to true if enabling aggoracle committee mode
+# ------------------------------------------------------------------------------
+# EnableAggOracleCommittee = false
+# 		[AggOracle.EVMSender.EthTxManager]
+# 				PrivateKeys = [
```
joanestebanr: This field must be defined, isn't?

config.toml.example
```diff
@@ -0,0 +1,189 @@
...
+# 					{Method =  "local", Path = "/app/keystore/aggoracle.keystore", Password = "testonly"},
+# 				]
+# 				StoragePath = "{{PathRWData}}/ethtxmanager-aggoracle.sqlite"
```
joanestebanr: I don’t think we should encourage changing `StoragePath`

config.toml.example
```diff
@@ -0,0 +1,189 @@
...
+# 					[AggOracle.EVMSender.EthTxManager.Etherman]
+# 						# This field should be populated with L2ChainID
+# 						L1ChainID = 0
```
joanestebanr: I don’t think we should encourage changing ` L1ChainID`

config.toml.example
```diff
@@ -0,0 +1,189 @@
...
+# [RPC]
+# Host = "0.0.0.0"
+# Port = 5576
+#
+# [REST]
+# Host = "0.0.0.0"
+# Port = 5577
+#
+# [BridgeL1Sync]
+# DBPath = "{{PathRWData}}/bridgel1sync.sqlite"
```
joanestebanr: I don’t think we should encourage changing  `DBPath`

config.toml.example
```diff
@@ -0,0 +1,189 @@
...
+#
+# [BridgeL2Sync]
+# DBPath = "{{PathRWData}}/bridgel2sync.sqlite"
+# BlockFinality = "LatestBlock"
```
joanestebanr: I don’t think we should encourage changing any field of  `BridgeL2Sync

config.toml.example
```diff
@@ -0,0 +1,189 @@
...
+#
+# [L2GERSync]
+# DBPath = "{{PathRWData}}/l2gersync.sqlite"
+# BlockFinality = "LatestBlock"
```
joanestebanr: I don’t think we should encourage changing any field of  `L2GERSync`

config.toml.example
```diff
@@ -0,0 +1,189 @@
...
+#
+# [AggSender]
+# StoragePath = "{{PathRWData}}/aggsender.sqlite"
+# AggsenderPrivateKey = {{AggsenderPrivateKey}}
```
joanestebanr: I don’t think we should encourage changing ` StoragePath` and is refered at the begining so it can be removed also

config.toml.example
```diff
@@ -0,0 +1,189 @@
...
+# ------------------------------------------------------------------------------
+# Set to "AggchainProof" when enabling FEP
+# ------------------------------------------------------------------------------
+# Mode = "PessimisticProof"
```
joanestebanr: This field `Mode` is mandatory, and I suggest to set the possible values

config.toml.example
```diff
@@ -0,0 +1,189 @@
...
+# 	[AggSender.MaxSubmitCertificateRate]
+# 		NumRequests = 20
+# 		Interval = "1h"
+# 	[AggSender.OptimisticModeConfig]
+# 		# By default use the same key that aggsender signs certs
+# 		TrustedSequencerKey = {{AggSender.AggsenderPrivateKey}}
+# 	[AggSender.ValidatorClient]
+# 		URL = ""
+# 		UseTLS = false
+#
```
joanestebanr, line 132: The rest of field in `Aggsender` must not be change, so I suggest to removed it

config.toml.example
```diff
@@ -0,0 +1,189 @@
...
+# [Validator]
+# EnableRPC = false
+# [Validator.ServerConfig]
+# 	Host = "0.0.0.0"
+# 	Port = 5578
+# [Validator.AgglayerClient]
+# 	Cached = true
+# 	[Validator.AgglayerClient.ConfigurationCache]
+# 		TTL = "5m"
+#		Capacity = 100
```
joanestebanr: For `Validator` I suggest to remove everything but `Validator.ServerConfig` and add `Mode` as mandatory

config.toml.example
temaniarpit27: doesnt seem like it is used anywhere
removing it

config.toml.example
temaniarpit27: it is auto discover. i dont see this getting used anywhere

config.toml.example
joanestebanr: This field is mandatory, instead of a comment must be set to some value

config.toml.example
joanestebanr: ?, don't need to be defined?

config.toml.example
```diff
@@ -0,0 +1,154 @@
...
+#
+# [AggSender]
+# ------------------------------------------------------------------------------
+# Set to "AggchainProof" when enabling FEP. Possible values - "PessimisticProof", "AggchainProof"
+# ------------------------------------------------------------------------------
+# Mode = "PessimisticProof"
+# ------------------------------------------------------------------------------
+# Set RequireValidatorCall to true to enable aggsender validators
+# ------------------------------------------------------------------------------
+# RequireValidatorCall = false
```
joanestebanr: The key `RequireValidatorCall` must be removed also

config.toml.example
temaniarpit27: by default it is "PessimisticProof", isnt it?

config.toml.example
temaniarpit27: https://github.com/agglayer/aggkit/pull/944/files#diff-514f3c6049dd26198894991be37accf7786d0b38e3a1fc21f201c14be271b642R113

temaniarpit27: defined here

config.toml.example
temaniarpit27: done

config.toml.example
temaniarpit27: done

config.toml.example
temaniarpit27: done

config.toml.example
temaniarpit27: done

config.toml.example
temaniarpit27: done

config.toml.example
temaniarpit27: done

config.toml.example
temaniarpit27: done

config.toml.example
temaniarpit27: done

config.toml.example
temaniarpit27: done

config.toml.example
temaniarpit27: removed

config.toml.example
temaniarpit27: removed

config.toml.example
temaniarpit27: done

config.toml.example
temaniarpit27: done

config.toml.example
temaniarpit27: removed

config.toml.example
joanestebanr: Yes, but this must be defined by the user

config.toml.example
temaniarpit27: done

joanestebanr:
The rule of thumb between config file and `default.go` is: 
- the content of DefaultMandatoryVars must be in config file
- the content of DefaultVars must be in config file commented with the default values in default.go
- the content of DefaultValues must be not in config file but some vars that make sense to configured as: Log.Level , REST.Port, AggSender.Mode.... so you need to evaluate each one to decide if make sense to be exposed or not

config.toml.example
temaniarpit27: done

config.toml.example
temaniarpit27: done

joanestebanr (approved):

PR #954, opened 2025-09-08
feat: remove `metadata` field

### LINKED ISSUE #939: feat: `metadata` field modification when `multisig` and no `multisig`
`Metadata` field in `Certificate` should be removed in `aggkit`, and a `ZeroHash` should be used in its place when calculating `CertificateID`.

goran-ethernal:
## 🔄 Changes Summary
This PR removes the `Certificate.Metadata` field completely from the `aggkit`, as it is no longer needed to be sent to the `agglayer` from version `0.7.0` of `aggkit`.

Only thing worth noting is that, when calculating `CertificateID`, instead of `metadata` field, which was used previously, we now use `ZeroHash`.

This PR also updates the certificate `hash` to sign on the `proposer` and `validator`. Now, when we don't have `aggchain params` field in certificate, we use a `ZeroHash` instead (as per [this](https://github.com/agglayer/agglayer/blob/0698a81708b0a22d2dc971d4b867a6883f7f3efc/crates/pessimistic-proof-core/src/local_state/commitment.rs#L186) and [this](https://github.com/agglayer/agglayer/blame/0698a81708b0a22d2dc971d4b867a6883f7f3efc/crates/pessimistic-proof-core/src/aggchain_data/aggchain_hash.rs#L39) code on `agglayer). 

This PR removes the `validate_db` feature from the `aggsender-validator` which was used in local testing the test databases that their certificates are valid. Since now, we need `agglayer` `GetNetworkStatus` API to get the latest settled claim, this feature now does not make sense, and is removed. Along side with the `validate_db`, the `aggsender-validator` `rpc` endpoints are removed which were used for testing, to test the local `db` validity.

## ⚠️ Breaking Changes
NA

## 📋 Config Updates
NA

## ✅ Testing

## 🐞 Issues
- Closes #939


aggsender/validator/cert_hash.go
```diff
@@ -23,7 +23,7 @@ func HashCertificateToSign(cert *agglayertypes.Certificate) (common.Hash, error)
 
 	claimsHash := crypto.Keccak256(claimsRawMetadata)
 
-	var aggchainParams []byte
+	aggchainParams := aggkitcommon.ZeroHash.Bytes()
```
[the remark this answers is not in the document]
goran-ethernal: This doesn't make any sense. Its a local variable that has a default value of `ZeroHash`, if the certificate has `AggchainDataProof`, the value for `aggchainParams` will be used from that value.

joanestebanr (approved):

joanestebanr (commented):
Update PR description explaining that you remore RPC service for validator

goran-ethernal:
> Update PR description explaining that you remore RPC service for validator

@joanestebanr Updated.

aggsender/validator/cert_hash.go
```diff
@@ -37,3 +33,17 @@ func HashCertificateToSign(cert *agglayertypes.Certificate) (common.Hash, error)
 		cert.CertificateID().Bytes(),
 	), nil
 }
+
+func getAggchainParams(cert *agglayertypes.Certificate) []byte {
```
Stefan-Ethernal, line 39: Nitpick: add a comment to the function

Stefan-Ethernal (approved):

aggsender/validator/cert_hash.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/954/commits/47b0bc29164f25a6754211ab6b46a92b4330bc69

PR #976, opened 2025-09-10
feat: update sync status api

### LINKED ISSUE #969: Update Sync status API

rachit77:
## 🔄 Changes Summary
- This PR enhances the sync status API by making it more descriptive. It now reports both the most recently processed block number and the latest block produced on the network
- Bridge API now works even if only anyone of either `l1bridgesync` or `l2bridgesync` have been started

## 🐞 Issues
- Closes #969


bridgeservice/docs/docs.go
```diff
@@ -943,22 +943,37 @@ const docTemplate = `{
...
-                    "type": "integer"
+                    "type": "integer",
+                    "example": 676797
                 },
                 "contract_deposit_count": {
-                    "type": "integer"
+                    "type": "integer",
+                    "example": 676797
+                },
+                "is_active": {
```
temaniarpit27, line 953: check if bridge apis work when bridge service is up and running
if yes, then this will always be active

rachit77: Configured the bridge service to start as long as at least one bridge syncer is running
[ed9b723](https://github.com/agglayer/aggkit/pull/976/commits/ed9b723f5585628a5c916dd2d822d0d893d2a969)

temaniarpit27 (approved):

cmd/run.go
```diff
@@ -95,23 +95,36 @@ func start(cliCtx *cli.Context) error {
...
+	for _, component := range components {
+		if component == aggkitcommon.BRIDGE ||
+			component == aggkitcommon.L1BRIDGESYNC ||
+			component == aggkitcommon.L2BRIDGESYNC {
+			hasBridgeComponent = true
+			break
+		}
+	}
+
+	if hasBridgeComponent && (l1BridgeSync != nil || l2BridgeSync != nil) {
```
Stefan-Ethernal, line 117: I don't follow this, why we would need the bridge service in case l1 or l2 bridge syncer is up and running? For example, the aggsender uses the bridge syncer (L2 if I'm not mistaken), but it does not need it the bridge service to be up and running.

bridgeservice/types/types.go
```diff
@@ -322,24 +322,26 @@ type L1InfoTreeLeafResponse struct {
...
+// NetworkSyncInfo represents the bridge synchronization status of a single network (L1 or L2)
+// @Description Network-specific bridge sync status with deposit counts and block information
+// @example {"contract_deposit_count":100,"bridge_deposit_count":100,"is_synced":true,"is_active":true,
+// "last_processed_block":1234,"network_block":2555}
 type NetworkSyncInfo struct {
-	ContractDepositCount uint32 `json:"contract_deposit_count"`
-	BridgeDepositCount   uint32 `json:"bridge_deposit_count"`
-	IsSynced             bool   `json:"is_synced"`
+	ContractDepositCount uint32 `json:"contract_deposit_count" example:"676797" description:"Bridge deposits in contract"`
+	BridgeDepositCount   uint32 `json:"bridge_deposit_count" example:"676797" description:"Bridge deposits in database"`
```
Stefan-Ethernal: Can we find more suitable name for this (e.g. `SynchronizedDepositCount`)?

bridgeservice/bridge.go
```diff
@@ -320,6 +320,11 @@ func (b *BridgeService) GetBridgesHandler(c *gin.Context) {
 
 	switch networkID {
 	case mainnetNetworkID:
+		if b.bridgeL1 == nil {
```
Stefan-Ethernal, line 323: What is rationale to make the syncers optional? Wouldn't be better to have always the l1 and l2 bridge syncers running (note that it if not always, it will be almost always) that we require both syncers to be running, so that bridge service make sense. WDYT?

cmd/run.go
rachit77: The purpose of this is to ensure that since we’ve now made it possible for bridge syncers to run independently, there may be situations where someone runs l2bridgesync (for example, during a PP to FEP migration) before starting aggsender. In such cases, they may want to check the sync status of l2bridgesync using our provided API. To support this, our REST API server needs to be running.

bridgeservice/types/types.go
rachit77: sounds more intuitive, i will change this accordingly

bridgeservice/bridge.go
rachit77: Yes, but during PP → FEP migrations we only need the L2 bridge syncer to be running, the L1 bridge syncer is not required. Since we’ve now made it possible to run all syncers independently through [this issue](https://github.com/agglayer/aggkit/issues/880).
I think the API server should also be able to start independently in case of bridge service

cmd/run.go
Stefan-Ethernal: Tricky, walking on the edge of knife there I'd say... :cold_sweat: 

From the bridge service perspective, in order to consider it **fully** functional, it would need both l1 and l2 bridge syncers to be running. I do understand that during migrations we are running only certain syncers and we need just the sync status endpoint etc. However, I'm afraid that somebody forgets to enable both l1 and l2 bridge syncers and we start to get bugs on the bridge service.

At least, under normal circumstances, we should always require both bridge syncers to be up and running... Now how to accomplish it, I'm not sure.

Stefan-Ethernal (approved):

cmd/run.go
rachit77: resolving this as we have discussed this over the call

bridgeservice/types/types.go
rachit77: done: c99d36d95b97ee94dea2ecfde5363f2e30a3b68f

PR #978, opened 2025-09-10
fix: localValidator log error, fix conversion aggchainData

joanestebanr:
## 🔄 Changes Summary
- LocalValidator log the errors instead of failing
- AggchainData can be nil
- Fix minor errors on logs

### DIFF
--- a/test/config/fep-config.toml.template
+++ b/test/config/fep-config.toml.template
@@ -170,7 +170,7 @@ AggSenderPrivateKey = {Path = "{{.zkevm_l2_sequencer_keystore_file_path}}", Pass
 # LatestBlock, SafeBlock, PendingBlock, FinalizedBlock, EarliestBlock
 # Default value is "LatestBlock"
 # ------------------------------------------------------------------------------
-BlockFinality = "LatestBlock"
+#BlockFinality = "LatestBlock"
 
 Mode = "AggchainProof"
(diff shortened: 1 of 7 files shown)


Stefan-Ethernal (approved):
LGTM, aside from [the error](https://github.com/agglayer/aggkit/pull/978/files#diff-e97e339398dbae9a74e7910822feea52df3240fe2862f55c189c7b170e917984R21) can be removed

goran-ethernal (approved):

test/config/fep-config.toml.template
```diff
@@ -170,7 +170,7 @@ AggSenderPrivateKey = {Path = "{{.zkevm_l2_sequencer_keystore_file_path}}", Pass
 # LatestBlock, SafeBlock, PendingBlock, FinalizedBlock, EarliestBlock
 # Default value is "LatestBlock"
 # ------------------------------------------------------------------------------
-BlockFinality = "LatestBlock"
+#BlockFinality = "LatestBlock"
```
Stefan-Ethernal, line 173: As a side note, maybe we can remove this, AFAIK there is no `BlockFinality` in the `AggSender` config (anymore)?
```suggestion
```

joanestebanr: I agree, in the next PR I'll remove definitely

PR #997, opened 2025-09-15
fix: set block timestamp and tx hash to pre-etrog claims

### LINKED ISSUE #995: pre-etrog claims don't have block timestamp and tx hashes set
Pre-Etrog claims don't have block timestamp and tx hashes set (http://34.89.33.65:5577/bridge/v1/claims?network_id=0&page_number=2935), the values are `0` and `0x0` respectively.

Stefan-Ethernal:
## 🔄 Changes Summary
- Notice that the claims that are done on the pre-Etrog bridge contracts, have the `block_timestamp` and `tx_hash` fields populated, which is not the case currently (http://34.89.33.65:5577/bridge/v1/claims?network_id=0&page_number=2935).

### :warning: Note
It is necessary to re-index the data, in order to have the `block_timestamp` and `tx_hash` data populated for pre-Etrog claims.

## ⚠️ Breaking Changes
- 🛠️ **Config**: N/A
- 🔌 **API/CLI**: N/A

## 📋 Config Updates
- 🧾 **Diff/Config snippet**: N/A

## ✅ Testing
- 🤖 **Automatic**: N/A
- 🖱️ **Manual**:
1. Run the `aggkit` bridge service and index some bridge data for rollups that have been running prior to Etrog hardfork
2. Notice that the `tx_hash` and `block_timestamp` fields are set to non zero values

## 🐞 Issues
- Closes #995

### DIFF
--- a/bridgesync/downloader.go
+++ b/bridgesync/downloader.go
@@ -146,14 +146,13 @@ func buildClaimEventHandler(contract *polygonzkevmbridgev2.Polygonzkevmbridgev2,
 		claim := &Claim{
 			BlockNum:           b.Num,
 			BlockPos:           uint64(l.Index),
+			BlockTimestamp:     b.Timestamp,
+			TxHash:             l.TxHash,
 			GlobalIndex:        claimEvent.GlobalIndex,
 			OriginNetwork:      claimEvent.OriginNetwork,
 			OriginAddress:      claimEvent.OriginAddress,
 			DestinationAddress: claimEvent.DestinationAddress,
 			Amount:             claimEvent.Amount,
-			BlockTimestamp:     b.Timestamp,
-			TxHash:             l.TxHash,
-			FromAddress:        l.Address,
 		}
 
 		if syncFullClaims {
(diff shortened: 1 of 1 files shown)


joanestebanr (approved):

bridgesync/downloader.go
```diff
@@ -146,14 +146,13 @@ func buildClaimEventHandler(contract *polygonzkevmbridgev2.Polygonzkevmbridgev2,
...
+			BlockTimestamp:     b.Timestamp,
+			TxHash:             l.TxHash,
 			GlobalIndex:        claimEvent.GlobalIndex,
 			OriginNetwork:      claimEvent.OriginNetwork,
 			OriginAddress:      claimEvent.OriginAddress,
 			DestinationAddress: claimEvent.DestinationAddress,
 			Amount:             claimEvent.Amount,
-			BlockTimestamp:     b.Timestamp,
-			TxHash:             l.TxHash,
-			FromAddress:        l.Address,
```
temaniarpit27, line 156: why did we remove this?

temaniarpit27 (dismissed):

bridgesync/downloader.go
Stefan-Ethernal: The `FromAddress` is set [here](https://github.com/agglayer/aggkit/blob/f8eb45673994c63e4b1c611d3e781c6411b2b702/bridgesync/processor.go#L256), in case full claims synchronization is enabled. It is misleading to set it to contract's address in case full claims sync is disabled. So IMO it is better to not set it, than set the misleading value.

Note that we are not setting [it](https://github.com/agglayer/aggkit/blob/f8eb45673994c63e4b1c611d3e781c6411b2b702/bridgesync/downloader.go#L146-L156) for Etrog claims either.

temaniarpit27:
@Stefan-Ethernal will we do a resync after this change?

temaniarpit27 (approved):

Stefan-Ethernal:
> @Stefan-Ethernal will we do a resync after this change?

Yep, the resync is required (and I've indicated this in the PR description). Rajesh communicated this to devtools.

PR #1011, opened 2025-09-15
fix: #912 store non-accepted-cert in file

### LINKED ISSUE #912: bug: Non Accepted certificate must be stored on file
# Description
The function `saveNonAcceptedCert` are still writing the certificate into DB that is going to fail in case of cert > 1Gb. 
If this function fails, it has no consequences; therefore, the last certificate rejected by Agglayer would not be saved, but rather the last one rejected under 1 GB. This is one of the key fields for investigating issues in production.

Related to PR #808 and #836

joanestebanr:
## 🔄 Changes Summary
- A rejected certificates must be stored on filesystem instead of database due the sqlite limitation to 1Gb
- The content of the **non_accepted_cert** in table KV have changed the format (still compatible): 
    - Inside the JSON stored the field `SignedCertificate` if stars with *@* means a full path filename with the JSON of the certificate and the field `CertificateHash`have the hash of the JSON of the certificate

## 🐞 Issues
- Closes #912

### DIFF
--- a/aggsender/db/aggsender_db_storage.go
+++ b/aggsender/db/aggsender_db_storage.go
@@ -522,6 +522,7 @@ func (a *AggSenderSQLStorage) SaveNonAcceptedCertificate(
 	shouldRollback := true
 	defer func() {
 		if shouldRollback {
+			a.logger.Infof("saveNonAcceptedCertificate Rolling back transaction")
 			if errRllbck := tx.Rollback(); errRllbck != nil {
 				a.logger.Errorf(errWhileRollbackFormat, errRllbck)
 			}
@@ -556,8 +557,8 @@ func (a *AggSenderSQLStorage) SaveNonAcceptedCertificate(
 
 	shouldRollback = false
 
-	a.logger.Debugf("inserted non-accepted certificate - Height: %d. CreatedAt: %s",
-		nonAcceptedCert.Height, time.Unix(int64(nonAcceptedCert.CreatedAt), 0))
+	a.logger.Debugf("inserted non-accepted certificate - Height: %d. CreatedAt: %s Filename: %s",
+		nonAcceptedCert.Height, time.Unix(int64(nonAcceptedCert.CreatedAt), 0), fullPathFilename)
 
 	return nil
 }
--- a/aggsender/db/aggsender_db_storage.go
+++ b/aggsender/db/aggsender_db_storage.go
@@ -18,12 +18,14 @@ import (
 	"github.com/agglayer/aggkit/db"
 	dbtypes "github.com/agglayer/aggkit/db/types"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/iden3/go-iden3-crypto/keccak256"
 	"github.com/russross/meddler"
 )
 
 const (
-	errWhileRollbackFormat = "error while rolling back tx: %w"
-	nonAcceptedCertKey     = "non_accepted_cert"
+	errWhileRollbackFormat  = "error while rolling back tx: %w"
+	nonAcceptedCertKey      = "non_accepted_cert"
+	nonAcceptedCertFilename = "last_rejected_cert.json"
 )
 
 var newTxer = db.NewTx
@@ -278,18 +280,14 @@ func (a *AggSenderSQLStorage) SaveOrUpdateCertificate(ctx context.Context, certi
 // saveSignedCertificateToFile saves the signed certificate content to a file in the configured certificate directory
 // and returns the file path
 func (a *AggSenderSQLStorage) saveSignedCertificateToFile(
-	certificateHeight uint64,
-	certificateID common.Hash,
-	signedCertContent string,
-	retryCount int) (string, error) {
+	fileName string,
+	signedCertContent string) (string, error) {
 	// Use the configured certificate directory
 	certDir := a.cfg.CertificatesDir
 	if err := os.MkdirAll(certDir, 0755); err != nil { //nolint:mnd
 		return "", fmt.Errorf("failed to create certificates directory %s: %w", certDir, err)
 	}
 
-	// Create a filename using the certificate ID
-	fileName := fmt.Sprintf("signed_cert_%d_%s_%d.json", certificateHeight, certificateID, retryCount)
 	filePath := filepath.Join(certDir, fileName)
 
 	// Write the signed certificate content to the file
@@ -513,6 +511,9 @@ func (a *AggSenderSQLStorage) GetLastSentCertificateHeaderWithProofIfInError(
 // and to allow for debugging and analysis of why they were not accepted.
 func (a *AggSenderSQLStorage) SaveNonAcceptedCertificate(
 	ctx context.Context, nonAcceptedCert *NonAcceptedCertificate) error {
+	if nonAcceptedCert == nil {
+		return fmt.Errorf("saveNonAcceptedCertificate param nonAcceptedCert is nil")
+	}
 	tx, err := newTxer(ctx, a.db)
 	if err != nil {
 		return fmt.Errorf("failed to create db transaction for non-accepted certificate persistence: %w", err)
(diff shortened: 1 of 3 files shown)


aggsender/db/types.go
```diff
@@ -84,9 +84,12 @@ func (c *certificateInfo) ID() string {
 }
 
 type NonAcceptedCertificate struct {
-	Height            uint64 `meddler:"height"`
+	Height uint64 `meddler:"height"`
+	// SignedCertificate is "@<file_path>" or the actual JSON content of the certificate
 	SignedCertificate string `meddler:"signed_certificate"`
 	CreatedAt         uint32 `meddler:"created_at"`
+	// CertificateHash of the file that containt the certificate in JSON format
```
Stefan-Ethernal:
```suggestion
	// CertificateHash of the file that contains the certificate in JSON format
```

aggsender/db/aggsender_db_storage.go
```diff
@@ -564,18 +577,33 @@ func (a *AggSenderSQLStorage) GetNonAcceptedCertificate() (*NonAcceptedCertifica
...
+		FullFilename := nonAcceptedCert.SignedCertificate[1:]
+		data, err := os.ReadFile(FullFilename)
+		if err != nil {
+			return nil, fmt.Errorf("getNonAcceptedCertificate: failed to read signed certificate file %s: %w",
+				FullFilename, err)
+		}
+		certHash := common.BytesToHash(keccak256.Hash([]byte(string(data))))
+		if certHash != nonAcceptedCert.CertificateHash {
+			return nil, fmt.Errorf("getNonAcceptedCertificate: certificate hash mismatch: expected %s, got %s (file: %s)",
+				nonAcceptedCert.CertificateHash.String(), certHash.String(), FullFilename)
```
Stefan-Ethernal: I'm proposing the rename here, for the sake of case consistency. Wdyt?
```suggestion
		certificateFilePath := nonAcceptedCert.SignedCertificate[1:]
		data, err := os.ReadFile(certificateFilePath)
		if err != nil {
			return nil, fmt.Errorf("getNonAcceptedCertificate: failed to read signed certificate file %s: %w",
				certificateFilePath, err)
		}
		certHash := common.BytesToHash(keccak256.Hash([]byte(string(data))))
		if certHash != nonAcceptedCert.CertificateHash {
			return nil, fmt.Errorf("getNonAcceptedCertificate: certificate hash mismatch: expected %s, got %s (file: %s)",
				nonAcceptedCert.CertificateHash.String(), certHash.String(), certificateFilePath)
```

aggsender/db/aggsender_db_storage.go
```diff
@@ -564,18 +577,33 @@ func (a *AggSenderSQLStorage) GetNonAcceptedCertificate() (*NonAcceptedCertifica
 	if err := json.Unmarshal([]byte(val), &nonAcceptedCert); err != nil {
 		return nil, fmt.Errorf("failed to unmarshal non-accepted certificate: %w", err)
 	}
+	if nonAcceptedCert.SignedCertificate != "" && nonAcceptedCert.SignedCertificate[0] == '@' {
```
Stefan-Ethernal: Wdyt? Also it would probably make sense to extract the `@` to a const.
```suggestion
	if strings.HasPrefix(nonAcceptedCert.SignedCertificate, "@") {
```

aggsender/db/aggsender_db_storage.go
```diff
@@ -521,13 +522,25 @@ func (a *AggSenderSQLStorage) SaveNonAcceptedCertificate(
...
+	fullPathFilename, err := a.saveSignedCertificateToFile(filename, nonAcceptedCert.SignedCertificate)
+	if err != nil {
+		return fmt.Errorf("saveNonAcceptedCertificate: failed to save signed certificate to file: %w", err)
+	}
+	entry := &NonAcceptedCertificate{
+		Height:            nonAcceptedCert.Height,
+		CreatedAt:         nonAcceptedCert.CreatedAt,
+		Error:             nonAcceptedCert.Error,
+		SignedCertificate: "@" + fullPathFilename,
+		CertificateHash:   common.BytesToHash(keccak256.Hash([]byte(nonAcceptedCert.SignedCertificate))),
```
Stefan-Ethernal: Simpler to reuse the `Keccak256` function from the go ethereum, since we are relying on its data structure for hash and then no need to convert anything from bytes to hash etc.
```suggestion
		CertificateHash:   crypto.Keccak256Hash([]byte(nonAcceptedCert.SignedCertificate)),
```

Stefan-Ethernal: As a side note, it seems that we could get rid off the `github.com/iden3/go-iden3-crypto` dependency, since the `keccak256` calculation can be reused from go ethereum's crypto package, and the `merkletree` package where it is also used, seem to not be in use in the aggkit, so maybe take this opportunity and remove the dependency and the `merkletree` package altogether (it can be a separate PR for sure). WDYT?

aggsender/db/aggsender_db_storage.go
```diff
@@ -521,13 +522,25 @@ func (a *AggSenderSQLStorage) SaveNonAcceptedCertificate(
...
+			a.logger.Infof("saveNonAcceptedCertificate Rolling back transaction")
 			if errRllbck := tx.Rollback(); errRllbck != nil {
 				a.logger.Errorf(errWhileRollbackFormat, errRllbck)
 			}
 		}
 	}()
-
-	raw, err := json.Marshal(nonAcceptedCert)
+	filename := nonAcceptedCertFilename
+	fullPathFilename, err := a.saveSignedCertificateToFile(filename, nonAcceptedCert.SignedCertificate)
```
rachit77, line 533: i am wondering what happens if aggkit crashes while writing to `last_rejected_cert.json`
to avoid leaving a corrupted file would it make sense to first write the content to a temporary file, ensure it’s flushed and then rename it to replace the previous `last_rejected_cert.json` instead of overwriting the file directly each time?

aggsender/db/aggsender_db_storage.go
joanestebanr: I understand that the idea is rename `FullFilename` in favour `certificateFilePath`? isn't? I applied that

aggsender/db/aggsender_db_storage.go
joanestebanr: I think that is overcomplicating adding a const for that, but anyway I applied your suggestion

aggsender/db/aggsender_db_storage.go
joanestebanr: This library is used for serval syncer for calculate `keccak256`, but MT use `poseidon`

aggsender/db/aggsender_db_storage.go
joanestebanr: That is the reason that I introduce the `certificateHash` you can check the integrity. This data is never used, basically is for debug, so it's enough to known that the data on filesystem doesn't match the expected one

aggsender/db/aggsender_db_storage.go
Stefan-Ethernal: That's right.

aggsender/db/aggsender_db_storage.go
```diff
@@ -564,18 +578,33 @@ func (a *AggSenderSQLStorage) GetNonAcceptedCertificate() (*NonAcceptedCertifica
...
 	}
+	if strings.HasPrefix(nonAcceptedCert.SignedCertificate, prefixFilename) {
+		// The content is pointing to a file
+		certificateFilePath := nonAcceptedCert.SignedCertificate[1:]
+		data, err := os.ReadFile(certificateFilePath)
+		if err != nil {
+			return nil, fmt.Errorf("getNonAcceptedCertificate: failed to read signed certificate file %s: %w",
+				certificateFilePath, err)
+		}
+		certHash := crypto.Keccak256Hash([]byte(string(data)))
```
Stefan-Ethernal: It should be as simple as this :slightly_smiling_face:

```suggestion
		certHash := crypto.Keccak256Hash(data)
```

Stefan-Ethernal (approved):

aggsender/db/aggsender_db_storage.go
joanestebanr: good point!, thanks you!

rachit77 (approved):

PR #1024, opened 2025-09-18
fix: multi chains (2 chains) pessimistic proof e2e tests

### LINKED ISSUE #1022: fix multi chains pessimistic proof tests
The `aggkit-002` (aggsender proposer for 2nd rollup) is unable to retrieve the multisig committee:
```json
{
  "level": "error",
  "ts": 1758261681.7834673,
  "caller": "aggsender/aggsender.go:301",
  "msg": "error polling validator committee: failed to get validators: failed to retrieve the latest multisig committee: the committee cannot be empty",
  "pid": 7,
  "version": "",
  "module": "aggsender",
  "stacktrace": "github.com/agglayer/aggkit/aggsender.(*AggSender).sendCertificateWithRetries.func1\n\t/app/aggsender/aggsender.go:301\ngithub.com/agglayer/aggkit/common.Execute[...]\n\t/app/common/retry_handler_delays.go:131\ngithub.com/agglayer/aggkit/aggsender.(*AggSender).sendCertificateWithRetries\n\t/app/aggsender/aggsender.go:293\ngithub.com/agglayer/aggkit/aggsender.(*AggSender).sendCertificates\n\t/app/aggsender/aggsender.go:267\ngithub.com/agglayer/aggkit/aggsender.(*AggSender).Start\n\t/app/aggsender/aggsender.go:200"
}
```

Stefan-Ethernal:
## 🔄 Changes Summary
Fix the PP multi chains E2E tests (2 chains):
- remove setting gas-price explicitly when claiming
- remove sending legacy transactions for bridges and claims

## ⚠️ Breaking Changes
- 🛠️ **Config**: N/A
- 🔌 **API/CLI**: N/A

## 📋 Config Updates

## ✅ Testing
- 🤖 **Automatic**: Make sure `Multi chains E2E test (2 chains)` are :green_circle: 
- 🖱️ **Manual**: N/A

## 🐞 Issues
- Closes #1022


.github/workflows/test-e2e.yml
```diff
@@ -294,23 +301,23 @@ jobs:
...
     steps:
       - run: exit 1
-        if: ${{ contains(fromJSON('["failure", "cancelled"]'), needs.test-multi-l2-networks-3-chains.result) }}
+        if: ${{ contains(fromJSON('["failure", "cancelled"]'), needs.test-multi-l2-networks-3-chains-op-pessimistic.result) }}
 
   notify-slack:
     name: Notify Slack
     needs:
-      - test-single-l2-network-fork12-pessimistic
+      - test-single-l2-network-op-pessimistic
```
temaniarpit27, line 316: are we replacing cdk with op?

.github/test_e2e_cdk_args_base.json
```diff
@@ -5,7 +5,6 @@
   "args": {
     "consensus_contract_type": "ecdsa_multisig",
     "use_agg_sender_validator": true,
```
temaniarpit27, line 7: why are we always using aggsender validator?

temaniarpit27: also, should the filename change?

temaniarpit27: maybe add a new case with aggsender validator and leave the old as it is?

.github/workflows/test-e2e.yml
Stefan-Ethernal: To provide you with the context:
- we are deprecating the `PolygonPessimisticConsensus` (and all the legacy rollup implementations AFAIK), in favor of `AggchainECDSAMultisig` (PP certificates) and `AggchainFEP` (FEP certificates),
- kurtosis cdk was not supporting the cdk erigon as the L2, up until two days ago, and the `op-stack` was the only option we had for running pessimistic proofs (that's why we have chosen to integrate this part into the CI first). After this PR, I'm going to start integrating the cdk erigon path as well,
- probably it is enough to have only single chain tests running against the cdk erigon setup. IMO it is better that way, because op-stack acts as the vanilla client.

.github/test_e2e_cdk_args_base.json
Stefan-Ethernal: > why are we always using aggsender validator?

This is the preferred option, as the most likely it is going to be like that in the production. For sure we can use the single signer as well, but that is more like a special case, rather than the general rule.

> also, should the filename change?

Sure, I can consider it, although maybe would leave it when I integrate the cdk erigon setup. Any ideas?

> maybe add a new case with aggsender validator and leave the old as it is?

To me we can do the opposite, as having the single signer is not considered as a safe option, so therefore we can consider adding a path to test it without the validators (although it is just a subset of what we currently have: if this more complex setup works, chances are the "single signer" setup would work as well).

goran-ethernal (approved):
I agree with @temaniarpit27 , maybe we should rename the file to be more precise.

temaniarpit27 (approved):

joanestebanr (approved):

PR #1029, opened 2025-09-18
feat: report unsupported component

### LINKED ISSUE #1027: fix: if you set a component unknown must emit an error
# Bug Report
Running something like that `aggkit run -components "this_is_not_a_component"` must generate an error

rachit77:
## 🔄 Changes Summary
- Throw error when an unsupported component is provided to aggkit

## 🐞 Issues
- Closes #1027


temaniarpit27 (approved):
remove aggkit binary

common/components_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := ValidateComponents(tt.components)
+
+			if tt.expectError {
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), tt.errorMsg)
+			} else {
+				assert.NoError(t, err)
```
Stefan-Ethernal:
```suggestion
				require.ErrorContains(t, err, tt.errorMsg)
			} else {
				require.NoError(t, err)
```

common/components.go
```diff
@@ -22,3 +24,37 @@ const (
...
+				AGGSENDERVALIDATOR,
+				L1INFOTREESYNC,
+				L2BRIDGESYNC,
+				L1BRIDGESYNC,
+				L2GERSYNC)
+		}
+	}
+
+	return nil
+}
```
Stefan-Ethernal: WDYT (my point is that I wanted to avoid enumerating the supported components twice, but instead rely on the set and sort it alphabetically)?

```suggestion
// ValidateComponents validates that all provided components are known/supported.
func ValidateComponents(components []string) error {
	validComponents := map[string]struct{}{
		AGGORACLE:          {},
		BRIDGE:             {},
		PROVER:             {},
		AGGSENDER:          {},
		L1INFOTREESYNC:     {},
		L2BRIDGESYNC:       {},
		L1BRIDGESYNC:       {},
		L2GERSYNC:          {},
		AGGCHAINPROOFGEN:   {},
		AGGSENDERVALIDATOR: {},
	}

	// build a sorted list of valid component names for error messages
	keys := make([]string, 0, len(validComponents))
	for k := range validComponents {
		keys = append(keys, k)
	}
	sort.Strings(keys) // ensures deterministic ordering
	validList := strings.Join(keys, ", ")

	for _, component := range components {
		if _, ok := validComponents[component]; !ok {
			return fmt.Errorf("unknown component: %s. Valid components are: %s",
				component, validList)
		}
	}

	return nil
}
```

Stefan-Ethernal (approved):
Generally LGTM, but leaving some suggestions

PR #1034, opened 2025-09-19
feat: improve request to committee logs #1023 [INTEGRATION]

### LINKED ISSUE #1023: log context canceled error on an INFO log level
Version: `v0.7.0-beta4`

The context canceled is expected to happen every time the committee is larger than the configured threshold and at the given point the signatures count >= threshold. Therefore we should figure out if it is possible to log this message on an `INFO` log level.
```json
{
  "level": "error",
  "ts": 1757510849.6013253,
  "caller": "aggsender/validator_poller.go:161",
  "msg": "validator RemoteValidator (URL=http://aggkit-001-aggsender-validator-001:5578, Address=0x97bC60cc5BcD071536F28F784378c8c1E95446e6) failed to validate the certificate: error validating certificate on aggsender validator service: aggsender validator failed to successfully validate certificate: rpc error: code = Canceled desc = context canceled",
  "pid": 7,
  "version": "",
  "module": "aggsender",
  "stacktrace": "github.com/agglayer/aggkit/aggsender.(*validatorPoller).executeValidation.func1\n\t/app/aggsender/validator_poller.go:161"
}
```

joanestebanr:
## 🔄 Changes Summary
- When we have enough signature from committee the rest of call are canceled, and must no be logged as a ERROR because is the expected behaviour

## 🐞 Issues
- Closes #1023

### DIFF
--- a/aggsender/block_notifier_polling.go
+++ b/aggsender/block_notifier_polling.go
@@ -168,7 +168,7 @@ func (b *BlockNotifierPolling) step(ctx context.Context,
 		newState := previousState.initialBlock(currentBlock)
 		return b.nextBlockRequestDelay(nil, nil), newState, eventToEmit
 	}
-	newState := previousState.incommingNewBlock(currentBlock)
+	newState := previousState.incomingNewBlock(currentBlock)
 	b.logger.Debugf("New block seen [finality:%s]: %d. blockRate:%s",
 		b.config.BlockFinalityType.String(), currentBlock, newState.previousBlockTime)
 	eventToEmit.BlockRate = *newState.previousBlockTime
@@ -221,7 +221,7 @@ func (s *blockNotifierPollingInternalStatus) initialBlock(block uint64) *blockNo
 	}
 }
 
-func (s *blockNotifierPollingInternalStatus) incommingNewBlock(block uint64) *blockNotifierPollingInternalStatus {
+func (s *blockNotifierPollingInternalStatus) incomingNewBlock(block uint64) *blockNotifierPollingInternalStatus {
 	now := timeNowFunc()
 	timePreviousBlock := now.Sub(s.lastBlockTime)
 	return &blockNotifierPollingInternalStatus{
--- a/aggsender/block_notifier_polling.go
+++ b/aggsender/block_notifier_polling.go
@@ -170,7 +170,7 @@ func (b *BlockNotifierPolling) step(ctx context.Context,
 	}
 	newState := previousState.incommingNewBlock(currentBlock)
 	b.logger.Debugf("New block seen [finality:%s]: %d. blockRate:%s",
-		b.config.BlockFinalityType, currentBlock, newState.previousBlockTime)
+		b.config.BlockFinalityType.String(), currentBlock, newState.previousBlockTime)
 	eventToEmit.BlockRate = *newState.previousBlockTime
 	return b.nextBlockRequestDelay(newState, nil), newState, eventToEmit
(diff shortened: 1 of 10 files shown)


aggsender/block_notifier_polling.go
```diff
@@ -170,7 +170,7 @@ func (b *BlockNotifierPolling) step(ctx context.Context,
 	}
 	newState := previousState.incommingNewBlock(currentBlock)
```
Stefan-Ethernal: Nitpick :slightly_smiling_face: 
```suggestion
	newState := previousState.incomingNewBlock(currentBlock)
```

aggsender/types/multisig_committee.go
```diff
@@ -91,6 +91,14 @@ func (m *MultisigCommittee) Threshold() *big.Int {
 	return m.threshold
```
Stefan-Ethernal, line 90: We can remove this function.

aggsender/types/multisig_committee.go
```diff
@@ -91,6 +91,14 @@ func (m *MultisigCommittee) Threshold() *big.Int {
...
 }
 
+// ThresholdInt returns the signature threshold as an int64.
+func (m *MultisigCommittee) ThresholdInt() (int64, error) {
+	if m.threshold.IsInt64() {
+		return m.threshold.Int64(), nil
+	}
+	return 0, fmt.Errorf("threshold %s doesn't fit in int64", m.threshold.String())
+}
```
Stefan-Ethernal: Maybe makes slightly more sense to represent it as `uint64` (since it shouldn't be negative), rename it and remove the original `Threshold` function. WDYT?

```suggestion
// Threshold returns the signature threshold required for quorum.
func (m *MultisigCommittee) Threshold() (uint64, error) {
	if m.threshold.IsUint64() {
		return m.threshold.Uint64(), nil
	}
	return 0, fmt.Errorf("threshold %d doesn't fit in uint64", m.threshold)
}
```

Btw maybe it makes sense to discuss with protocol team to change it on their end as well to int64/uint64 (if it is not already late, because of ongoing audit etc.)?

aggsender/validator_poller.go
```diff
@@ -220,30 +230,26 @@ func (vp *validatorPoller) processResults(
 	for res := range resultsCh {
 		if res.err != nil {
 			errs = append(errs, res.err)
-
-			metrics.ValidatorError(res.validator.Address())
-
+			vp.log.Errorf("validatorRequest returned an error: %s", res.String())
```
Stefan-Ethernal, line 229: Does it make sense to revert the metrics?


```suggestion
            metrics.ValidatorError(res.validator.Address())
			vp.log.Errorf("validatorRequest returned an error: %s", res.String())
```

aggsender/types/multisig_committee.go
joanestebanr: The type in the [contract](https://github.com/agglayer/agglayer-contracts/blob/e2a68464f2f34a759731089ed26d2769fce1c2d3/contracts/v2/lib/AggchainBase.sol#L77) is `uint256`, so I agree that we can use `uint64`

aggsender/validator_poller.go
joanestebanr: good catch!

Stefan-Ethernal (approved):
Just fix the failing `Test_ECDSAMultisigCommitteeQuery_GetMultisigCommittee`, otherwise LGTM

PR #1035, opened 2025-09-19
fix: rate_limit object in case no sleep required

### LINKED ISSUE #1033: bug: false warning in rate_limit_wrapper
The rate_limit_wrapper emit a false `warning` log: 
```
[aggkit-001] {"level":"warn","ts":1758268785.8932679,"caller":"agglayer/rate_limit_wrapper.go:60","msg":"rate limit reached for SendCertificate, sleeping for 0s. Rate:RateLimit{cfg: RateLimitConfig{NumRequests: 20, Period: 1h0m0s}, bucket len: 20}","pid":7,"version":"","module":"aggsender"}
```

version affected: 0.7.0-beta4, 0.7.0-beta5, and develop

joanestebanr:
## 🔄 Changes Summary
- Fix false warning each time a certificate is submitted.
The object `RateLimit` was changed in a way that no longer response with a `nil` if no sleep required

## 🐞 Issues
- Closes #1033

## 🔗 Related PRs
- #824

### DIFF
--- a/common/rate_limit.go
+++ b/common/rate_limit.go
@@ -66,24 +66,27 @@ func (r *RateLimit) Call(msg string, allowToSleep bool) *time.Duration {
 	if r == nil || !r.cfg.Enabled() {
 		return nil
 	}
-	var sleepTime time.Duration
+	var returnSleepTime *time.Duration
 	now := TimeProvider()
 	r.cleanOutdatedCalls(now)
 
 	// Rate limit check BEFORE adding the current call
 	if len(r.calls) >= r.cfg.NumRequests {
-		sleepTime = r.cfg.Interval.Duration - now.Sub(r.calls[0])
+		sleepTime := r.cfg.Interval.Duration - now.Sub(r.calls[0])
 		if allowToSleep {
 			if msg != "" {
 				log.Infof("Rate limit reached, sleeping for %s for %s", sleepTime, msg)
 			}
 			time.Sleep(sleepTime)
+		} else {
+			// If no sleep, we return the time
+			returnSleepTime = &sleepTime
 		}
 	}
 
 	// Add the current call to the tracking
 	r.calls = append(r.calls, now)
-	return &sleepTime
+	return returnSleepTime
 }
 
 func (r *RateLimit) cleanOutdatedCalls(now time.Time) {
--- a/common/rate_limit_test.go
+++ b/common/rate_limit_test.go
@@ -56,8 +56,8 @@ func TestRateLimitSleepTime(t *testing.T) {
 	elapsed := time.Since(start)
 	require.Nil(t, sleepTime)
 	// The call should have slept for approximately 5 seconds
-	require.True(t, elapsed >= time.Second*1, "Expected to sleep for at least 1 seconds (1min- 59sec), but slept for %v", elapsed)
-	require.True(t, elapsed <= time.Second*2, "Expected to sleep for at most 1  seconds (error margin of 1sec), but slept for %v", elapsed)
+	require.True(t, elapsed >= time.Second*1, "Expected to sleep for at least 1 seconds (1minute - 59 seconds), but slept for %v", elapsed)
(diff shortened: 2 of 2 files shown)


common/rate_limit.go
```diff
@@ -66,24 +66,27 @@ func (r *RateLimit) Call(msg string, allowToSleep bool) *time.Duration {
...
 	if len(r.calls) >= r.cfg.NumRequests {
-		sleepTime = r.cfg.Interval.Duration - now.Sub(r.calls[0])
+		sleepTime := r.cfg.Interval.Duration - now.Sub(r.calls[0])
 		if allowToSleep {
 			if msg != "" {
 				log.Infof("Rate limit reached, sleeping for %s for %s", sleepTime, msg)
 			}
 			time.Sleep(sleepTime)
+		} else {
+			// If no sleep, we return the time
```
rachit77, line 82: How about this?
```suggestion
			// If not allowed to sleep, return a *time.Duration indicating how long
            // the caller should wait. Returns nil when no wait is required.
```

common/rate_limit.go
```diff
@@ -66,24 +66,27 @@ func (r *RateLimit) Call(msg string, allowToSleep bool) *time.Duration {
...
 	}
-	var sleepTime time.Duration
+	var returnSleepTime *time.Duration
 	now := TimeProvider()
 	r.cleanOutdatedCalls(now)
 
 	// Rate limit check BEFORE adding the current call
 	if len(r.calls) >= r.cfg.NumRequests {
-		sleepTime = r.cfg.Interval.Duration - now.Sub(r.calls[0])
+		sleepTime := r.cfg.Interval.Duration - now.Sub(r.calls[0])
```
rachit77, line 75: I am wondering are we handling possibility of negetive `sleepTime` here?

rachit77 (approved):

common/rate_limit.go
joanestebanr: *TDLR;* So I think that we don't have to think about this case because even a negative value of `sleepTime` have no consequences

In fact, the code is a revert of PR #824, but I agree that is a interesting point... 
In a regular case the `r.calls[0]` is going to be always >= `now` by how is done the code. But if we think about any situation that something goes wrong and we got `r.calls[0]` < `now` the `sleepTime` var is going to be negative and `Sleep` function documentation says: `A negative or zero duration causes Sleep to return immediately.` ([source](https://go.dev/src/time/sleep.go)).

common/rate_limit_test.go
```diff
@@ -39,26 +37,27 @@ func TestRateLimitSleepTime(t *testing.T) {
...
 	start := time.Now()
-	sut.Call("test", true)
+	sleepTime = sut.Call("test", true)
 	elapsed := time.Since(start)
-
+	require.Nil(t, sleepTime)
 	// The call should have slept for approximately 5 seconds
-	require.True(t, elapsed >= time.Second*4, "Expected to sleep for at least 4 seconds, but slept for %v", elapsed)
-	require.True(t, elapsed <= time.Second*6, "Expected to sleep for at most 6 seconds, but slept for %v", elapsed)
+	require.True(t, elapsed >= time.Second*1, "Expected to sleep for at least 1 seconds (1min- 59sec), but slept for %v", elapsed)
```
Stefan-Ethernal:
```suggestion
	require.True(t, elapsed >= time.Second*1, "Expected to sleep for at least 1 seconds (1min - 59sec), but slept for %v", elapsed)
```

common/rate_limit_test.go
```diff
@@ -39,26 +37,27 @@ func TestRateLimitSleepTime(t *testing.T) {
...
+	require.True(t, elapsed <= time.Second*2, "Expected to sleep for at most 1  seconds (error margin of 1sec), but slept for %v", elapsed)
```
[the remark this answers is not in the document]
Stefan-Ethernal: Nitpick, I believe the assertion msg is not correct:
```suggestion
	require.True(t, elapsed <= time.Second*2, "Expected to sleep for at most 2 seconds (error margin of 1 second), but slept for %v", elapsed)
```

Stefan-Ethernal (approved):

PR #1051, opened 2025-09-23
feat: handle L2 GER syncing 

### LINKED ISSUE #1032: Handle L2GERSync when GER synced is not available on L1InfoTreeSync
There could be a case where bridge is running on a different container and l1infotreesync is not yet synced in the bridge container

rachit77:
## 🔄 Changes Summary
- This PR introduces change that handles L2 GER syncing when GER is not available on L1

## 🐞 Issues
- Closes #1032


l2gersync/evm_downloader_sovereign.go
```diff
@@ -151,3 +166,15 @@ func (d *downloaderSovereign) buildAppender(
 
 	return appender
 }
+
+// isL1InfoTreeSyncUpToDate checks if L1InfoTreeSync is up to date with L1 blocks
+func (d *downloaderSovereign) isL1InfoTreeSyncUpToDate(ctx context.Context) (bool, error) {
+	// We need to cast to the concrete type to access IsUpToDate method
+	if l1InfoTreeSync, ok := d.l1InfoTreeSync.(interface {
```
goran-ethernal: Why not add this function to the `L1InfoTreeQuerier` interface?

l2gersync/evm_downloader_sovereign_test.go
```diff
@@ -111,3 +118,101 @@ func TestDownloaderSovereign_Download(t *testing.T) {
 	mockL2Client.AssertExpectations(t)
 	mockL1InfoTreeSync.AssertExpectations(t)
 }
+
+// mockL1InfoTreeSyncWithIsUpToDate extends the generated mock to include IsUpToDate method
+type mockL1InfoTreeSyncWithIsUpToDate struct {
```
goran-ethernal: You wouldn't need this mock if you add the function to the `L1InfoTreeQuerier` interface.

l2gersync/evm_downloader_sovereign.go
```diff
@@ -131,9 +134,21 @@ func (d *downloaderSovereign) buildAppender(
...
+		isUpToDate, err := d.isL1InfoTreeSyncUpToDate(ctx)
+		if err != nil {
+			log.Warnf("Failed to check if L1InfoTreeSync is up to date: %v", err)
+		}
+
+		// L1InfoTreeSync is up to date, proceed with GER lookup
 		l1InfoTreeLeaf, err := d.l1InfoTreeSync.GetInfoByGlobalExitRoot(insertGEREvent.NewGlobalExitRoot)
 		if err != nil {
-			log.Fatalf("GER %s received from L2 is not present in L1InfoTreeSync: %v",
+			if isUpToDate {
```
goran-ethernal, line 144: I am unsure about what this `IsUpToDate` does. As far as I know, `L1InfoTreeSyncer` is always syncing finalized blocks. Why do we require it to have synced up the latest finalized block before querying the `GER` from it?

l2gersync/evm_downloader_sovereign.go
```diff
@@ -131,9 +134,21 @@ func (d *downloaderSovereign) buildAppender(
 			return fmt.Errorf("error parsing UpdateHashChainValue event log %+v: %w", l, err)
 		}
 
+		// Check if L1InfoTreeSync is up to date before trying to get GER info
+		ctx := context.Background()
```
temaniarpit27: i dont think we need this check for uptodate
we can just check for mismatch, and if mismatches, check on l1 if ger exists

l2gersync/evm_downloader_sovereign.go
temaniarpit27: +1

l2gersync/evm_downloader_sovereign.go
rachit77: Because the `L1InfoTreeQuerier` interface is also used by `aggsender` in addition to `l2gersync`. If I add an `IsUpToDate` method to `L1InfoTreeQuerier`, I would then need to add a dummy implementation in `aggsender` as well, which I wanted to avoid

l2gersync/evm_downloader_sovereign_test.go
rachit77: I agree we can remove this if we go ahead with adding of function to `L1InfoTreeQuerier` interface.

l2gersync/evm_downloader_sovereign.go
rachit77: Instead of checking directly from L1, I’d prefer to check against `l1infotreesync`, since we already index all that data from L1 into the syncer DB.
`l2gersync` always depends on `l1infotreesync` being in sync with the GER it is indexing. This is because `l2gersync` populates the `l1infotreeindex` in its DB using data from the `l1infotreesync` DB. So if you’re syncing GER *x* with `l2gersync`, then `l1infotreesync` must have already indexed GER *x* from L1 otherwise `l2gersync` can’t populate the index.

Given that dependency, it makes much more sense to perform the check against `l1infotreesync`.

rachit77: [40c0422](https://github.com/agglayer/aggkit/pull/1051/commits/40c04223233546d35c7ca76d0821472ee7b28757)

moved `isL1InfoTreeSyncUpToDate` method call inside the error block so that it will be called only in case when ger in not present on L1InfoTreeSync

goran-ethernal: Ok, I've just checked, `aggoracle` is inserting `GERs` on `L2` based on the last `L1 info tree leaf` from the `L1 Info Tree Syncer`, and since `L1 Info Tree Syncer` is only syncing `Finalized` blocks, I would say, this solution is then good enough in my opinion.

l2gersync/evm_downloader_sovereign.go
goran-ethernal: Both `aggsender` and `l2_ger_syncer` packages have their own `L1InfoTreeQuerier` interfaces, they are not related. You can add this function to the interface in the `l2_ger_syncer` package, and remove this code and your mock.

rachit77: done [40c0422](https://github.com/agglayer/aggkit/pull/1051/commits/40c04223233546d35c7ca76d0821472ee7b28757)

l2gersync/evm_downloader_sovereign.go
rachit77: pasting the answer from slack here for future reference if needed:

The core purpose of that PR is to identify and flag GERs that are inserted on L2 during event indexing in l2gersync but are not present on L1.
The IsUpToDate check is not a blanket requirement for all GER queries. It only becomes necessary when we encounter a GER on L2 that cannot be found in the l1infotreesync database.

Scenario Example:
Consider a distributed setup with two separate aggkit instances:
Instance A: Runs aggoracle, aggsender, etc., but excludes l2gersyncer and bridge service
Instance B: Runs l2gersyncer
In this scenario, if the l1infotreesyncer on Instance B falls behind Instance A's sync progress, we may receive new GERs inserted on L2 through l2gersync before our L1 sync has caught up.
During this lag period, we cannot definitively determine whether a GER found on L2 is invalid or simply not yet synced from L1. Without ensuring that l1infotreesync is fully synchronized, we risk incorrectly flagging valid GERs as invalid due to incomplete L1 data.

l2gersync/evm_downloader_sovereign_test.go
rachit77: done: [e162bd4](https://github.com/agglayer/aggkit/pull/1051/commits/e162bd43496515a84af4a8089dce4dd345e5daad)

aggsender/types/interfaces.go
```diff
@@ -65,6 +65,7 @@ type L1InfoTreeSyncer interface {
 	GetProcessedBlockUntil(ctx context.Context, blockNumber uint64) (uint64, common.Hash, error)
 	GetInfoByIndex(ctx context.Context, index uint32) (*l1infotreesync.L1InfoTreeLeaf, error)
 	GetLatestL1InfoLeafUntilBlock(ctx context.Context, blockNum uint64) (*l1infotreesync.L1InfoTreeLeaf, error)
+	IsUpToDate(ctx context.Context, l1Client aggkittypes.BaseEthereumClienter) (bool, error)
```
goran-ethernal, line 68: Why do you need to expand the interface in the `aggsender` if this function is not used in the `aggsender` package?

rachit77: because we are using [l2GERReaderFactory](https://github.com/agglayer/aggkit/blob/ac00454d4c65c7d188128d077f3e30d5be42c73c/aggsender/flows/factory.go#L26) in aggsender which reference the `L1InfoTreeQuerier` from `l2gersync`

and `l1InfoTreeSyncer` being passed [here](https://github.com/agglayer/aggkit/blob/ac00454d4c65c7d188128d077f3e30d5be42c73c/aggsender/flows/factory.go#L120-L121) is of type [aggsender L1InfoTreeSyncer](https://github.com/agglayer/aggkit/blob/ac00454d4c65c7d188128d077f3e30d5be42c73c/aggsender/types/interfaces.go#L58) while it expects it to be of type[ l2gersync L1InfoTreeQuerier](https://github.com/agglayer/aggkit/blob/ac00454d4c65c7d188128d077f3e30d5be42c73c/l2gersync/l2_evm_ger_reader.go#L25-L26)

l2gersync/evm_downloader_sovereign.go
```diff
@@ -131,10 +134,21 @@ func (d *downloaderSovereign) buildAppender(
 			return fmt.Errorf("error parsing UpdateHashChainValue event log %+v: %w", l, err)
 		}
 
+		// L1InfoTreeSync is up to date, proceed with GER lookup
```
goran-ethernal: This comment is not on the right place it seems.

l2gersync/evm_downloader_sovereign.go
```diff
@@ -151,3 +165,8 @@ func (d *downloaderSovereign) buildAppender(
 
 	return appender
 }
+
+// isL1InfoTreeSyncUpToDate checks if L1InfoTreeSync is up to date with L1 blocks
+func (d *downloaderSovereign) isL1InfoTreeSyncUpToDate(ctx context.Context) (bool, error) {
```
goran-ethernal: I don't think you need this wrapper function, just call the `d.l1InfoTreeSync.IsUpToData` on the place you need it.

l2gersync/evm_downloader_sovereign.go
```diff
@@ -131,10 +134,21 @@ func (d *downloaderSovereign) buildAppender(
 			return fmt.Errorf("error parsing UpdateHashChainValue event log %+v: %w", l, err)
 		}
 
+		// L1InfoTreeSync is up to date, proceed with GER lookup
 		l1InfoTreeLeaf, err := d.l1InfoTreeSync.GetInfoByGlobalExitRoot(insertGEREvent.NewGlobalExitRoot)
 		if err != nil {
-			log.Fatalf("GER %s received from L2 is not present in L1InfoTreeSync: %v",
-				insertGEREvent.NewGlobalExitRoot, err)
+			ctx := context.Background()
+			isUpToDate, err := d.isL1InfoTreeSyncUpToDate(ctx)
```
goran-ethernal: Could this case be a false negative if, for example, `l1 rpc` node is not available?

rachit77: if` l1 rpc `node is not available than `isUpToDate` will be `false` and `l2gersync` will be continue functioning normally as it used to do before adding this logic

goran-ethernal (approved):

l1infotreesync/l1infotreesync.go
```diff
@@ -322,3 +323,22 @@ func (s *L1InfoTreeSync) GetProcessedBlockUntil(ctx context.Context, blockNum ui
...
+	}
+
+	lastProcessedBlock, err := s.processor.GetLastProcessedBlock(ctx)
+	if err != nil {
+		return false, fmt.Errorf("failed to get last processed block: %w", err)
+	}
+
+	finalizedBlock, err := l1Client.BlockByNumber(ctx, big.NewInt(int64(rpc.FinalizedBlockNumber)))
+	if err != nil {
+		return false, fmt.Errorf("failed to get finalized L1 block: %w", err)
```
Stefan-Ethernal:
```suggestion
		return false, fmt.Errorf("failed to get the latest finalized L1 block: %w", err)
```

temaniarpit27 (approved):

Stefan-Ethernal (approved):

PR #1053, opened 2025-09-24
feat: include `global_index` to bridge responses

### LINKED ISSUE #1048: feat: expose global index to bridges endpoint
In order to be able to map the certain bridge with the claim, we should expose the `global_index` field as part of bridge and claim response. Also we should be able to filter by this field the `claims` endpoint.

TBD
Should we:
- persist it in a database (slightly more performant, but requires reindexing),
- calculate it on fly, and set only in a response.

Stefan-Ethernal:
## 🔄 Changes Summary
- Include `global_index` to bridge responses (it is calculated on fly and added to the response only, but not persisted in the db).
- Provide a way to filter the claims based on the `global_index`.

## ⚠️ Breaking Changes
N/A

## 📋 Config Updates
N/A

## ✅ Testing
- 🤖 **Automatic**: CI checks

## 🐞 Issues
- Closes #1048


bridgeservice/bridge.go
```diff
@@ -408,6 +418,12 @@ func (b *BridgeService) GetClaimsHandler(c *gin.Context) {
 		}
 	}
 
+	globalIndexRaw := c.Query(globalIndexParam)
+	var globalIndex *big.Int
+	if globalIndexRaw != "" {
+		globalIndex, _ = new(big.Int).SetString(globalIndexRaw, 0)
```
temaniarpit27: i think we should check for err here, if the passed value is not int

temaniarpit27 (approved):
LGTM

bridgeservice/bridge.go
Stefan-Ethernal: Agree, fixed https://github.com/agglayer/aggkit/pull/1053/commits/8105b30eb03feb1b8941a382fcf1264faeef06fa

rachit77 (approved):
LGTM

PR #1055, opened 2025-09-24
feat: txn sender

### LINKED ISSUE #1054: add tx_sender to bridges
The apps team required to add tx_sender field (the one who initiated the transaction, namely the root level call from address) to both bridges response.

temaniarpit27:
## 🔄 Changes Summary
- add `tx_sender` field in bridges and claims APIs

## 🐞 Issues
- Closes #1054

## ⚠️ Breaking Changes
- Need resyncing of bridgel1 and bridgel2 DBs


bridgesync/migrations/bridgesync0005.sql
```diff
@@ -0,0 +1,8 @@
+-- +migrate Down
+ALTER TABLE bridge DROP COLUMN tx_sender;
+ALTER TABLE claim DROP COLUMN tx_sender;
+
+-- +migrate Up
+ALTER TABLE bridge ADD COLUMN tx_sender VARCHAR;
+ALTER TABLE claim ADD COLUMN tx_sender VARCHAR;
```
Stefan-Ethernal: Should we make it non-nullable?

```suggestion
ALTER TABLE bridge ADD COLUMN tx_sender VARCHAR NOT NULL;
ALTER TABLE claim ADD COLUMN tx_sender VARCHAR NOT NULL;
```

bridgeservice/docs/docs.go
```diff
@@ -637,6 +637,11 @@ const docTemplate = `{
                     "description": "Hash of the transaction that included the bridge event",
                     "type": "string",
                     "example": "0xdef4567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef"
+                },
+                "tx_sender": {
+                    "description": "Address of the transaction sender who initiated the bridge transaction",
```
rachit77, line 648: The distinction between `from_address` and `tx_sender` isn’t immediately clear from the descriptions.

temaniarpit27: fixed

bridgesync/downloader.go
```diff
@@ -375,46 +418,35 @@ func findCall(rootCall call, targetAddr common.Address, callback func(call) (boo
...
-	err := client.Call(callFrame, debugTraceTxEndpoint, txHash, tracerCfg{Tracer: callTracerType})
-	if err != nil {
-		return err
-	}
-
 	// Check if the root call was successful
-	if callFrame.Err != nil {
-		return fmt.Errorf("execution reverted in root call (block %d, tx %s): %s", c.BlockNum, txHash.Hex(), *callFrame.Err)
+	if rootCall.Err != nil {
+		return fmt.Errorf("execution reverted in root call (block %d): %s", c.BlockNum, *rootCall.Err)
```
rachit77: we are losing the txn hash in this error path
How about wrapping it at the top level of `setClaimCalldataFromRoot` so that the error includes both the block number and the tx hash

bridgeservice/docs/docs.go
```diff
@@ -609,7 +609,7 @@ const docTemplate = `{
                     "example": 42161
                 },
                 "from_address": {
-                    "description": "Address that initiated the bridge transaction",
+                    "description": "Address that initiated the txn on bridge contract. It can be intermediary contract or EOA",
```
rachit77:
```suggestion
                    "description": "Address that initiated the transaction on bridge contract. It can be intermediary contract or EOA",
```

I think we should use `transaction` instead of `txn` in docs for the sake of consistency

rachit77 (approved):

bridgeservice/docs/docs.go
temaniarpit27: done

rachit77:
just on a side note we can also edit some [internal-claims.bats](https://github.com/agglayer/e2e/blob/main/tests/aggkit/internal-claims.bats) tests to verify the changes

temaniarpit27:
> just on a side note we can also edit some [internal-claims.bats(https://github.com/agglayer/e2e/blob/main/tests/aggkit/internal-claims.bats) tests to verify the changes

done already. tested locally

bridgesync/downloader.go
temaniarpit27: ideally it should be check after rootcall fetching only. made changes

bridgesync/downloader.go
```diff
@@ -275,7 +319,14 @@ func buildMigrateLegacyTokenHandler(contract *bridgel2sovereignchain.Bridgel2sov
...
 
-		foundCall, err := extractCall(client, bridgeAddr, l.TxHash, logger)
+		// Extract root call first to get calldata
+		rootCall, err := extractRootCall(client, bridgeAddr, l.TxHash)
+		if err != nil {
+			return fmt.Errorf("failed to extract the MigrateLegacyToken event root call (tx hash: %s): %w", l.TxHash, err)
+		}
+
+		// Extract the specific call to the bridge contract
+		foundCall, err := findCall(*rootCall, bridgeAddr, nil, logger)
```
Stefan-Ethernal, line 331: Would it make sense that the `findCall` in case receives nil `rootCall` (and in that case it first extracts the root call)? That way, we would not be forced to invoke the 2 functions explicitly. WDYT?

bridgesync/downloader.go
```diff
@@ -108,7 +108,14 @@ func buildBridgeEventHandler(contract *polygonzkevmbridgev2.Polygonzkevmbridgev2
 			return fmt.Errorf("error parsing BridgeEvent log %+v: %w", l, err)
 		}
 
-		foundCall, err := extractCall(client, bridgeAddr, l.TxHash, logger)
+		// Extract root call first to get both TxSender and calldata
+		rootCall, err := extractRootCall(client, bridgeAddr, l.TxHash)
+		if err != nil {
+			return fmt.Errorf("failed to extract bridge event root call (tx hash: %s): %w", l.TxHash, err)
+		}
+
```
Stefan-Ethernal, line 118: We are not checking if the `rootCall` is reverted or not, like we do [here](https://github.com/agglayer/aggkit/blob/c9f3d3bbccbbf20e8464717d18ec9f50a1d650f3/bridgesync/downloader.go#L171-L174)

PR #1057, opened 2025-09-25
feat: retain policy

### LINKED ISSUE #253: aggsender: improve policy of purge old certificates from DB
# Feature Request
- Currently you can choose to keep inError certificates: `KeepCertificatesHistory` 
-  Choose how many certificates to keep on DB

joanestebanr:
## 🔄 Changes Summary
- Allow to choose the retain policy for certificates: how many keep in database

## ⚠️ Breaking Changes
- 🛠️ **Config**: 
- Deprecated field `AggSender.KeepCertificatesHistory` in favour of a new struct: 
```
	[AggSender.StorageRetainCertificatesPolicy]
		RetainCertificatesCount = 0 # 0 means keep all certificates
		KeepCertificatesHistory = true
```

## 📋 Config Updates
- Add new retain policy configuration: 
- *RetainCertificatesCount*: How many certs, 1 (means current cert, and another), 0 means all certificates
- *KeepCertificatesHistory*: that have been moved from `AggSender.KeepCertificatesHistory` it allow to keep track of retries
```
	[AggSender.StorageRetainCertificatesPolicy]
		RetainCertificatesCount = 0 # 0 means keep all certificates
		KeepCertificatesHistory = true
```

## 🐞 Issues
- Closes #253


aggsender/db/aggsender_db_storage.go
```diff
@@ -46,16 +57,25 @@ func (r RuntimeData) IsCompatible(storage RuntimeData) error {
 	return nil
 }
 
+type AggSenderStorageMaintenancer interface {
```
goran-ethernal:
```suggestion
type AggSenderStorageMaintainer interface {
```

aggsender/db/aggsender_db_storage.go
```diff
@@ -46,16 +57,25 @@ func (r RuntimeData) IsCompatible(storage RuntimeData) error {
...
+	// Move to certificate_info_history the certificate identified by CertificateKey
+	MoveCertificateToHistory(tx dbtypes.Querier, height uint64) error
+	// Delete from certificate_info and certificate_info_history the certificate CertificateKey
```
goran-ethernal:
```suggestion
	// Delete from certificate_info and certificate_info_history the certificate identified by CertificateKey
```

aggsender/db/aggsender_db_storage.go
```diff
@@ -324,17 +378,17 @@ func (a *AggSenderSQLStorage) SaveLastSentCertificate(ctx context.Context, certi
...
+	if err := a.handleCertificateFile(certInfo); err != nil {
+		return err
 	}
 
-	if certInDB != nil {
-		// we already have a certificate with this height
-		// we need to delete it before inserting the new one
-		if err = a.moveCertificateToHistoryOrDelete(tx, certInDB); err != nil {
-			return fmt.Errorf("saveLastSentCertificate moveCertificateToHistory Err: %w", err)
+	if a.retainPolicy != nil {
```
goran-ethernal: When can this be nil? We always have a default retain policy which is not delete anything, so I would say for the sake of code simplicity, I would remove this condition, and always go through `retainPolicy`.

aggsender/db/aggsender_db_storage.go
```diff
@@ -597,11 +580,77 @@ func (a *AggSenderSQLStorage) GetNonAcceptedCertificate() (*NonAcceptedCertifica
...
+		height); err != nil {
+		return fmt.Errorf("error moving certificate height: %d to history: %w", height, err)
+	}
+	return a.deleteCertificates(tx, tableCertificate, height)
+}
+
+// Delete from certificate_info and certificate_info_history the certificate CertificateKey
+// if you don't need a tx just pass nil
+// It required to be in certificate_info table, if not found it returns ErrNoCertDeleted error
+func (a *AggSenderSQLStorage) DeleteCertificate(tx dbtypes.Querier, height uint64, mustDelete bool) error {
```
goran-ethernal: Why not use the delete flag type here?

aggsender/db/aggsender_db_storage.go
```diff
@@ -597,11 +580,77 @@ func (a *AggSenderSQLStorage) GetNonAcceptedCertificate() (*NonAcceptedCertifica
 	return &nonAcceptedCert, nil
 }
 
+// Move to certificate_info_history the certificate identified by CertificateKey
+func (a *AggSenderSQLStorage) MoveCertificateToHistory(tx dbtypes.Querier, height uint64) error {
+	a.logger.Debugf("moving certificate to history - height: %d", height)
+	if _, err := tx.Exec(`INSERT INTO certificate_info_history SELECT * FROM certificate_info WHERE height = $1;`,
```
goran-ethernal, line 588: As far as I remember, we were updating the retry count of the certificate before moving it to the history table, plus, history table has a `height + retry_count` primary key, so we were able to store same certificates with different retry counts in the history table. I don't see that done here. Maybe I am wrong, and it is done in a different function.

aggsender/db/aggsender_db_storage.go
```diff
@@ -179,10 +211,38 @@ func getCertificateByHeight(db dbtypes.Querier,
...
+
+// getCertificatesHeightOlderThanHeight returns a list of certificate heights older than the provided height
+func getCertificatesHeightOlderThanHeight(db dbtypes.Querier, table tableName,
+	olderThanHeight uint64) ([]uint64, error) {
+	type heightRow struct {
+		Height uint64 `meddler:"height"`
+	}
+	var rows []*heightRow
+	if err := meddler.QueryAll(db, &rows,
+		fmt.Sprintf("SELECT DISTINCT height FROM %s WHERE height < $1;", table), olderThanHeight); err != nil {
```
goran-ethernal: If we say `DISTINCT` what happens to certificates that had multiple retry counts and are saved in the history table?

aggsender/db/aggsender_db_storage.go
joanestebanr: You're right!, I would left only a check of `a.retainPolicy` that if it's nil it returns an error

aggsender/db/aggsender_db_storage.go
joanestebanr: oops!, I forget to change it! thanks you!

aggsender/db/aggsender_db_storage.go
joanestebanr: This function returnts only the heights stored in the table, that is the reason of the usage of `DISTINCT`

aggsender/db/aggsender_db_storage.go
joanestebanr: I double check just in case, and I confirm that each certificate is created with `height` , `retry_count` assigned at the beginning. We never change it, basically the first is `retry_count`=0, next is 1.. and so on. The certificates in main table (`certificate_info`) also have the `retry_count` to the real retry number. So we can move it without modifiying the value.
The SQL is exactly the same as before: https://github.com/agglayer/aggkit/blob/c53488ffeb3f723d996ce27ba2dcd12f708a8403/aggsender/db/aggsender_db_storage.go#L356

goran-ethernal: Cool. If we do not have a UT that tests this, it would be nice to add it.

aggsender/db/aggsender_db_storage.go
```diff
@@ -133,17 +154,28 @@ func (a *AggSenderSQLStorage) GetCertificateHeadersByStatus(
...
 	}
 
 	// Add ordering by creation date (oldest first)
-	query += " ORDER BY height ASC"
+	condition += " ORDER BY height ASC"
 
+	return a.getAllCerts(nil, tableCertificate, condition, args)
+}
+
+func (a *AggSenderSQLStorage) getAllCerts(tx dbtypes.Querier, table tableName,
```
goran-ethernal: `getAllCerts` implies you want to get all the certs from `db`. This makes more sense to me.

```suggestion
func (a *AggSenderSQLStorage) getCerts(tx dbtypes.Querier, table tableName,
```

aggsender/db/aggsender_db_storage.go
```diff
@@ -324,18 +378,15 @@ func (a *AggSenderSQLStorage) SaveLastSentCertificate(ctx context.Context, certi
...
+		return err
 	}
 
-	if certInDB != nil {
-		// we already have a certificate with this height
-		// we need to delete it before inserting the new one
-		if err = a.moveCertificateToHistoryOrDelete(tx, certInDB); err != nil {
-			return fmt.Errorf("saveLastSentCertificate moveCertificateToHistory Err: %w", err)
-		}
+	if a.retainPolicy == nil {
```
goran-ethernal: This can never be nil as per current code.

aggsender/db/aggsender_db_storage.go
```diff
@@ -597,11 +577,77 @@ func (a *AggSenderSQLStorage) GetNonAcceptedCertificate() (*NonAcceptedCertifica
...
+
+// Delete from certificate_info and certificate_info_history all certificates older than olderThanHeight
+func (a *AggSenderSQLStorage) DeleteOldCertificates(tx dbtypes.Querier, olderThanHeight uint64) error {
+	// We get list of heights from certificate_info table,  we assume that
+	// history table require the same heights into certificate_info table
+	heights, err := getCertificatesHeightOlderThanHeight(tx, tableCertificate, olderThanHeight)
+	if err != nil {
+		return fmt.Errorf("error getting old certificate heights: %w", err)
+	}
+	for _, height := range heights {
```
goran-ethernal: I am thinking if there is a simpler way to do this, and not have `n` queries. Since we are using `MaybeDelete`, I think we can just call `DELETE from table where height < olderThanHeight` for both `certificate_info` and `certificate_info_history` tables. And for files, we can write a simple function that will remove all files that contain height with lower number than provided, since all files are named `signed_cert_<height>_<certificateID>_<retry>.json`.

This is also kind of confusing, we first get all certificates that are older than given height, then we call `DeleteCertificates` where we again get the same certificate from both tables before actually removing them. I know this is done because of the files, but it is kind of confusing. It makes more sense to go with the point in this approach, I think it is simpler:

config/default.go
```diff
@@ -245,6 +244,9 @@ RequireValidatorCall = false
 		MinConnectTimeout = "5s"
 		RequestTimeout = "30s"
 		UseTLS = false
+	[AggSender.StorageRetainCertificatesPolicy]
+		RetainNCertificates = 0 # 0 means keep all certificates
```
Stefan-Ethernal: I find it slightly more readable, WDYT?
```suggestion
		RetainCertificatesCount = 0 # 0 means keep all certificates
```

aggsender/db/retain_certificates_policy.go
```diff
@@ -0,0 +1,98 @@
...
+//
+// So there are two policies to retain certificates:
+//   1. retain N (>1 or all) cert and all retries (currently: KeepCertificatesHistory=true)
+//   2. retain N (>1 or all) cert for each height no retries (currently: KeepCertificatesHistory=false)
+// Two params:
+//   - How many certificates to retain
+//   - keep history of retries or not
+//
+// Another ideas:
+//   - by size: that could be another of configuring that but maybe is not needed
```
Stefan-Ethernal, line 29: Why don't we create an issue on the backlog for this and evaluate whether we are going to implement it?

aggsender/db/retain_certificates_policy.go
```diff
@@ -0,0 +1,98 @@
...
+// 		- save(height 10 / retry 1) -> move (height 10 / retry 0) INTO certificate_info_history
+//                                  -> insert (height 10 / retry 1) INTO certificate_info
+//      - save(height 11 / retry 0) -> insert (height 11 / retry 0) INTO certificate_info
+//
+// So there are two policies to retain certificates:
+//   1. retain N (>1 or all) cert and all retries (currently: KeepCertificatesHistory=true)
+//   2. retain N (>1 or all) cert for each height no retries (currently: KeepCertificatesHistory=false)
+// Two params:
+//   - How many certificates to retain
+//   - keep history of retries or not
```
Stefan-Ethernal, line 26: It would be beneficial explaining this in the docs as well. WDYT?

aggsender/db/aggsender_db_storage.go
```diff
@@ -86,17 +106,18 @@ var _ AggSenderStorage = (*AggSenderSQLStorage)(nil)
 
 // AggSenderSQLStorageConfig is the configuration for the AggSenderSQLStorage
 type AggSenderSQLStorageConfig struct {
-	DBPath                  string
-	CertificatesDir         string
-	KeepCertificatesHistory bool
+	DBPath                   string
+	CertificatesDir          string
+	RetainCertificatesPolicy StorageRetainCertificatesPolicy
```
Stefan-Ethernal, line 111: WDYT to use interface here as well? It seems to me that it is doable, because this struct is not used to unmarshal the TOML values to it. WDYT?
```suggestion
	RetainCertificatesPolicy StorageRetainCertificatesPolicier
```

aggsender/db/aggsender_db_storage.go
Stefan-Ethernal: Also AFAIK, the `tx` param is always nil, so maybe we can simplify the function signature, by removing it?
```suggestion
func (a *AggSenderSQLStorage) getAllCerts(table tableName,
```

aggsender/db/aggsender_db_storage.go
```diff
@@ -22,14 +22,25 @@ import (
...
+	tableCertificateHistory tableName = "certificate_info_history"
+
+	MustDelete  DeleteFlag = true  // the delete action must affect at least one row
+	MaybeDelete DeleteFlag = false // the delete action may affect zero rows
 )
 
-var newTxer = db.NewTx
+var (
+	newTxer          = db.NewTx
+	ErrNoCertDeleted = errors.New("no cert deleted")
```
Stefan-Ethernal:
```suggestion
	ErrNoCertDeleted = errors.New("no certificates deleted")
```

aggsender/db/aggsender_db_storage.go
```diff
@@ -179,10 +211,38 @@ func getCertificateByHeight(db dbtypes.Querier,
 		"SELECT * FROM certificate_info WHERE height = $1;", height); err != nil {
 		return nil, getSelectQueryError(height, err)
 	}
-
 	return &certificateInfo, nil
 }
 
+// getCertificatesByHeight returns a certificates in the required table
+func getCertificatesByHeight(db dbtypes.Querier, table tableName,
```
Stefan-Ethernal, line 218: Consider making this a method, so then you can remove the `db` param and reuse the `db` reference from the instance. WDYT?
```suggestion
func (a *AggSenderSQLStorage) getCertificatesByHeight(table tableName,
```

aggsender/db/aggsender_db_storage.go
```diff
@@ -179,10 +211,38 @@ func getCertificateByHeight(db dbtypes.Querier,
...
+	var certificates []*certificateInfo
+	if err := meddler.QueryAll(db, &certificates,
+		fmt.Sprintf("SELECT * FROM %s WHERE height = $1;", table), height); err != nil {
+		return nil, err
+	}
+	return certificates, nil
+}
+
+// getCertificatesHeightOlderThanHeight returns a list of certificate heights older than the provided height
+func getCertificatesHeightOlderThanHeight(db dbtypes.Querier, table tableName,
```
Stefan-Ethernal, line 229: Consider making this a method, so then you can remove the db param and reuse the db reference from the instance. WDYT?
```suggestion
func (a *AggSenderSQLStorage) getCertificatesHeightOlderThanHeight(table tableName,
```

aggsender/db/types.go
```diff
@@ -37,21 +52,51 @@ type certificateInfo struct {
...
+
+// SignedCertificateFilename returns the filename if SignedCertificate is a file reference
+func (c *certificateInfo) SignedCertificateFilename() *string {
+	if c == nil || c.SignedCertificate == nil {
+		return nil
+	}
+	if strings.HasPrefix(*c.SignedCertificate, PrefixFilename) {
+		filename := strings.TrimPrefix(*c.SignedCertificate, PrefixFilename)
+		return &filename
+	}
```
Stefan-Ethernal, line 75: The suggestion came from the IDE, I'd say pretty neat replacement. WDYT?
```suggestion
	if after, ok := strings.CutPrefix(*c.SignedCertificate, PrefixFilename); ok {
		filename := after
		return &filename
	}
```

aggsender/db/types.go
```diff
@@ -37,21 +52,51 @@ type certificateInfo struct {
...
+	return nil
```
Stefan-Ethernal, line 76: Would it be ok to return here the string (empty string would denote the missing value, instead of nil). WDYT?
```suggestion
func (c *certificateInfo) SignedCertificateFilename() string {
	if c == nil || c.SignedCertificate == nil {
		return ""
	}
	if after, ok := strings.CutPrefix(*c.SignedCertificate, PrefixFilename); ok {
		filename := after
		return filename
	}
	return ""
```

aggsender/db/aggsender_db_storage.go
Stefan-Ethernal: Also maybe it doesn't make sense parameterizing the `tableName`, as it is always the `certificate_info` at the moment. WDYT?

aggsender/db/aggsender_db_storage.go
```diff
@@ -597,11 +577,77 @@ func (a *AggSenderSQLStorage) GetNonAcceptedCertificate() (*NonAcceptedCertifica
...
+		a.deleteCertificateFile(certInfo.SignedCertificateFilename())
+	}
+	if _, err := tx.Exec(fmt.Sprintf(`DELETE FROM %s WHERE height = $1;`, table), height); err != nil {
+		return fmt.Errorf("error deleting certificate history height %d: %w", height, err)
+	}
+	return nil
+}
+
+// Delete from certificate_info and certificate_info_history all certificates older than olderThanHeight
+func (a *AggSenderSQLStorage) DeleteOldCertificates(tx dbtypes.Querier, olderThanHeight uint64) error {
```
Stefan-Ethernal: WDYT, slightly shorter but still understandable (or even `cutoffHeight`)?
```suggestion
func (a *AggSenderSQLStorage) DeleteOldCertificates(tx dbtypes.Querier, maxHeight uint64) error {
```

aggsender/db/aggsender_db_storage.go
joanestebanr: Yes, there are plenty of this here `aggsender/db/retain_certificates_policy_test.go` that cover almost all possible configurations

aggsender/db/aggsender_db_storage.go
joanestebanr: > getAllCerts implies you want to get all the certs from db. This makes more sense to me.

done 093dd566d9a5c83acfbfb434ee8ba8f972390cad

> Also AFAIK, the tx param is always nil, so maybe we can simplify the function signature, by removing it?

I prefer to keep it, because IMO, all functions that execute a SQL must receive a `tx` to keep a coherence with the multiples queries

aggsender/db/aggsender_db_storage.go
joanestebanr: I agree, but is never checked that is !=nil, so,  if for some unexpected reason, it's nil it's better returns a  error instead that a crash with a panic. I think that is not increasing the complexity of the code, becuase is really a simple check, and neither the performace. WDTY?

goran-ethernal: But, you already have a validation that retain policy can not be nil: https://github.com/agglayer/aggkit/blob/ac88f10eb499872c950f2b42fb5f19d36919f361/aggsender/db/retain_certificates_policy.go#L40-L45.

So in my opinion this check is redundant.

aggsender/db/aggsender_db_storage.go
joanestebanr: The name of the file depend of the contents of DB (you can change configuration, or by code you can change the pattern of the filename). 
I propose to do next: 
- Do a select with all filenames to be removed
- Remove it
- delete all certs `height < olderThanHeight` with 1 sentence
WDYT?

aggsender/db/aggsender_db_storage.go
joanestebanr: You are checking the config but not the object `AggSenderSQLStorage`, anyway, if you have a strong opinion about that I can remove, for sure. Done 5fba9e20677a8ebe051b15e32dc1bec28b9966cc

config/default.go
joanestebanr: done: 9d03f6d5ce83086a5f7c2c24a934bf1b42a59486

aggsender/db/retain_certificates_policy.go
joanestebanr: In fact, it exists #253 since. It was created long time ago

aggsender/db/aggsender_db_storage.go
joanestebanr: I'm on the fence. This struct is a config, so I expect to have key value, not the behaviour (interface) that is not used.  By other hand in the object  `AggSenderSQLStorage` I want the behaviour, so I use the interface. WDYT?

aggsender/db/aggsender_db_storage.go
joanestebanr: I did in that way because pre-existing helper function were done in that way https://github.com/agglayer/aggkit/blob/9d03f6d5ce83086a5f7c2c24a934bf1b42a59486/aggsender/db/aggsender_db_storage.go#L207
I agree, that I prefer doing as a member of the object, but that implies refactor several function to match the same criteria. Also you can't save the param `db dbtypes.Querier` because is required to pass the transaction
WDYT?

aggsender/db/aggsender_db_storage.go
joanestebanr: > Consider making this a method, so then you can remove the db param and reuse the db reference from the instance. WDYT?

Same response than before as before

>Also maybe it doesn't make sense parameterizing the tableName, as it is always the certificate_info at the moment. WDYT?

In fact, with the refactor sugestion here https://github.com/agglayer/aggkit/pull/1057#discussion_r2391220489 I would need to call also in history table to get the filenames of the certificates to delete

aggsender/db/types.go
joanestebanr: For me is more clear the current version

aggsender/db/types.go
joanestebanr: What is the benefit of that? If you prefer I can return a `(string, bool)` that is more idiomatic, WDYT?

aggsender/db/retain_certificates_policy.go
joanestebanr: done: 5fc6ca7a5af9ca674574af82cb1c463f287c8604

aggsender/db/retain_certificates_policy.go
Stefan-Ethernal: I thought about this idea for further improvement, configuring the certificates retain policy by its size...

aggsender/db/types.go
Stefan-Ethernal: We are avoiding de-referencing the pointer, which I find strange tbh (especially in case of string, where it is completely fine using the empty string to denote the missing value).

aggsender/db/aggsender_db_storage.go
Stefan-Ethernal: I'm fine with either, the reason why I've suggested it, is because we:
- return a pointer when constructing the default retain policy [here](https://github.com/agglayer/aggkit/blob/5fba9e20677a8ebe051b15e32dc1bec28b9966cc/aggsender/db/retain_certificates_policy.go#L65-L70)
- then we do the [de-referencing](https://github.com/agglayer/aggkit/blob/5fba9e20677a8ebe051b15e32dc1bec28b9966cc/aggsender/validator/validate_db.go#L37)

I don't see a reason to do it. If it needs to be a value type, then constructor function should not return a pointer.

If we switch to use the interface, then we are consistent (as we don't need to de-reference the pointer). WDYT?

aggsender/db/aggsender_db_storage.go
joanestebanr: done: 64bcafc42ee62ce8da13720c5afc2e92b2421925

goran-ethernal (approved):

Stefan-Ethernal (approved):

aggsender/db/retain_certificates_policy.go
joanestebanr: I also take in consideration it but is really complicated to implement (since we move the certificates to files)  and may be the use would  want the full history or with just 10 cert is more that enought for debug

PR #1060, opened 2025-09-29
feat: add `aggchain params` verification

goran-ethernal:
## 🔄 Changes Summary
This PR adds verification of the `aggchain params` field on the `FEP` certificates.

It also separates the `builder` and `verifier` flows (`PP` and `FEP`).
`Builder` flows are used in `aggsender-proposer` to build certificates.
`Verifier` flows are used in `aggsender-validator` to verify certificates. Note here, that validator also needs to build certificates, so `verifier` flows `inherit` the builder flows as well. This makes them inherit their `build certificate` functions, and common fields which are needed in the `verifier` as well.

## ⚠️ Breaking Changes
NA

## 📋 Config Updates
Added new parameter to `Validator.FEPConfig` called `OpNodeURL` which is the URL to the op l2 node used for retreiving op data for FEP certificates:

```toml
[Validator.FEPConfig]
    OpNodeURL = "{{OpNodeURL}}"
```

## ✅ Testing

## 🐞 Issues
- Closes #1060


aggsender/flows/factory.go
```diff
@@ -80,7 +80,7 @@ func NewFlow(
 		}
 
 		aggchainFEPQuerier, err := query.NewAggchainFEPQuerier(logger, types.AggchainProofMode,
-			cfg.SovereignRollupAddr, l1Client)
+			cfg.SovereignRollupAddr, cfg.OptimisticModeConfig.OpNodeURL, l1Client)
```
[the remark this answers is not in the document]
goran-ethernal: This is on the instantiation of flow inside the `aggsender-proposer`, the `op node url` is in the `OptimisticModeConfig` in the proposer.

aggsender/validator/validate_certificate.go
```diff
@@ -119,6 +115,16 @@ func (a *CertificateValidator) ValidateCertificate(ctx context.Context, params t
 		return fmt.Errorf("failed to verify claim proofs: %w", err)
 	}
 
+	// Verify AggchainData specific to each flow
+	if err := a.flow.VerifyAggchainData(
```
joanestebanr: This part is not generic, maybe the function `VerifyAggchainData` must not be part of the common interface. The same as `aggchainFEPQuerier` that maybe must be part of the flow not of the generic validator. WDTY?

aggsender/optimistic/optimistic_aggregation_proof_public_values_query.go
```diff
@@ -43,35 +44,60 @@ func NewOptimisticAggregationProofPublicValuesQuery(
...
 		ClaimRoot:        claimRoot,
 		L2BlockNumber:    requestedEndBlock,
 		RollupConfigHash: rollupConfigHash,
 		MultiBlockVKey:   multiBlockVKey,
-		ProverAddress:    o.proverAddress,
+		ProverAddress:    trustedSequencerAddr,
 	}, nil
 }
+
+func (o *OptimisticAggregationProofPublicValuesQuery) trustedSequencerAddr() (common.Address, error) {
```
joanestebanr: This code is duplicated here: https://github.com/agglayer/aggkit/blob/c5b7d013bf0bbba08248f9a7cc03295dd6375302/aggsender/optimistic/optimistic_sign.go#L73

aggsender/validator/validate_certificate.go
goran-ethernal: Yeah, I was thinking about this, should be part of the flows, or should it just be a part of the validator. I would perhaps just move this verification to the validator itself. WDYT?

aggsender/query/aggchain_fep_rollup_query.go
```diff
@@ -64,6 +72,7 @@ func NewAggchainFEPQuerier(
 	logger *log.Logger,
 	aggsenderMode types.AggsenderMode,
 	aggchainFEPAddr common.Address,
+	opNodeURL string,
```
Stefan-Ethernal: WDYT, recieve the config instead of `aggchainFEPAddr` and `opNodeURL` (if possible due to imports)?

```suggestion
	cfg validator.FEPConfig,
```

aggsender/validator/validate_certificate.go
Stefan-Ethernal: Is it possible to move the `VerifyAggchainData` to the `flow_aggchain_prover` and make it internal to it?

goran-ethernal: Ok, I've made it more generic: https://github.com/agglayer/aggkit/pull/1060/commits/8d3d55a1e449795e9da6dc6248bba836b9e5ff06

aggsender/optimistic/optimistic_aggregation_proof_public_values_query.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/1060/commits/04bc4d017633e56d7a1d79f2e6ad2c38576051b8

aggsender/flows/flow_aggchain_prover.go
```diff
@@ -357,6 +360,58 @@ func (a *AggchainProverFlow) UpdateAggchainData(
...
+	aggchainDataProof, ok := cert.AggchainData.(*agglayertypes.AggchainDataProof)
+	if !ok {
+		// If the AggchainData is of type MultisigWithProof, we extract the proof and verify it
+		aggchainDataMultisigWithProof, ok := cert.AggchainData.(*agglayertypes.AggchainDataMultisigWithProof)
+		if !ok {
+			return fmt.Errorf("aggchainProverFlow: certificate AggchainData is of unknown type %T", cert.AggchainData)
+		}
+
+		aggchainDataProof = aggchainDataMultisigWithProof.AggchainProof
+	}
```
Stefan-Ethernal: IMO, it is cleaner this way

```suggestion
	var aggchainDataProof *agglayertypes.AggchainDataProof

	switch v := cert.AggchainData.(type) {
	case *agglayertypes.AggchainDataProof:
		aggchainDataProof = v
	case *agglayertypes.AggchainDataMultisigWithProof:
		aggchainDataProof = v.AggchainProof
	default:
		return fmt.Errorf("aggchainProverFlow: certificate AggchainData is of unknown type %T", cert.AggchainData)
	}
```

aggsender/optimistic/optimistic_aggregation_proof_public_values_query.go
```diff
@@ -43,35 +47,59 @@ func NewOptimisticAggregationProofPublicValuesQuery(
...
 		ClaimRoot:        claimRoot,
 		L2BlockNumber:    requestedEndBlock,
 		RollupConfigHash: rollupConfigHash,
 		MultiBlockVKey:   multiBlockVKey,
-		ProverAddress:    o.proverAddress,
+		ProverAddress:    trustedSequencerAddr,
 	}, nil
 }
+
+func getTrustedSequencerAddr(aggchainFEPContract FEPContractQuerier) (common.Address, error) {
```
Stefan-Ethernal: IMO it would be better to name it as `trustedSigner`, because the trustedSequencer dates from the zkevm days, when we had the L2 sequencer that was trusted entity... But now the semantic has changed.

aggsender/query/aggchain_fep_rollup_query.go
goran-ethernal: It can't because it is instantiated from both regular and validator aggsender which have different config structs. This is ok in my opinion, it doesn't have too much parameters in the constructor.

aggsender/flows/flow_aggchain_prover.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/1060/commits/91fd403435cdc6ba89b32dd03edeec8265e20ce2

aggsender/optimistic/optimistic_sign.go
```diff
@@ -76,7 +76,7 @@ func validateSignerAgainstContract(
 	signerAddr common.Address,
 	requireKeyMatch bool,
 ) (common.Address, error) {
-	signers, err := contract.GetAggchainSigners(nil)
+	trustedSequencerAddr, err := getTrustedSequencerAddr(contract)
```
Stefan-Ethernal: Nitpick, consider renaming

```suggestion
	trustedSignerAddr, err := getTrustedSignerAddr(contract)
```

aggsender/validator/validate_certificate.go
```diff
@@ -119,6 +112,15 @@ func (a *CertificateValidator) ValidateCertificate(ctx context.Context, params t
 		return fmt.Errorf("failed to verify claim proofs: %w", err)
```
Stefan-Ethernal, line 112: Should the `verifyClaimProofs` also be a part of `VerifyCertificate` function (it can be declared in the common flow for example), WDYT?

Stefan-Ethernal (approved):

aggsender/validator/validate_certificate.go
goran-ethernal: `verifyClaimProofs` is a common part of certificate verification, same for both flows, so I would only leave flow specific things in the `flow.VerifyCertificate` function.

aggsender/optimistic/optimistic_sign.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/1060/commits/f9b748c5519f66966982027024eb2ab79a880b39

aggsender/optimistic/optimistic_aggregation_proof_public_values_query.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/1060/commits/f9b748c5519f66966982027024eb2ab79a880b39

aggsender/aggsender.go
```diff
@@ -121,7 +121,7 @@ func New(
 	)
 
 	aggchainFEPCaller, err := query.NewAggchainFEPQuerier(logger, mode,
-		cfg.SovereignRollupAddr, l1Client)
+		cfg.SovereignRollupAddr, cfg.OptimisticModeConfig.OpNodeURL, l1Client)
```
joanestebanr: It's a bit strange that you use optimistic configuration here, because it suggest that is only required for optimistic but it's no longer true

aggsender/query/aggchain_fep_rollup_query.go
```diff
@@ -77,13 +86,22 @@ func NewAggchainFEPQuerier(
 			aggchainFEPAddr.String(), err)
 	}
 
-	return newAggchainFEPQuerier(logger, aggchainFEPAddr, aggChainFEPContract)
+	opPublicValuesQuerier := optimistic.NewOptimisticAggregationProofPublicValuesQuery(
```
joanestebanr: why do that? instead of using the object OptimisticAggregationProofPublicValuesQuery in the place that you need? 
Doing in that way you increase the dependencies of `AggchainFEPRollupQuerier` just for one case in the validator.

aggsender/query/l1info_tree_data_query.go
```diff
@@ -177,3 +177,16 @@ func (l *L1InfoTreeDataQuerier) getLatestProcessedFinalizedBlock(ctx context.Con
 		"Expected hash: %s, got: %s", lastProcessedBlockNum,
 		lastFinalizedL1Block.Hash().String(), lastProcessedBlockHash.String())
 }
+
+// GetInfoByIndex returns the L1 Info tree leaf for the given index
+func (l *L1InfoTreeDataQuerier) GetInfoByIndex(
+	ctx context.Context, index uint32) (*l1infotreesync.L1InfoTreeLeaf, error) {
```
joanestebanr, line 183: maybe make sense to return directly the obj intead of a pointer (`*l1infotreesync.L1InfoTreeLeaf`) because  if the pointer is `nil` you returns an error.

aggsender/query/aggchain_fep_rollup_query.go
goran-ethernal: As per our discussion, builder and verifier flows are now separated: https://github.com/agglayer/aggkit/pull/1060/commits/baa310e7942826f46f6542bf6865edad7653a678

aggsender/aggsender.go
goran-ethernal: As per our discussion, builder and verifier flows are now separated: https://github.com/agglayer/aggkit/pull/1060/commits/baa310e7942826f46f6542bf6865edad7653a678

aggsender/query/l1info_tree_data_query.go
goran-ethernal: That doesn't really make sense to me. The syncer returns a pointer, and if it can not find that index it will return an error, so I don't think we have to do anything here.

aggsender/query/aggchain_fep_rollup_query.go
```diff
@@ -32,14 +31,20 @@ func (n *noOpAggchainFEPRollupQuerier) IsFEP() bool {
 	return false
 }
 
+func (n *noOpAggchainFEPRollupQuerier) GetAggregationProofPublicValuesData(
```
joanestebanr, line 34: This object `noOpAggchainFEPRollupQuerier` is not longer required, isn't?

joanestebanr (approved):

aggsender/query/aggchain_fep_rollup_query.go
goran-ethernal: I just checked, it is used, because we instantiate the `aggchainFEPRollupQuerier` because of the validator, and because I didn't want to have the checks in code when calculating the end block of the last settled certificate for the mode of aggsender (PP, FEP), I created the `noOpQuerier` which is instantiated when mode is `PP`.

PR #1073, opened 2025-10-01
feat: fix e2e multisign FEP 

joanestebanr:
## 🔄 Changes Summary
- Fix e2e test
- Introduce a new RPC for L1InfoTreeSync to allow to debug 
- It add `mode` in the response of RPC API `aggsender_status` that is the aggsender mode: `PessimisticProof` or `AggchainProof`
- It relax the rateLimit of submit certificates to aggsender

## 📋 Config Updates


cmd/run.go
```diff
@@ -96,8 +96,11 @@ func start(cliCtx *cli.Context) error {
 
 	// Create WaitGroup for backfill goroutines synchronization
 	var backfillWg sync.WaitGroup
-
+	var rpcServices []jRPC.Service
 	l1InfoTreeSync := runL1InfoTreeSyncerIfNeeded(cliCtx.Context, components, *cfg, l1Client)
+	if l1InfoTreeSync != nil {
+		rpcServices = append(rpcServices, l1InfoTreeSync.GetRPCServices()...)
```
goran-ethernal, line 102: Why not add this to `runL1InfoTreeSyncerIfNeeded` function?

joanestebanr: Is the way that is done in the rest

l1infotreesync/l1infotreesync.go
```diff
@@ -281,6 +294,12 @@ func (s *L1InfoTreeSync) GetFirstInfo() (*L1InfoTreeLeaf, error) {
 	}
 	return s.processor.GetFirstInfo()
 }
+func (s *L1InfoTreeSync) GetInfoLeafByRoot(root common.Hash) (*L1InfoTreeLeaf, error) {
```
goran-ethernal: To be consistent with other function names that return the `L1InfoTreeLeaf``:

```suggestion
func (s *L1InfoTreeSync) GetInfoByRoot(root common.Hash) (*L1InfoTreeLeaf, error) {
```

aggsender/aggsender.go
```diff
@@ -352,11 +353,12 @@ func (a *AggSender) sendCertificate(ctx context.Context) (*agglayertypes.Certifi
 	if err != nil {
 		return nil, fmt.Errorf("error building certificate: %w", err)
 	}
-
-	if _, err := a.localValidator.ValidateAndSignCertificate(ctx, certificate, certificateParams.ToBlock); err != nil {
-		a.log.Errorf("error validating certificate locally: %w", err)
+	if a.cfg.RequireLocalValidatorCheck {
```
goran-ethernal: Why do we need this check since we only log in the case of an error on local validator?

joanestebanr: you're totally right!!!!, I got confused yesterday when I saw the error, I'm going to revert it and change the log.Error for a warning

.github/workflows/test-e2e.yml
```diff
@@ -196,11 +197,11 @@ jobs:
       - build-aggkit-image
       - read-aggkit-args
       - get-kurtosis-cdk-commit
-    uses: agglayer/e2e/.github/workflows/aggkit-e2e-single-chain.yml@25ece31a112cc6f48f2dada8fdaf69350eedb8dc
+    uses: agglayer/e2e/.github/workflows/aggkit-e2e-single-chain.yml@eb195bccd8bddbb5e2eba26b6274664f1417d44d
     secrets: inherit
     with:
       kurtosis-cdk-ref: ${{ needs.get-kurtosis-cdk-commit.outputs.kurtosis-commit }}
-      agglayer-e2e-ref: 25ece31a112cc6f48f2dada8fdaf69350eedb8dc
+      agglayer-e2e-ref: eb195bccd8bddbb5e2eba26b6274664f1417d44d # main <- feat/fix_e2e_multisig
```
Stefan-Ethernal, line 204: Wdyt to use this ref for the other e2e tests as well? IMO it is better to keep it aligned.

config/default.go
```diff
@@ -6,7 +6,7 @@ package config
 const DefaultMandatoryVars = `
 L1URL = "http://localhost:8545"
 L2URL = "http://localhost:8123"
-OpNodeURL = "http://localhost:8080"
+OpNodeURL = "OpNodeURL_must_be_defined_in_config_file"
```
Stefan-Ethernal: Why we can not we set an empty string here, since it is marked as a mandatory variable (hopefully it should error on the startup in case it is missing value). Wdyt?

Stefan-Ethernal (approved):

.github/workflows/test-e2e.yml
joanestebanr: becase right now is pointing to a branch, to be able to merge the e2e branch into `main` I need to merge this PR because I added that the status RPC call reponse with the mode of aggsender. 
So I'll need to change the ref to the final one, and then I agree, we can change all of them

config/default.go
joanestebanr: Yes, I agree, it's better... this was a fast fix to see it in logs... I would change it

joanestebanr: 98449af0c7f93364f1cf9f3fbb13575251412f84

l1infotreesync/l1infotreesync.go
joanestebanr: f131a1c3496af5da1e0efc4b952557970225fdd1

aggsender/optimistic/config.go
```diff
@@ -18,3 +20,13 @@ type Config struct {
 	// This is useful to ensure that the signer is the trusted sequencer, and not a random signer.
 	RequireKeyMatchTrustedSequencer bool `mapstructure:"RequireKeyMatchTrustedSequencer"`
 }
+
+func (c Config) Validate() error {
+	if c.SovereignRollupAddr == (ethCommon.Address{}) {
```
goran-ethernal:
```suggestion
	if c.SovereignRollupAddr == aggkitcommon.ZeroAddress {
```

goran-ethernal (approved):

PR #1075, opened 2025-10-02
feat: add txn_sender in bridges API

### LINKED ISSUE #1054: add tx_sender to bridges

temaniarpit27:
## 🔄 Changes Summary
- add tx_sender in bridges and claims API

## 🐞 Issues
- Closes #1054


bridgeservice/types/types.go
```diff
@@ -140,6 +140,9 @@ type BridgeResponse struct {
 
 	// Unique hash representing the bridge event, often used as an identifier
 	BridgeHash Hash `json:"bridge_hash" example:"0xabc1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcd"`
+
+	// Address of the transaction sender who initiated the bridge transaction
+	TxnSender Address `json:"txn_sender" example:"0xabc1234567890abcdef1234567890abcdef1234"`
```
Stefan-Ethernal: Adding one more hex digit, so that it is valid Ethereum address format

```suggestion
	TxnSender Address `json:"txn_sender" example:"0xabc1234567890abcdef1234567890abcdef12345"`
```

aggsender/validator/validate_db_test.go
```diff
@@ -83,7 +83,7 @@ func TestValidateFullAggsenderDB(t *testing.T) {
 		certificateValidator)
 
 	mockRollupDataQuerier.EXPECT().GetRollupData(mock.Anything).Return(
-		polygonrollupmanager.PolygonRollupManagerRollupDataReturn{}, nil)
+		polygonrollupmanager.PolygonRollupManagerRollupDataReturn{}, nil).Maybe()
```
Stefan-Ethernal, line 86: It seems like `.Maybe()` is not required?

```suggestion
		polygonrollupmanager.PolygonRollupManagerRollupDataReturn{}, nil)
```

bridgesync/downloader.go
```diff
@@ -190,8 +208,18 @@ func buildClaimEventHandlerPreEtrog(contract *polygonzkevmbridge.Polygonzkevmbri
 			Amount:             claimEvent.Amount,
 		}
 
+		// Extract TxnSender from the root call
```
Stefan-Ethernal: This comment is redundant, because we are not setting `TxnSender` for claims
```suggestion
```

bridgesync/downloader.go
```diff
@@ -157,8 +165,18 @@ func buildClaimEventHandler(contract *polygonzkevmbridgev2.Polygonzkevmbridgev2,
...
```
Stefan-Ethernal: Redundant comment
```suggestion
```

bridgesync/downloader.go
```diff
@@ -110,7 +110,14 @@ func buildBridgeEventHandler(
...
+		rootCall, err := extractRootCall(client, bridgeAddr, l.TxHash)
+		if err != nil {
+			return fmt.Errorf("failed to extract bridge event root call (tx hash: %s): %w", l.TxHash, err)
+		}
+
+		// Extract the specific call to the bridge contract
+		foundCall, err := findCall(*rootCall, bridgeAddr, nil, logger)
 		if err != nil {
 			return fmt.Errorf("failed to extract bridge event calldata (tx hash: %s): %w", l.TxHash, err)
 		}
```
Stefan-Ethernal: IMO it would be better if we can have the function that is able to return a found call via single function call, because this way we are enforcing all the handlers to retrieve the call by invoking 2 functions (even when the `rootCall` is not necessary, because we are not populating the `tx_sender`). WDYT?

bridgesync/backfill_tx_sender_test.go
```diff
@@ -0,0 +1,954 @@
...
+		mockClient := mocks.NewEthClienter(t)
+		logger := log.WithFields("module", "test")
+		bridgeAddr := common.HexToAddress("0x1234")
+
+		backfiller, err := NewBackfillTxnSender(invalidPath, mockClient, bridgeAddr, logger)
+		require.NoError(t, err) // sql.Open doesn't validate paths
+		require.NotNil(t, backfiller)
+
+		// Try to use the database to trigger the error
+		ctx := context.Background()
```
Stefan-Ethernal, line 137: Nitpick
```suggestion
		ctx := t.Context()
```

bridgesync/backfill_tx_sender_test.go
```diff
@@ -0,0 +1,954 @@
...
+	err := migrations.RunMigrations(dbPath)
+	require.NoError(t, err)
+
+	// Create database connection
+	database, err := db.NewSQLiteDB(dbPath)
+	require.NoError(t, err)
+	defer database.Close()
+
+	// Create test data with empty txn_sender
+	ctx := context.Background()
```
Stefan-Ethernal, line 35:
```suggestion
	ctx := t.Context()
```

bridgesync/backfill_tx_sender_test.go
```diff
@@ -0,0 +1,954 @@
...
+	// Create backfill instance
+	logger := log.WithFields("module", "test")
+	backfiller, err := NewBackfillTxnSender(dbPath, mockClient, common.HexToAddress("0x1234"), logger)
+	require.NoError(t, err)
+	defer backfiller.Close()
+
+	// Test getting records needing backfill count
+	bridgeCount, err := backfiller.getRecordsNeedingBackfillCount(ctx, "bridge")
+	require.NoError(t, err)
+	assert.Equal(t, 1, bridgeCount)
```
Stefan-Ethernal: Maybe substitute all the asserts with requires, WDYT?

```suggestion
	require.Equal(t, 1, bridgeCount)
```

bridgesync/backfill_tx_sender.go
```diff
@@ -0,0 +1,277 @@
...
+func (b *BackfillTxnSender) bulkUpdateTxnSender(
+	ctx context.Context,
+	tableName string,
+	updates []RecordUpdate,
+) error {
+	if len(updates) == 0 {
+		return nil
+	}
+
+	// Build a CASE WHEN statement for bulk update
```
Stefan-Ethernal: WDYT about this, it uses better known syntax IMO, and the end result should be the same (updating in batches), but I don't have strong opinion

```go
tx, err := db.BeginTx(ctx, nil)
if err != nil {
    return err
}

stmt, err := tx.PrepareContext(ctx, `
    UPDATE bridge
    SET txn_sender = ?
    WHERE block_num = ? AND block_pos = ?;
`)
if err != nil {
    return err
}
defer stmt.Close()

for _, update := range updates {
    _, err := stmt.ExecContext(ctx, update.TxnSender, update.BlockNum, update.BlockPos)
    if err != nil {
        _ = tx.Rollback()
        return err
    }
}

return tx.Commit()
```

bridgesync/migrations/migrations_test.go
```diff
@@ -174,6 +174,7 @@ func TestMigration0002(t *testing.T) {
 		TxHash             string   `meddler:"tx_hash"`
 		FromAddress        string   `meddler:"from_address"`
 		Calldata           []byte   `meddler:"calldata"`
+		TxnSender          string   `meddler:"txn_sender"`
```
rachit77, line 177: add require statement to check `txn_sender` and for that we need to insert a txn sender into db for test as well

aggsender/validator/validate_db_test.go
temaniarpit27: needed otherwise tests are failing

bridgesync/downloader.go
```diff
@@ -377,46 +402,68 @@ func findCall(rootCall call, targetAddr common.Address, callback func(call) (boo
...
+		return nil, nil, err
 	}
 
-	_, err = findCall(*callFrame, bridge,
+	// Return root call only if needed
+	if needTxnSender {
+		return foundCall, rootCall, nil
+	}
+	return foundCall, nil, nil
+}
```
Stefan-Ethernal, line 434: IMO I'd return a `rootCall` anyway. 

Also, as a minor further improvement in terms of readability we could consider creating a struct containing `RootCall` and `FoundCall` fields, as that way it would be clearer from the caller perspective which one is the `rootCall` and what is the `foundCall` one. WDYT?

```suggestion
func extractCallData(
	client aggkittypes.RPCClienter,
	bridgeAddr common.Address,
	txHash common.Hash,
	logger *logger.Logger,
) (*call, *call, error) {
	// Extract root call first
	rootCall, err := extractRootCall(client, bridgeAddr, txHash)
	if err != nil {
		return nil, nil, err
	}

	// Find the specific call to the bridge contract
	foundCall, err := findCall(*rootCall, bridgeAddr, nil, logger)
	if err != nil {
		return nil, nil, err
	}

	return foundCall, rootCall, nil
}
```

cmd/run.go
```diff
@@ -770,3 +804,53 @@ func createRollupDataQuerier(ctx context.Context,
...
+	backfiller, err := bridgesync.NewBackfillTxnSender(
+		cfg.DBPath,
+		client,
+		cfg.BridgeAddr,
+		log.WithFields("module", "tx-sender-backfill"),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create backfill instance: %w", err)
+	}
+	defer backfiller.Close()
```
Stefan-Ethernal, line 838: We might consider to call the `Close` as soon as all the records are backfilled. WDYT?

bridgesync/backfill_tx_sender.go
```diff
@@ -0,0 +1,272 @@
...
+		bridgeAddr: bridgeAddr,
+	}, nil
+}
+
+// BackfillAll processes bridge table to backfill txn_sender field
+func (b *BackfillTxnSender) BackfillAll(ctx context.Context) error {
+	b.log.Info("Starting txn_sender backfilling process")
+
+	// Process bridge table
+	if err := b.backfillTable(ctx, "bridge"); err != nil {
```
Stefan-Ethernal:
```suggestion
	if err := b.backfillTable(ctx, bridgeTableName); err != nil {
```

bridgesync/backfill_tx_sender.go
```diff
@@ -0,0 +1,272 @@
...
+		return fmt.Errorf("failed to backfill bridge table: %w", err)
```
Stefan-Ethernal:
```suggestion
		return fmt.Errorf("failed to backfill %s table: %w", bridgeTableName, err)
```

bridgesync/backfill_tx_sender.go
```diff
@@ -0,0 +1,272 @@
...
+	"github.com/ethereum/go-ethereum/common"
+)
+
+const (
+	// Batch size for processing records
+	batchSize = 100
+)
+
+// BackfillTxnSender handles the backfilling of txn_sender field for bridge records
+type BackfillTxnSender struct {
```
Stefan-Ethernal, line 20: Can we at some point consider implementing this as a generic component?

The draft idea that popped out is
The `BackfillManager` (to call it this way for example), would facilitate:
- gathering records that are being updated (prepared based on heuristic provided, by the `BackfillDataCollector` interface or whatever we agree upon naming). There could be multiple `BackfillDataColector`s and in this particular case, everything related to population of the `txn_sender` would be encapsulated in the `TxSenderDataCollector` (it should have the knowledge on how to gather the data and prepare update statemenets)
- run the statements, prepared by the `BackfillDataCollector` instances

WDYT?

bridgesync/backfill_tx_sender.go
```diff
@@ -0,0 +1,272 @@
...
+		return nil
+	}
+
+	tx, err := b.db.BeginTx(ctx, nil)
+	if err != nil {
+		return fmt.Errorf("failed to bulk update txn_sender: %w", err)
+	}
+	defer func() {
+		if err != nil {
+			_ = tx.Rollback()
```
Stefan-Ethernal: Can we at least log the `err` and the rollback error?

Use [this](https://github.com/agglayer/aggkit/blob/8b7c1b1cd10f041b1675e101aca1b3daab74a841/l2gersync/processor.go#L95) as an example. WDYT?

bridgesync/backfill_tx_sender.go
temaniarpit27: linter was showing error in this case, so did this way

bridgesync/backfill_tx_sender.go
temaniarpit27: linter was showing error in this case, so did this way

bridgesync/backfill_tx_sender.go
temaniarpit27: yes, i already had this in mind but decide not to overcomplicate this for now.
we can definitely make changes according to what comes next.
https://github.com/agglayer/aggkit/issues/1088

rachit77 (approved):
LGTM

bridgesync/backfill_tx_sender.go
Stefan-Ethernal: No issues on my end though :thinking: 

```bash
stefan@Stefan:~/go/src/Polygon/aggkit$ make lint
export "GOROOT=$(go env GOROOT)" && $(go env GOPATH)/bin/golangci-lint run --timeout 5m
0 issues.

stefan@Stefan:~/go/src/Polygon/aggkit$ golangci-lint --version
golangci-lint has version 2.4.0 built with go1.25.0 from (unknown, modified: ?, mod sum: "h1:qz6O6vr7kVzXJqyvHjHSz5fA3D+PM8v96QU5gxZCNWM=") on (unknown)
```

cmd/run.go
temaniarpit27: can you explain more? we are defering which should take care of all the scenarios
i prefer defer than explicity closing after a certain operation

Stefan-Ethernal: I was thinking, this would hold the db handle until the aggkit process is shut down, although if I understand correctly, we can release it as soon as `BacfkillAll` function is done (in that case we could add a `defer` statement [there](https://github.com/agglayer/aggkit/blob/8b7c1b1cd10f041b1675e101aca1b3daab74a841/bridgesync/backfill_tx_sender.go#L48-L58)). WDYT?

Stefan-Ethernal (approved):

PR #1077, opened 2025-10-02
fix: hardcode finalized block for bridge l1 syncer

### LINKED ISSUE #1078: handle empty block finality in evm downloader
In case the block finality is set to `Empty`, which is case for the `BridgeL1Sync` component, due to config `BridgeL1Sync.BlockFinality` being deprecated, we need to hardcode the block finality to the `FinalizedBlock`, instead of silently setting it to `Empty`.

Stefan-Ethernal:
## 🔄 Changes Summary
- In case the block finality is set to `Empty`, which is case for the `BridgeL1Sync` component, due to config `BridgeL1Sync.BlockFinality` being deprecated, we are hardcoding the block finality to the `FinalizedBlock`.
- Also in case, `Empty` block finality is received to the `EVMDownloader`, we are returning an error.
- Remove the `BlockFinality` from `BridgeSync`, as it was not being used anywhere, except the unit tests.
- Rename `BlockNumberFinality.GreaterThan` to `BlockNumberFinality.LessFinalThan`, and make it slightly more understandable.
- Remove redundant check when declaring block as finalized in evm downloader. It is enough checking whether the given block is less or equal to the last finalized block (whatever we have configured as the `finalized` commitment).

## ⚠️ Breaking Changes
N/A

## 📋 Config Updates
N/A

## ✅ Testing
- 🤖 **Automatic**: `Test_ResolveBlockFinality` unit test
- 🖱️ **Manual**:
1. Spin up the aggkit with bridge service
2. Note that there is not the following log entry in the aggkit log for `L1BridgeSyncer`, since it is now initialized with the `finalized` keyword for block finality:
```go
"finalized block type FinalizedBlock is greater than block finality , setting finalized block type to "
```

## 🐞 Issues
- Closes #1078

## 🔗 Related PRs

## 📝 Notes


joanestebanr (commented):
What is the case that `BlockFinality` is `Empty`?

Stefan-Ethernal:
> What is the case that `BlockFinality` is `Empty`?

In the `BridgeL1Sync` config, the `BlockFinality` is omitted (it was supposed to be `FinalizedBlock`, but it isn't the case, since it is omitted and an empty string implicitly):

```toml
[BridgeL1Sync]
DBPath = "{{PathRWData}}/bridgel1sync.sqlite"
InitialBlockNum = 0
BridgeAddr = "{{polygonBridgeAddr}}"
SyncBlockChunkSize = 100
RetryAfterErrorPeriod = "1s"
WaitForNewBlocksPeriod = "3s"
DBQueryTimeout = "{{defaultDBQueryTimeout}}"
```

We could provide [here](https://github.com/agglayer/aggkit/blob/a74692996df2a4b8ebb46766d80fb9f4b58e6746/bridgesync/bridgesync.go#L84) the `aggkittypes.FinalizedBlock` for sure, but with the fix I made we are sure that we are not going to set the `Empty` block finality.

sync/evmdownloader_test.go
```diff
@@ -32,6 +32,62 @@ const (
...
+			name: "finalized block type greater due to offset",
+			finality: aggkittypes.BlockNumberFinality{
+				Block:  aggkittypes.Safe,
+				Offset: 1,
+			},
+			finalizedBlockType: aggkittypes.BlockNumberFinality{
+				Block:  aggkittypes.Safe,
+				Offset: 5,
+			},
+			expectedFinality: aggkittypes.BlockNumberFinality{
```
goran-ethernal: If we are choosing the greater finality, in this case, shouldn't the finality with bigger offset be chosen?

Stefan-Ethernal: We are not choosing the greater finality, but the stricter one (in case `finalizedBlockType` is greater than finalized, we are "lowering the `finalizedBlockType` to match the `finality` parameter). 

Now, since the offset can be both positive and negative, the `safe/1` is stricter than `safe/5` in this case (`safe/1` goes only 1 block in future, relative to the safe block, whereas the `safe/5` resolves to safe block+5). Agree?

goran-ethernal (approved):

temaniarpit27 (approved):

rachit77 (approved):
LGTM

sync/evmdownloader.go
```diff
@@ -202,7 +218,7 @@ func (d *EVMDownloader) Download(ctx context.Context, fromBlock uint64, download
 func (d *EVMDownloader) reportBlocks(downloadedCh chan EVMBlock, blocks EVMBlocks, lastFinalizedBlock uint64) {
 	for _, block := range blocks {
 		d.log.Debugf("sending block %d to the driver (with events)", block.Num)
-		block.IsFinalizedBlock = d.finalizedBlockType.IsFinalized() && block.Num <= lastFinalizedBlock
+		block.IsFinalizedBlock = block.Num <= lastFinalizedBlock
```
joanestebanr, line 204: why have you removed the condition `d.finalizedBlockType.IsFinalized()`? do you think that the source of true is the configured block as 'finalized' so you believe that, isn't?

sync/evmdownloader.go
```diff
@@ -100,6 +93,29 @@ func NewEVMDownloader(
...
+// block finality (ordering: Finalized ≤ Safe ≤ Latest ≤ Pending), then the block finality
+// is downgraded to match the target finality, and a warning is logged.
+//
+// Otherwise, the finalizedBlockType is returned as-is.
+//
+// Example:
+//   - finality=Finalized, finalizedBlockType=Safe  → returns Finalized (with warning)
+//   - finality=Safe,       finalizedBlockType=Safe → returns Safe
+//   - finality=Safe,       finalizedBlockType=Latest → returns Safe (with warning)
+func resolveBlockFinality(logger *log.Logger,
```
joanestebanr: I don't see the usage of this function: 
- The case that I understand that try to fix is: 
- `finality=Safe,       finalizedBlockType=Latest`
And I don't see any reason to change `finalizedBlockType`... `Safe` is going to be always final (because `Latest>Safe`). 
Maybe you can just show a warning in case
`finality<finalizedBlockType`. WDYT?

joanestebanr (approved):

sync/evmdownloader.go
Stefan-Ethernal: Exactly — the configured commitment level (e.g. `finalized`, `safe`, etc.) is already taken into account when resolving `lastFinalizedBlock`. Therefore, this condition is redundant. In fact, it could even cause incorrect behavior — for example, if safe is set as the `finalizedBlockType`, such blocks would be wrongly excluded from being treated as finalized.

sync/evmdownloader.go
Stefan-Ethernal: This logic was already defined [here](https://github.com/agglayer/aggkit/blob/81b76ce04db6c033a014d450a346407dce987868/sync/evmdownloader.go#L74-L80), where the `finalizedBlockType` is overridden to be **stricter** in case the `finality` is configured to be more strict than the `finalizedBlockType` (not the other way around):

```go
if finalizedBlockType.GreaterThan(&finality) {
	fbtEthermanType = finality
	logger.Warnf("finalized block type %s is greater than block finality %s, setting finalized block type to %s",
		finalizedBlockType.String(), finality.String(), fbtEthermanType.String())
}
```

In short, we make `finalizedBlockType` stricter when `finality` is stricter -
e.g. `finality=Safe`, `finalizedBlockType=Latest` → `finalizedBlockType=Safe`.

PR #1101, opened 2025-10-10
feat: limit number of network id in bridge and claim api

### LINKED ISSUE #1103: limit number of network id in bridge and claim api
Limit number of network ids that can be passed with bridge and claim api to reduce DoS attack surface

rachit77:
## 🔄 Changes Summary
- This PR adds a limit on number of network id that can be provided with bridge and claim api to filter destination/source network id in api query

## 🐞 Issues
- Closes #1103


bridgeservice/bridge.go
```diff
@@ -376,7 +376,7 @@ func (b *BridgeService) GetBridgesHandler(c *gin.Context) {
 // @Param network_id query uint32 true "Origin network ID"
 // @Param page_number query uint32 false "Page number (default 1)"
 // @Param page_size query uint32 false "Page size (default 100)"
-// @Param network_ids query []uint32 false "Filter by one or more destination network IDs"
+// @Param network_ids query []uint32 false "Filter by one or more source network IDs (maximum 5 allowed)"
```
temaniarpit27, line 379: destination or source?

temaniarpit27 (approved):

bridgeservice/bridge.go
rachit77: resolving this as discussed over the huddle

bridgeservice/utils.go
```diff
@@ -78,13 +80,20 @@ func parseUintQuery[T UintParam](c *gin.Context, key string, mandatory bool, def
 }
 
 // parseUint32SliceParam parses a slice of uint32 parameters from the request context
+// and enforces a maximum limit on the number of network IDs to prevent DoS attacks
 func parseUint32SliceParam(c *gin.Context, key string) ([]uint32, error) {
```
joanestebanr: The function `parseUint32SliceParam` looks like generic one, I see that is only used for  parsing `networkID` but I suggest or to pass the max number in the function or to rename it to something like `parseNetworkIDSliceParam`. WDTY?

rachit77: makes sense i will rename `parseUint32SliceParam` to `parseNetworkIDSliceParam`

rachit77: done: b5495e17eaf5028d1ec594e1f857659cc3b0298e

Stefan-Ethernal (approved):

PR #1105, opened 2025-10-10
feat: unify all e2e to latest commitid in e2e main branch

joanestebanr:
## 🔄 Changes Summary
- Fix all aggkit e2e tests to work with the changes in `aggkit_bridge_service.bash`

### DIFF
--- a/.github/test_e2e_cdk_erigon_args_base.json
+++ b/.github/test_e2e_cdk_erigon_args_base.json
@@ -13,6 +13,7 @@
     "binary_name": "aggkit",
     "aggkit_components": "aggsender",
     "aggkit_image": "aggkit:local",
+    "agglayer_image": "ghcr.io/agglayer/agglayer:0.4.0-rc.18",
     "agglayer_contracts_image": "jhkimqd/agglayer-contracts:v0.0.0-rc.0.fix.1.create.genesis-fork.12",
     "agglayer_prover_primary_prover": "mock-prover",
     "zkevm_use_real_verifier": false,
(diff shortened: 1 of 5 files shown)


goran-ethernal (approved):

.github/test_e2e_cdk_erigon_args_base.json
```diff
@@ -13,6 +13,7 @@
     "binary_name": "aggkit",
     "aggkit_components": "aggsender",
     "aggkit_image": "aggkit:local",
+    "agglayer_image": "ghcr.io/agglayer/agglayer:0.4.0-rc.18",
```
temaniarpit27, line 16: what image is kurtosis using?

temaniarpit27 (approved):

.github/test_e2e_cdk_erigon_args_base.json
joanestebanr: -rc.12

test/run-local-e2e.sh
Stefan-Ethernal, line 1: Btw would it make sense to add fail fast for the other test suites as well (e.g. `single-l2-network-op-pessimistic` etc):

```bash
bats <test_file> || exit 1
```

Stefan-Ethernal (approved):

vcastellm (approved):

test/run-local-e2e.sh
joanestebanr: Totally agree! I think that  we must create another PR to work on it. Also to add fail-fast in the same file (right now if one of the tests fails it keep running the rest)

PR #1109, opened 2025-10-11
feat: Add latest block finality support to `bridgeL1Sync`

rachit77:
## 🔄 Changes Summary
- This PR enables bridgeL1Sync to run with latest block finality and reintroduces the L1 reorg detector as it was required by bridge hub

## 📋 Config Updates
- 🧾 **Diff/Config snippet**: 

1. Added `[ReorgDetectorL1]` config
```
[ReorgDetectorL1]
DBPath = "{{PathRWData}}/reorgdetectorl1.sqlite"
FinalizedBlock = "FinalizedBlock" 
```

2. Added `BlockFinality = "FinalizedBlock"` in `[BridgeL1Sync]`

## ✅ Testing
- 🖱️ **Manual**: [Optional: Steps to verify]

## 🐞 Issues
- Closes #1107

PR #1120, opened 2025-10-14
feat: Add tests for L1 reorg detector in bridge syncer

### LINKED ISSUE #1113: Review and fix reorg detector
we are introducing reorg detector in bridgesyncl1 and want to make sure reorg detector is working as expected

rachit77:
## 🔄 Changes Summary
- This PR introduces tests to validate the L1 reorg detector used by the bridge syncer

## 📋 Config Updates
- changed the default `BlockFinality` of `[BridgeL1Sync]` from `FinalizedBlock` to `LatestBlock`

## ✅ Testing
- 🖱️ **Manual**: Added test cases [bridgesync/e2e_test.go](https://github.com/agglayer/aggkit/blob/c6b5011fafeb1151c30e77dca11c2811e3c30b09/bridgesync/e2e_test.go)

## 🐞 Issues
- Closes #1113 

## 🔗 Related PRs
- https://github.com/agglayer/aggkit/pull/1109


db/sqlite.go
```diff
@@ -21,10 +21,19 @@ var (
...
 func NewSQLiteDB(dbPath string) (*sql.DB, error) {
-	return sql.Open("sqlite3", fmt.Sprintf(
+	db, err := sql.Open("sqlite3", fmt.Sprintf(
 		"file:%s?_txlock=immediate&_foreign_keys=on&_journal_mode=WAL&_busy_timeout=30000",
 		dbPath,
 	))
+	if err != nil {
+		return nil, err
+	}
+	// TODO - check if this is needed
```
rachit77: ideally this is not required

bridgesync/processor.go
```diff
@@ -827,7 +827,8 @@ func (p *processor) ProcessBlock(ctx context.Context, block sync.Block) error {
 		}
 	}()
 
-	if _, err := tx.Exec(`INSERT INTO block (num, hash) VALUES ($1, $2)`, block.Num, block.Hash.String()); err != nil {
+	// TODO - need to make sure this change is correct and does not affect the existing code
```
rachit77: ideally this shouldn't we required as once reorg is detected than blocks from `block table` is removed and hence there should not be any conflict on primary key

reorgdetector/reorgdetector.go
```diff
@@ -40,6 +40,7 @@ type ReorgDetector struct {
 
 	subscriptionsLock sync.RWMutex
 	subscriptions     map[string]*Subscription
+	headersCache      map[uint64]*types.Header
```
joanestebanr, line 43: I don't see what is the usage of this map

sync/evmdownloader.go
```diff
@@ -311,6 +327,32 @@ func (d *EVMDownloaderImplementation) WaitForNewBlocks(
 			if blockNumber > latestSyncedBlock {
 				return blockNumber
 			}
+			// If blockNumber <= latestSyncedBlock, a reorg may have occurred
+			// Get the block header to verify the hash and notify the reorg detector
+			if blockNumber <= latestSyncedBlock {
+				trackedBlock, err := d.reorgDetector.GetTrackedBlockByBlockNumber(d.reorgDetectorID, blockNumber)
```
joanestebanr, line 336: Maybe this code would make more sene to put it `reorgDetector.AddBlockToTrack`... WDYT?

temaniarpit27: AddBlockToTrack is majorly for tracking purpose
the goal here is identifying if there is a reorg

reorgdetector/reorgdetector.go
temaniarpit27: this is used to keep track of headers of chain
once a chain reorgs, it is used to check the header mismatch

temaniarpit27 (approved):

sync/evmdownloader.go
goran-ethernal: I think having reorg detection out of reorg detector is not a clean approach. I do know it is not easy to do this in the reorg detector itself, because these are the blocks that only the downloader knows, but maybe it can be done differently, without the need to pass reorg detector around.

Why can't be just do what we did before and just report these blocks in the download function to reorg detector, using the the provided channel when the `WaitForNewBlocks` returns a block.

types/block_finality.go
```diff
@@ -119,14 +120,17 @@ func (b BlockNumberFinality) IsLatest() bool {
 	return b.Block == Latest && b.Offset >= 0
 }
 
-// BlockNumber gets the safe block number from RPC
-func (b *BlockNumberFinality) BlockNumber(ctx context.Context, requester ethereum.ChainReader) (uint64, error) {
+// BlockNumber gets the safe block number and block header from RPC
+func (b *BlockNumberFinality) BlockNumber(
```
goran-ethernal, line 124: I would create a new function called `Block` or `Header`, and call it in places where it needs the block itself and not just number.

sync/evmdownloader.go
temaniarpit27: so basically there is no channel currently for reporting blocks to reorg detector
we need all blocks to be reported to reorg detector this is the only place where we can get them
also, as @joanestebanr mentioned, we dont have all the blocks being handled, only the ones which have certain events in it and that is another flow. so in the original flow, only blocks which have events were being tracked but we wanted to track all of them
i know this is not clean but making changes in other flows was more riskier than changing here and i didnt want to change those.

config/default.go
```diff
@@ -153,7 +153,7 @@ MaxRequestsPerIPAndSecond = 10
 
 [BridgeL1Sync]
 DBPath = "{{PathRWData}}/bridgel1sync.sqlite"
-BlockFinality = "FinalizedBlock"
+BlockFinality = "LatestBlock"
```
rachit77, line 154: maybe we should make `FinalizedBlock` by default
wdyt @temaniarpit27 ?

temaniarpit27: no otherwise we need to tell people to use `LatestBlock` for bridge service and would be hassle
and this DB is only used for bridge service so latest makes more sense

sync/evmdownloader.go
goran-ethernal: I completely understand. I do not like the current reorg detector implementation neither, it should sync all blocks to properly handle reorgs, and not just the ones that downloader sends. Hopefully, @joanestebanr will fix this with his proposal 😄 .

bridgesync/processor.go
```diff
@@ -784,11 +784,27 @@ func (p *processor) Reorg(ctx context.Context, firstReorgedBlock uint64) error {
...
 		p.log.Errorf("failed to commit reorg transaction: %v", err)
 		return err
 	}
 
+	newTx, err := db.NewTx(dbCtx, p.db)
+	if err != nil {
+		p.log.Errorf("failed to start transaction for reorg: %v", err)
+		return err
+	}
+	err = p.exitTree.InitCache(newTx)
```
goran-ethernal: Curious, why do you need to initialize tree cache, and couldn't you just use the `db` transaction that was already open?

reorgdetector/reorgdetector.go
```diff
@@ -114,11 +117,13 @@ func (rd *ReorgDetector) GetFinalizedBlockType() aggkittypes.BlockNumberFinality
...
 }
 
+// GetDB returns the database connection for testing purposes
+func (rd *ReorgDetector) GetDB() *sql.DB {
+	return rd.db
+}
+
 // AddBlockToTrack adds a block to the tracked list for a subscriber
 func (rd *ReorgDetector) AddBlockToTrack(ctx context.Context, id string, num uint64, hash common.Hash) error {
-	if rd.IsDisabled() {
```
goran-ethernal, line 119: This function is no longer used.

types/block_finality.go
```diff
@@ -129,6 +133,26 @@ func (b *BlockNumberFinality) BlockNumber(ctx context.Context, requester ethereu
 	return b.Block.ApplyOffset(blockHeader.Number.Uint64(), b.Offset), nil
 }
 
+// Header gets the block header with offset from RPC
+func (b *BlockNumberFinality) BlockHeaderWithOffset(
```
goran-ethernal: I would just name it `BlockHeader`, to be consistent with the `BlockNumber` function. Offset is something internal to block finality. Also, update the function comment.

sync/evmdriver.go
```diff
@@ -199,10 +200,6 @@ func (d *EVMDriver) handleNewBlock(ctx context.Context, b EVMBlock) error {
 }
 
 func (d *EVMDriver) trackNonFinalizedBlock(ctx context.Context, b EVMBlock) error {
-	if b.IsFinalizedBlock {
```
goran-ethernal, line 202: Why remove this? This was added so that we do not send already finalized blocks to reorg detector. I would even rename this function to `trackBlock`.

sync/evmdownloader.go
joanestebanr: I don't see the point of this code: 
- When you add the block to reorg using `AddBlockToTrack`  you can check if the hash is the same as the tracked one
- In the code you check the `hash` but just emit a warning, no action is taken

sync/evmdriver.go
```diff
@@ -180,7 +181,7 @@ reset:
 		<-blockProcessingDone
 		return
 	case firstReorgedBlock := <-d.reorgSub.ReorgedBlock:
-		d.log.Warnf("Reorg detected from block %d, stopping block processing...", firstReorgedBlock)
```
joanestebanr, line 183: why do you remove the `stopping...`, is stopping it, isn't?

types/block_finality.go
joanestebanr: I think that it's redundant `WithOffset`, maybe you can rename to `BlockHeader` that are more aligned to `BlockNumber`, WDYT?

temaniarpit27: +1

sync/evmdriver.go
temaniarpit27: +1

types/block_finality.go
rachit77: done 935f3be19c74eead7eeedf9448ccaa0d6711a178

rachit77: fixed

reorgdetector/reorgdetector.go
rachit77: fixed bb807758382b183fe07a6284006c9e4f0990de43

sync/evmdriver.go
rachit77: fixed: ff2a209b32165d5c53898e05d88a3158f44ffa0f

Makefile
```diff
@@ -94,7 +94,7 @@ build-docker-nc: ## Builds a docker image with the aggkit binary - but without b
 
 .PHONY: test-unit
 test-unit: ## Runs the unit tests
-	trap '$(STOP)' EXIT; MallocNanoZone=0 go test -count=1 -short -race -p 1 -covermode=atomic -coverprofile=coverage.out -timeout 15m ./...
+	trap '$(STOP)' EXIT; MallocNanoZone=0 go test -count=1 -short -race -p 1 -covermode=atomic -coverprofile=coverage.out -timeout 30m ./...
```
Stefan-Ethernal: Which tests are now time consuming, and do we really need to have the timeout set to 30 mins?

types/block_finality_test.go
```diff
@@ -282,3 +287,51 @@ func TestBlockNumberFinalityJSONSchema(t *testing.T) {
...
+		require.NoError(t, err)
+		require.Equal(t, offsetHeader, result)
+	})
+
+	t.Run("Error on first call", func(t *testing.T) {
+		mockClient := mocks.NewBaseEthereumClienter(t)
+		blockFinality := BlockNumberFinality{Block: Latest, Offset: 0}
+
+		testErr := fmt.Errorf("first call error")
+		mockClient.On("HeaderByNumber", ctx, (*big.Int)(nil)).Return(nil, testErr).Once()
```
Stefan-Ethernal: Use the expecter, as it is compile time safe. The same comment applies for other places as well.

```suggestion
		mockClient.EXPECT().HeaderByNumber(ctx, (*big.Int)(nil)).Return(nil, testErr).Once()
```

types/block_finality.go
```diff
@@ -129,6 +133,26 @@ func (b *BlockNumberFinality) BlockNumber(ctx context.Context, requester ethereu
 	return b.Block.ApplyOffset(blockHeader.Number.Uint64(), b.Offset), nil
 }
 
+// Header gets the block header from RPC
```
Stefan-Ethernal: Update the comment
```suggestion
// BlockHeader gets the block header from RPC
```

types/block_finality.go
```diff
@@ -119,8 +120,11 @@ func (b BlockNumberFinality) IsLatest() bool {
 	return b.Block == Latest && b.Offset >= 0
 }
 
-// BlockNumber gets the safe block number from RPC
-func (b *BlockNumberFinality) BlockNumber(ctx context.Context, requester ethereum.ChainReader) (uint64, error) {
+// BlockNumber gets the safe block number RPC
```
Stefan-Ethernal:
```suggestion
// BlockNumber gets the block number from RPC with offset taken into account
```

types/block_finality.go
```diff
@@ -119,8 +120,11 @@ func (b BlockNumberFinality) IsLatest() bool {
...
+func (b *BlockNumberFinality) BlockNumber(
+	ctx context.Context,
+	requester ethereum.ChainReader,
+) (uint64, error) {
 	blockHeader, err := requester.HeaderByNumber(ctx, b.Block.toBigInt())
```
Stefan-Ethernal, line 128: Just thinking out loud, as it seems like the logic of this function is not entirely correct. 

To me it makes sense to do it this way:
1. apply offset to given block number
2. try to retrieve the header by number:
   a. if not found, it means that it is an error for sure, as the given block is not yet created (offset might be set to some future block for instance),
   b. if found, we return the block number with the offset applied

Does it make sense?

cc @joanestebanr

types/block_finality.go
```diff
@@ -129,6 +133,26 @@ func (b *BlockNumberFinality) BlockNumber(ctx context.Context, requester ethereu
...
+		return nil, err
+	}
+
+	blockNumberOffset := b.Block.ApplyOffset(blockHeader.Number.Uint64(), b.Offset)
+	blockHeaderOffset, err := requester.HeaderByNumber(ctx, new(big.Int).SetUint64(blockNumberOffset))
+	if err != nil {
+		log.Errorf("BlockNumberFinality.Header: Error getting block %s. Err: %s", b.String(), err.Error())
+		return nil, err
+	}
+	return blockHeaderOffset, nil
```
Stefan-Ethernal: IMO we can simplify this function, by first invoking the `BlockNumber` (to resolve the actual block number) and then just returning `requester.HeaderByNumber`. WDYT?

```suggestion
	blockNum, err := b.BlockNumber(ctx, requester)
	if err != nil {
		return nil, fmt.Errorf("failed to resolve offseted block by number (block number: %s): %w", b, err)
	}

	return requester.HeaderByNumber(ctx, new(big.Int).SetUint64(blockNum))
```

sync/evmdriver.go
rachit77: fixed: 3db5ae4df6941a8b58bd762739644c5dd23e1646

we removed this for local testing

bridgesync/processor.go
Stefan-Ethernal: What happens if the transaction gets rollbacked? Wouldn't it mean that the in memory representation of `exitTree` will be out of date in that case?

bridgesync/e2e_test.go
```diff
@@ -118,3 +122,516 @@ func getFinalizedBlockNumber(t *testing.T, ctx context.Context, client simulated
 	require.NoError(t, err)
 	return lastBlockHeader.Number.Uint64()
 }
+
+// TestBridgeL1SyncerWithReorgDetector tests the bridge L1 syncer with reorg detector
+func TestBridgeL1SyncerWithReorgDetector(t *testing.T) {
+	ctx := context.Background()
+	dbPathSyncer := path.Join(t.TempDir(), "bridgesyncTestWithReorgs_sync.sqlite")
+	dbPathReorg := path.Join(t.TempDir(), "bridgesyncTestWithReorgs_reorg.sqlite")
+	blocktime := time.Second * 6
```
goran-ethernal: This looks too much 🤔. Do we need block time this long?

bridgesync/e2e_test.go
```diff
@@ -118,3 +122,516 @@ func getFinalizedBlockNumber(t *testing.T, ctx context.Context, client simulated
...
+
+	t.Log("✅ Same hash reorg edge case test completed successfully")
+}
+
+// TestBridgeL1SyncerWithMultipleReorgs tests the bridge L1 syncer with multiple reorgs
+func TestBridgeL1SyncerWithMultipleReorgs(t *testing.T) {
+	ctx := context.Background()
+	dbPathSyncer := path.Join(t.TempDir(), "bridgesyncTestWithReorgs_sync.sqlite")
+	dbPathReorg := path.Join(t.TempDir(), "bridgesyncTestWithReorgs_reorg.sqlite")
+	blocktime := time.Second * 6
```
goran-ethernal: Also, do we need block time this long?

bridgesync/e2e_test.go
Stefan-Ethernal: I feel like the 1 second should be long enough, but I'd be happier with the fraction of the second if possible...

sync/evmdownloader.go
```diff
@@ -283,7 +297,8 @@ func (d *EVMDownloaderImplementation) GetLastFinalizedBlock(ctx context.Context)
 	if blockFinality == nil {
 		blockFinality = &d.blockFinality
 	}
-	return blockFinality.BlockNumber(ctx, d.ethClient)
+	blockNumber, err := blockFinality.BlockNumber(ctx, d.ethClient)
+	return blockNumber, err
```
Stefan-Ethernal:
```suggestion
	return blockFinality.BlockNumber(ctx, d.ethClient)
```

sync/evmdownloader.go
```diff
@@ -54,6 +54,8 @@ type EVMDownloader struct {
 	finalizedBlockType         *aggkittypes.BlockNumberFinality
 	stopDownloaderOnIterationN int
 	addressesToQuery           []common.Address
+	reorgDetector              ReorgDetector
+	reorgDetectorID            string
```
Stefan-Ethernal: We don't need this field...

```suggestion
```

sync/evmdownloader.go
```diff
@@ -54,6 +54,8 @@ type EVMDownloader struct {
 	finalizedBlockType         *aggkittypes.BlockNumberFinality
 	stopDownloaderOnIterationN int
 	addressesToQuery           []common.Address
+	reorgDetector              ReorgDetector
```
Stefan-Ethernal: Unused

```suggestion
```

bridgesync/e2e_test.go
```diff
@@ -118,3 +122,516 @@ func getFinalizedBlockNumber(t *testing.T, ctx context.Context, client simulated
...
+
+	// Wait for syncer to process
+	helpers.WaitForSyncerToCatchUp(ctx, t, syncer, client)
+
+	// Step 4: Record the block hash to fork from later (fork from the current block to ensure reorg detection)
+	t.Log("Step 4: Recording block hash for fork point")
+	forkBlockNum, err := client.Client().BlockNumber(ctx)
+	require.NoError(t, err)
+	// Fork from the current block (which should be tracked) to ensure reorg detection
+	forkFromBlock := forkBlockNum
```
Stefan-Ethernal: Why we should have a copy of forkBlockNum, as it is not altered anywhere?

```suggestion
```

joanestebanr (approved):

types/block_finality.go
rachit77: done: 93c0c1e287c157184f4aeb3c25c2d90ddad2789f

types/block_finality.go
rachit77: done: 93c0c1e287c157184f4aeb3c25c2d90ddad2789f

bridgesync/e2e_test.go
rachit77: decreased it to 100ms
93c0c1e287c157184f4aeb3c25c2d90ddad2789f

bridgesync/e2e_test.go
rachit77: fixed: 93c0c1e287c157184f4aeb3c25c2d90ddad2789f

Makefile
rachit77: reverted: 08e52dc4bf22ac5eb134e1b641a6e60acc609cc9 
bridgesync e2e test were consuming this increased time

types/block_finality_test.go
rachit77: fixed: 8bbb797d15b7a1d63a88741bcb87348acc81496d

sync/evmdownloader.go
rachit77: done: 8bbb797d15b7a1d63a88741bcb87348acc81496d

sync/evmdownloader.go
rachit77: we are using `reorgDetectorID` at bunch of places for logging, in the trackedBlocks mapping ([here](https://github.com/agglayer/aggkit/blob/8bbb797d15b7a1d63a88741bcb87348acc81496d/reorgdetector/reorgdetector.go#L125-L126)) etc

Stefan-Ethernal: I'm just saying that we don't need it in the `EVMDownloader` per se, as we are only setting it in the constructor but not using it anywhere else. 

OTOH we have those two fields repeated in the `EVMDownloaderImplementation`, which is [used](https://github.com/agglayer/aggkit/blob/8bbb797d15b7a1d63a88741bcb87348acc81496d/sync/evmdownloader.go#L242-L243) on different places, and it should remain.

bridgesync/e2e_test.go
rachit77: done: 056f1ec0385fe1656a765bab4d2e32efcaa48b59

types/block_finality.go
rachit77: example scenario:
if someone sets the parameters such that `blockFinality` is `finalized` and the `offset` is `1000`, the function would end up returning a future block, since the `ApplyOffset` function doesn’t handle this case correctly.

sync/evmdownloader.go
rachit77: done: 5e0ce8cd8d346ae4568e6748578da825560a7cff

sync/evmdownloader.go
rachit77: done: 5e0ce8cd8d346ae4568e6748578da825560a7cff

types/block_finality.go
joanestebanr: I agree with @Stefan-Ethernal if you're not able to retreive the target block (e.g `finalized/10` finalized(80)+10 -> block(90) )  you have to return an error. Your example of  `finalized/1000`  is a case of misconfiguration

types/block_finality.go
rachit77: done: 66beb28ec8bc736d5067c31be3cacaffb8b2a48d

bridgesync/processor.go
rachit77: removed the cache initialize here
as we don't actually require it here since we already rebuilt it [here](https://github.com/agglayer/aggkit/blob/66beb28ec8bc736d5067c31be3cacaffb8b2a48d/tree/appendonlytree.go#L40-L41) in case of last index mismatch

types/block_finality.go
rachit77: fixed:  c7dee3b61136bf456b76a15d559d1275059f5a78

goran-ethernal (approved):

Stefan-Ethernal (approved):

PR #1122, opened 2025-10-14
fix: aggchain params verification

### LINKED ISSUE #1099: fix: `AggchainParams` verification on `aggsender-validator`
The `AggchainParams` field verification in `FEP` certificates on `aggsender-validator` keeps failing.

The reason is we are not calculating the `AggchainParams` with all the fields.
The fields required for building `AggchainParams` can be found [here ](https://github.com/agglayer/provers/blob/9107723fe86082db56e845bfdcc6d5f4f47f4090/crates/aggchain-proof-core/src/full_execution_proof.rs#L139)on the prover repo. Also, the way how the necessary data is gathered is [here](https://github.com/agglayer/provers/blob/9107723fe86082db56e845bfdcc6d5f4f47f4090/crates/aggchain-proof-builder/src/lib.rs#L224).

goran-ethernal:
## 🔄 Changes Summary
This PR fixes the `aggchain params` field verification for `FEP` certificates.

`AggchainParams` is calculated by `keccak256` hashing of `abi.encodePacked` data:

```rust
struct AggchainParamsValues {
     bytes32 l2_pre_root;
     bytes32 claim_root;
     uint256 claim_block_num;
     bytes32 rollup_config_hash;
     bool optimistic_mode;
     address trusted_sequencer;
     bytes32 range_vkey_commitment;
     bytes32 aggregation_vkey_hash;
}

pub fn encoded_aggchain_params(&self) -> Vec<u8> {
     AggchainParamsValues::abi_encode_packed(&AggchainParamsValues::from(self))
}

/// Compute the chain-specific commitment forwarded to the PP.
pub fn aggchain_params(&self) -> Digest {
     keccak256(self.encoded_aggchain_params().as_slice())
}
```
Since `go-ethereum` does not provide `abi.EncodePacked`, this PR uses a custom function built for doing the same thing for the `AggchainParams` struct.

Since we see that `AggchainParams` is essentially built from public inputs of `aggchain proof` + fields `optimistic_mode` and `aggregation_vkey_hash`, this PR utilized the already present public inputs calculation to calculate the `AggchainParams` itself.

## ⚠️ Breaking Changes
NA

## 📋 Config Updates
NA

## ✅ Testing

## 🐞 Issues
- Closes #1099


aggsender/validator/validate_certificate.go
```diff
@@ -52,16 +52,10 @@ func (a *CertificateValidator) ValidateCertificate(ctx context.Context, params t
 	if params.Certificate == nil {
 		return ErrNilCertificate
 	}
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/1122/commits/58727e091a0f8f9e44d0dfc003dac76669fb3a18

aggsender/types/fep_inputs.go
```diff
@@ -0,0 +1,147 @@
...
+// AggregationProofPublicValues represents the public values used in the aggregation proof.
+type AggregationProofPublicValues struct {
+	L1Head              common.Hash
+	L2PreRoot           common.Hash
+	ClaimRoot           common.Hash
+	L2BlockNumber       uint64
+	RollupConfigHash    common.Hash
+	MultiBlockVKey      common.Hash
+	TrustedSigner       common.Address
+	AggregationVKeyHash common.Hash // not used in the hash calculation
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/1122/commits/58727e091a0f8f9e44d0dfc003dac76669fb3a18

joanestebanr (approved):

aggsender/validator/validate_certificate.go
```diff
@@ -52,16 +52,10 @@ func (a *CertificateValidator) ValidateCertificate(ctx context.Context, params t
...
-		err                        error
-	)
-
-	if params.PreviousCertificate != nil {
-		previousCertificateToBlock, err = a.certQuerier.GetLastSettledCertificateToBlock(ctx, params.PreviousCertificate)
-		if err != nil {
-			return fmt.Errorf("failed to get last settled certificate block: %w", err)
-		}
+
+	previousCertificateToBlock, err := a.certQuerier.GetLastSettledCertificateToBlock(ctx, params.PreviousCertificate)
```
Stefan-Ethernal, line 56: Are you sure it is safe to assume that the `params.PreviousCertificate` will be always set?

We are setting it [here](https://github.com/agglayer/aggkit/blob/58727e091a0f8f9e44d0dfc003dac76669fb3a18/aggsender/validator/validator_service.go#L89), but what if this condition is not met, then it will remain undefined?

```go
if req.PreviousCertificateId != nil && req.PreviousCertificateId.Value != nil {
```

aggsender/query/agg_proof_fep_inputs_query.go
```diff
@@ -0,0 +1,66 @@
...
+// GetAggchainParams generates the aggchain parameters required for the verification of aggchain proof
+func (f *FEPInputsQuery) GetAggchainParams(
+	lastProvenBlock, requestedEndBlock uint64,
+	l1InfoTreeLeafHash common.Hash) (*types.AggchainParams, error) {
+	publicValues, err := f.GetPublicInputs(lastProvenBlock, requestedEndBlock, l1InfoTreeLeafHash)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get FEP public inputs: %w", err)
+	}
+
+	isOptimisticModeOn, err := f.aggchainFEPContract.OptimisticMode(nil)
```
Stefan-Ethernal: nitpick
```suggestion
	isOptimisticModeEnabled, err := f.aggchainFEPContract.OptimisticMode(nil)
```

aggsender/types/fep_inputs.go
```diff
@@ -0,0 +1,147 @@
...
+		a.TrustedSigner.String(),
+		a.MultiBlockVKey.String(),
+		a.AggregationVKeyHash.String(),
+	)
+}
+
+// Hash calculates the hash of the AggchainParams using a custom encoding,
+// the same as abi.EncodePacked, since go-ethereum does not provide a function for that.
+func (a *AggchainParams) Hash() (common.Hash, error) {
+	buf := new(bytes.Buffer)
```
Stefan-Ethernal, line 120: Probably it'd be a good idea to pre-allocate the size, since all of the fields are fixed length. WDYT?

```suggestion
	const totalSize = 213
	buf := bytes.NewBuffer(make([]byte, 0, totalSize)) // preallocate ~213 bytes
```

aggsender/validator/validate_certificate.go
goran-ethernal: No, but the `GetLastSettledCertificateToBlock` handles this. I wrote a comment in that function, that even if we do not have a previously settled certificates (meaning, this is the first certificate in the network), we need to check what is the starting block on the `AggchainFEP` rollup if this is a FEP network. All that is already done in the `GetLastSettledCertificateToBlock` function.

aggsender/types/fep_inputs.go
goran-ethernal: I would leave it as is, because, if something changes, it will not be so easy to detect.

aggsender/query/agg_proof_fep_inputs_query.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/1122/commits/5dfe6aec72e73131c146d579f29088cb1e7ed04e

Stefan-Ethernal (approved):

PR #1124, opened 2025-10-16
fix: determine pre-etrog bridges and encode global indexes differently

### LINKED ISSUE #1076: [BUG] Pre etrog bridges have wrong global index
for pre etrog bridges - global index for claims is wrong. we are using the new global index format which is wrong for pre etrog claims.
we need to send the correct global index for these bridges
we can just send the deposit count for these bridges.
only zkevm has these bridges.
we dont want to have a complex solution as it is not needed everywhere. only for testing
for zkevm network, we can hardcode config, and before/after a certain block, we can populate bridge accordingly

Stefan-Ethernal:
## 🔄 Changes Summary
The pre-etrog fork of the protocol assumed existence of a single rollup and back in the days, the claim global index was the `leafIndex` of certain exit tree.

Since we are exposing the global index field even for the bridges (in order to be able to link certain bridge with the claim), for pre-etrog bridges, originating from the L1, the encoding wasn't correct, because it included `mainnetFlag` indicator as well.

This PR introduce support to deduce the Etrog L1 upgrade block number, by querying the `AgglayerManager.Initialized` event and querying for the `version=2` (which belongs to Etrog). Bridges that were made prior to this block, would have assigned `depositCount` to `globalIndex`. After this block, the bridges would be encoded as:
```go
[ mainnetFlag | rollupIndex | depositCount ]
```

## ⚠️ Breaking Changes
N/A

## 📋 Config Updates
N/A

## ✅ Testing

## 🐞 Issues
- Closes #1076 

## 🔗 Related PRs
N/A

## 📝 Notes
N/A


etherman/rollup_data_querier.go
```diff
@@ -120,3 +181,10 @@ func (r *RollupDataQuerier) GetRollupData(blockNumber *big.Int) (
 
 	return rollupData, nil
 }
+
+// GetAgglayerManagerUpgradeBlock returns the rollup manager upgrade block for the given version ID.
+// If the version ID is not found, it returns false.
+func (r *RollupDataQuerier) GetUpgradeBlock(ctx context.Context, versionID uint8) (uint64, bool) {
+	blockNum, isFound := r.rollupManagerUpgradedMap[versionID]
```
goran-ethernal:
```suggestion
	return r.rollupManagerUpgradedMap[versionID]
```

config/network.go
```diff
@@ -38,6 +40,10 @@ type L1NetworkConfig struct {
 	POLTokenAddr gethcommon.Address `json:"polTokenAddress"`
 	// GlobalExitRootManagerAddr Address of the L1 GlobalExitRootManager contract
 	GlobalExitRootManagerAddr gethcommon.Address `json:"polygonZkEVMGlobalExitRootAddress"`
+	// BlocksChunkSize defines the number of blocks to be queried in each chunk when
```
goran-ethernal: Comment looks incomplete.

config/default.go
```diff
@@ -28,6 +28,8 @@ genesisBlockNumber = 0
 	polygonRollupManagerAddress = "0x0000000000000000000000000000000000000000"
 	polTokenAddress = "0x0000000000000000000000000000000000000000"
 	polygonZkEVMAddress = "0x0000000000000000000000000000000000000000"
+	BlocksChunkSize = 50000
```
goran-ethernal: This looks too big.

etherman/rollup_data_querier.go
Stefan-Ethernal: Unfortunatelly, this wouldn't work, because when indexing the map, it returns only the value.

OTOH if we decide that we don't need [this](https://github.com/agglayer/aggkit/blob/fc18b27dbb16e1be168b01ab6ba7bd76516f25f9/bridgeservice/bridge.go#L370-L371) log, then I'm fine, just returning the value. WDYT?

config/default.go
Stefan-Ethernal: Does 1000 looks good enough?  https://github.com/agglayer/aggkit/pull/1124/commits/def1407856e54942c5954b5f57040dd7656d2442

Btw, I've gotten this value, by querying against the mainnet (using the cdk contracts tooling as a playground for it) and using publicly (free) RPC endpoint. It was supporting up to 50k blocks and that's how I've got that value.

config/network.go
Stefan-Ethernal: Thanks, fixed https://github.com/agglayer/aggkit/pull/1124/commits/def1407856e54942c5954b5f57040dd7656d2442

bridgeservice/bridge.go
```diff
@@ -355,10 +365,16 @@ func (b *BridgeService) GetBridgesHandler(c *gin.Context) {
 		return
 	}
 
+	etrogUpgradeL1Block, isFound := b.agglayerManagerUpgradeQuery.GetUpgradeBlock(ctx, etrogVersionID)
```
temaniarpit27: i am not much in favor of this.
making this call on every network, on every bridge call when we know there is only 1 network

temaniarpit27: maybe call this when bridgesyncer starts and store it in some config or something like that?

Stefan-Ethernal: This is only retrieving the value from the [map](https://github.com/agglayer/aggkit/blob/def1407856e54942c5954b5f57040dd7656d2442/etherman/rollup_data_querier.go#L187-L191), and it is pretty cheap, since it is constructed only on [startup](https://github.com/agglayer/aggkit/blob/def1407856e54942c5954b5f57040dd7656d2442/etherman/rollup_data_querier.go#L63-L67). So I'd say, there is no benefit in storing it in some object, since it is already done basically (but we are storing the map)...

We can consider moving the check if the `DestinationNetwork` is `zkevm`, before fetching the upgrade blocks, and not call the `GetUpgradeBlock` function.

Btw, I feel like this PR is already an overkill, since there is only a single rollup, and the solution is not elegant either, but it is what it is.. :cry:

temaniarpit27: still it will increase RPC calls for all networks which in turn increase their cost

Stefan-Ethernal: This [part](https://github.com/agglayer/aggkit/blob/105f412d432d3877a5be66f01a07acf8054ca9ed/etherman/rollup_data_querier.go#L63-L68), where we are filtering the `Initialized` events, is invoked only once, namely on the aggkit startup.

I see your point and what I'm proposing is to hardcode the zkevm rollup id and execute it only for this network.

IMO we cannot do much better than this (other than exposing the `UpgradeToEtrogBlockNumber` to the config and populating it through the kurtosis, but I'd rather not go on that path, because I don't feel confident enough in changing the Kurtosis codebase).

WDYT?

Stefan-Ethernal: @temaniarpit27 now we are going to populate the upgrade blocks map only in case the given rollup, for which we are running the aggkit, is created pre-etrog https://github.com/agglayer/aggkit/pull/1124/commits/7a64c662454b260c36e8042d48290510a86ef098.

temaniarpit27 (approved):

bridgeservice/bridge.go
```diff
@@ -355,10 +365,16 @@ func (b *BridgeService) GetBridgesHandler(c *gin.Context) {
...
+	if !isFound {
+		b.logger.Debugf("failed to get Etrog (initializer version id=%d) L1 upgrade block, "+
```
goran-ethernal: I wouldn't even log this at all, because you are only initializing the map if `rollupID == 1 (zkevm)`. For other networks, this will always be `false`, and it can be confusing in the logs.

Stefan-Ethernal: Agreed, fixed https://github.com/agglayer/aggkit/pull/1124/commits/85932689ab1ca01d834d9c446f09982f04332059

goran-ethernal (approved):

PR #1141, opened 2025-10-21
fix: e2e aggsender require validator 

joanestebanr:
## 🔄 Changes Summary
- Change committee configuration to require at least 1 validator to sign certificate


Stefan-Ethernal (approved):

joanestebanr:
This case is no longer required because the multisign case is testted in the rest of e2e (no erigon based)

PR #1144, opened 2025-10-22
fix: validate remote validator signatures

adria0pol:
## 🔄 Changes Summary
- At this moment, external validator signatures are not checked, since these signatures are packed and sent to the agglayer node, if any validator generates a bad signature the certificate will be rejected and aggkit are not going to be able to recover from this error.


joanestebanr (approved):

aggsender/validator/remote_validator.go
```diff
@@ -76,6 +77,15 @@ func (v *RemoteValidator) ValidateAndSignCertificate(
 	certificate *agglayertypes.Certificate,
 	lastL2BlockInCert uint64,
 ) ([]byte, error) {
+	// Get the hash of the certificate, fail fast if something is wrong in it
+
```
Stefan-Ethernal, line 81: Nitpick, no need to have the empty line IMO
```suggestion
```

aggsender/validator/remote_validator.go
```diff
@@ -96,5 +106,19 @@ func (v *RemoteValidator) ValidateAndSignCertificate(
...
+	// We do not support ethereum legacy v+27 signatures
+
+	recoveredPublicKey, err := crypto.SigToPub(certificateHash[:], signature)
+	if err != nil {
+		return nil, fmt.Errorf("error validating remote validator signature: %w", err)
+	}
+
+	recoveredAddress := crypto.PubkeyToAddress(*recoveredPublicKey)
+	if v.address != recoveredAddress {
+		return nil, fmt.Errorf("error validating remote validator signature, mismatch expected:%v current:%v",
```
Stefan-Ethernal, line 117:
```suggestion
		return nil, fmt.Errorf("error validating remote validator signature, mismatch in addresses (expected: %s, recovered: %s"),
```

aggsender/validator/remote_validator.go
```diff
@@ -76,6 +77,15 @@ func (v *RemoteValidator) ValidateAndSignCertificate(
...
 	lastL2BlockInCert uint64,
 ) ([]byte, error) {
+	// Get the hash of the certificate, fail fast if something is wrong in it
+
+	certificateHash, err := HashCertificateToSign(certificate)
+	if err != nil {
+		return nil, fmt.Errorf("internal error getting certificate hash: %w", err)
+	}
+
+	// Request remote signature
```
Stefan-Ethernal: Is this comment misplaced?
```suggestion
```

aggsender/validator/remote_validator.go
```diff
@@ -96,5 +106,19 @@ func (v *RemoteValidator) ValidateAndSignCertificate(
...
+	recoveredPublicKey, err := crypto.SigToPub(certificateHash[:], signature)
+	if err != nil {
+		return nil, fmt.Errorf("error validating remote validator signature: %w", err)
+	}
+
+	recoveredAddress := crypto.PubkeyToAddress(*recoveredPublicKey)
+	if v.address != recoveredAddress {
+		return nil, fmt.Errorf("error validating remote validator signature, mismatch expected:%v current:%v",
+			v.address, recoveredAddress)
+	}
```
Stefan-Ethernal, line 119: Does it make sense to run this check immediately after we have calculated the certificate hash, WDYT?
https://github.com/agglayer/aggkit/blob/a1ed47ce82dbc3f13155acaab0bb9347f00f8e53/aggsender/validator/remote_validator.go#L82-L85

aggsender/validator/remote_validator.go
adria0pol: Apply changes on @Stefan-Ethernal review

aggsender/validator/remote_validator.go
adria0pol: Getting a different address doesn’t necessarily mean that the recovered address is the actual signer of the certificate, so I’m a bit hesitant to use the term “address” here.

aggsender/validator/remote_validator.go
adria0pol: ~Agree!~ 

update:

I execute it here because if it fails we do not have to do an external call for verifying the certificate. For sure what you say favours the readeability via code blocks.

Stefan-Ethernal (approved):

PR #1145, opened 2025-10-22
fix: `imported-bridge-exit` commitment hash in optimistic mode

goran-ethernal:
## 🔄 Changes Summary
This PR fixes the `imported bridge exit` commitment hash calculation in the `optimistic` mode.

`imported bridge exits` are consisted of `bridge exit` part, `global index` and `claim data`. When creating the `bridge exit` part from the `Claim` event, we need to do `keccak256` on the `Metadata` field of the `Claim`, like it is done for the certificates, and in `interop` repo [here](https://github.com/agglayer/interop/blob/6f892711bb533dfb9d4bc6b3717db77ad06c9634/crates/unified-bridge/src/bridge_exit.rs#L66).

While calculating the hash of the `imported bridge exits` commitment in the `optimistic` mode, when we recreated `bridge exit` part from the `Claim`, we did not do `keccak256` on the `Metadata` field, like we do for certificates.

This PR fixes this, and makes the conversion of the `bridge exit` part of the `Claim` use the same code as we do when building certificates. This fixes the hash calculation on the `optimistic` mode.

## ⚠️ Breaking Changes
NA

## 📋 Config Updates
NA

## ✅ Testing

### DIFF
--- a/aggsender/optimistic/optimistichash/calculate_hash_commit_imported_bridges.go
+++ b/aggsender/optimistic/optimistichash/calculate_hash_commit_imported_bridges.go
@@ -3,7 +3,7 @@ package optimistichash
 import (
 	"math/big"
 
-	agglayertypes "github.com/agglayer/aggkit/agglayer/types"
+	"github.com/agglayer/aggkit/aggsender/converters"
 	"github.com/agglayer/aggkit/bridgesync"
 	aggkitcommon "github.com/agglayer/aggkit/common"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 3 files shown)


aggsender/optimistic/optimistichash/calculate_hash_commit_imported_bridges.go
```diff
@@ -51,21 +51,6 @@ func (o *optimisticCommitImportedBrigesData) hash() common.Hash {
...
-		TokenInfo: &agglayertypes.TokenInfo{
-			OriginNetwork:      claim.OriginNetwork,
-			OriginTokenAddress: claim.OriginAddress,
-		},
-		DestinationNetwork: claim.DestinationNetwork,
-		DestinationAddress: claim.DestinationAddress,
-		Amount:             claim.Amount,
-		Metadata:           claim.Metadata,
-	}
+	be := converters.ConvertBridgeExitFromClaim(claim)
```
vcastellm, line 54: Shouldn't we use the `ConvertBridgeExitFromClaim` from optimistic and non-optimistic?

goran-ethernal: I am not sure what you mean @vcastellm ?

goran-ethernal: There is no difference, optimistic, non-optimistic mode, bridge exit struct is the same, and its hash is calculated the same for both modes. This just uses the right code for conversion, which we use in certificates, as I explained in the PR description.

Stefan-Ethernal (approved):

temaniarpit27 (approved):

PR #1148, opened 2025-10-23
feat: run the l1 info tree syncer with `safe` block finality

### LINKED ISSUE #1170: Aggoracle should process GERs at a faster pace
The intermediate approach is to rely on `safe` finalization (instead of `finalized`) in the l1 info tree syncer component, to figure out what is the latest GER, that should be injected.

Re-introduce `L1InfoTreeSync.BlockFinality`, because l1 info tree sync is used by multiple components, so it should be possible to configure the finalization

Stefan-Ethernal:
## 🔄 Changes Summary
- Change the L1 Info Tree Syncer block finality to `safe`, in order to make `aggoracle` runs faster. Re-introduce `BlockFinality` in the `L1InfoTreeSyncer` config and consequently the `ReorgDetector`.
- Simplify `keccak256` hash calculations, by relying on go ethereum library.
- Remove `URLRPCL1` config parameters from `Aggoracle` and `L1InfoTreeSync`, since they are unused.
- Introduce abstractions for merkle trees, in order to test l1 info tree syncer processor more deeply:
    - `ReadTreer` – provides read-only access to tree data (roots, leaves, proofs).
    - `LeafWriter` – defines write operations for adding new leaves to the tree.
    - `ReorganizeTreer` – extends ReadTreer with the ability to handle blockchain reorgs.
    - `FullTreer` – combines all capabilities (`ReadTreer`, `LeafWriter`, `ReorganizeTreer`), representing a fully functional, mutable Merkle tree.

## ⚠️ Breaking Changes
N/A

## 📋 Config Updates
- Re-introduce `L1InfoTreeSync.BlockFinality` config param.
- Remove `Aggoracle.URLRPCL1` config param.
- Remove `L1InfoTreeSync.URLRPCL1` config param.

## ✅ Testing

## 🐞 Issues
- Closes #1170 

## 🔗 Related PRs
N/A

## 📝 Notes
N/A


goran-ethernal (approved):

cmd/run.go
```diff
@@ -526,10 +526,9 @@ func runL1InfoTreeSyncerIfNeeded(
 	l1InfoTreeSync, err := l1infotreesync.New(
 		ctx,
 		cfg.L1InfoTreeSync,
-		aggkittypes.FinalizedBlock,
 		l1Client,
 		l1infotreesync.FlagNone,
-		aggkittypes.FinalizedBlock,
+		aggkittypes.SafeBlock,
```
joanestebanr: This param maybe make sense to be configurable, WDYT? 
Currently this is going to procude a warning [here](https://github.com/agglayer/aggkit/blob/7030cccf7d64dbf901da3c9145d2cae35eccc2de/sync/evmdownloader.go#L75). Maybe this can be change to greater than Safe, but allow to define in the config file also, maybe just finalized or safe or something like that

Stefan-Ethernal: Re-introduced the reorg detector for l1 info tree syncer and providing `FinalizedBlock` from `ReorgDetector` cfg (https://github.com/agglayer/aggkit/pull/1148/commits/592185ed769f47f417b4f31a49d43df658b5fad4).

l1infotreesync/config.go
```diff
@@ -2,19 +2,29 @@ package l1infotreesync
...
-	InitialBlock               uint64         `mapstructure:"InitialBlock"`
-	RetryAfterErrorPeriod      types.Duration `mapstructure:"RetryAfterErrorPeriod"`
-	MaxRetryAttemptsAfterError int            `mapstructure:"MaxRetryAttemptsAfterError"`
+	// DBPath is the path of the database where the L1 Info Tree data will be stored
+	DBPath string `mapstructure:"DBPath"`
+	// GlobalExitRootAddr is the address of the GlobalExitRoot manager contract on L1
+	GlobalExitRootAddr common.Address `mapstructure:"GlobalExitRootAddr"`
+	// RollupManagerAddr is the address of the RollupManager/AgglayerManager contract
+	RollupManagerAddr common.Address `mapstructure:"RollupManagerAddr"`
+	// Possible values: LatestBlock, SafeBlock, PendingBlock, FinalizedBlock, EarliestBlock
```
joanestebanr: Missed description of the field, also the possible values missed the possibility to set an offset (e.g `FinalizedBlock/10`)

config/default.go
```diff
@@ -93,8 +93,8 @@ FinalizedBlock = "LatestBlock"
 DBPath = "{{PathRWData}}/L1InfoTreeSync.sqlite"
 GlobalExitRootAddr = "{{L1NetworkConfig.GlobalExitRootManagerAddr}}"
 RollupManagerAddr = "{{L1NetworkConfig.RollupManagerAddr}}"
+BlockFinality = "SafeBlock"
```
joanestebanr, line 96: are we sure that we want `Safe` as default? the situation of changing to `Safe` is for always or just an specific use case?

config/config.go
```diff
@@ -191,10 +190,6 @@ var (
 			FieldNamePattern: "AggOracle.BlockFinality",
 			Reason:           aggOracleBlockFinalityDeprecated,
 		},
-		{
-			FieldNamePattern: "L1InfoTreeSync.BlockFinality",
-			Reason:           l1InfoTreeSyncBlockFinalityDeprecated,
-		},
 		{
 			FieldNamePattern: "LastGERSync.SyncMode",
```
joanestebanr, line 195: Maybe make sense to introduce the fields  `Aggoracle.URLRPCL1` and `L1InfoTreeSync.URLRPCL1` as deprecated?

joanestebanr (commented):
The coverage is below 80%

config/default.go
Stefan-Ethernal: The original requirement was to make the `safe`, in order to make the `aggoracle` runs faster (and consequentially be able to inject GERs at a higher frequency). Also I understand this is just an intermediary approach until syncers and reorg detector is refactored. Note that, in the near future, we are aiming to have GERs injected on every 6th L1 block, which is considered as good enough (and thus probably unlikely to be reorged). As you can see on the [Etherscan](https://etherscan.io/blocks_forked?ps=100&p=1), the reorgs are happening, but are often pretty shallow (most often the fork is only 1 block deep)

As it is highly unlikely that the `safe` block gets reorged, I think it makes sense to run on this `BlockFinality` by default. If for any reason we experience some issues, we can always override it through a config.

IMO it would be an overkill at this point to refactor this, and create a separate configuration for the `aggoracle` on the Kurtosis CDK and what not.

cc @arnaubennassar

config/config.go
Stefan-Ethernal: Fixed https://github.com/agglayer/aggkit/pull/1148/commits/7485d4c8202d2c2ad8b2a28ba2bb7c66fb8d864a

l1infotreesync/config.go
Stefan-Ethernal: Fixed https://github.com/agglayer/aggkit/pull/1148/commits/7485d4c8202d2c2ad8b2a28ba2bb7c66fb8d864a

config/default.go
vcastellm: I'm in for setting this as Default, the "highly unlikely" in the `safe` definition is quite high.

> No, safe blocks have not been reorged in Ethereum. A "safe" block is one that has received attestations from two-thirds of the validator set, making it highly unlikely to be reorganized under normal network conditions. Reorganizations of "safe" blocks would require a catastrophic attack on the network, as it is understood to be crypto-economically secure.

vcastellm (approved):
LGTM

PR #1199, opened 2025-10-31
feat: validate positive offset in block finality

### LINKED ISSUE #1198: Validate Block finality offset
Validate the `offset` configuration based on the `blockFinality` value.
Each `blockFinality` type imposes a strict upper limit on the positive `offset` that can be set.

rachit77:
## 🔄 Changes Summary
- This PR adds validation for positive offsets in block finality.

## 🐞 Issues
- Closes #1198


cmd/run.go
```diff
@@ -66,6 +66,26 @@ func start(cliCtx *cli.Context) error {
 		return fmt.Errorf("invalid L2 RPC config: %w", err)
 	}
 
+	if err := cfg.L2GERSync.Validate(); err != nil {
```
joanestebanr: I think that doesn't make sense to test all of them, just the ones that are going to be instantiated (related to components). So I suggest to move the validate in the constructor of each component instead of `run.go`

reorgdetector/config.go
```diff
@@ -23,6 +24,14 @@ type Config struct {
 	FinalizedBlock aggkittypes.BlockNumberFinality `jsonschema:"enum=LatestBlock, enum=SafeBlock, enum=PendingBlock, enum=FinalizedBlock, enum=EarliestBlock" mapstructure:"FinalizedBlock"` //nolint:lll
 }
 
+// Validate checks if the configuration is valid
+func (c *Config) Validate() error {
+	if err := c.FinalizedBlock.ValidateOffset(); err != nil {
```
joanestebanr: For L2 are you sure that apply same limits than L1? maybe op-succinct have another values and only make sense to apply this validation for L1

types/block_finality.go
```diff
@@ -120,6 +123,47 @@ func (b BlockNumberFinality) IsLatest() bool {
...
 }
 
+// ValidateOffset validates that the positive offset does not exceed the maximum allowed
+// for the specific block finality type. Each block finality type has a strict upper limit
+// on the positive offset that can be set:
+//   - LatestBlock: cannot have positive offset (limit = 0)
+//   - PendingBlock: cannot have positive offset (limit = 0) as pending blocks don't exist yet
+//   - SafeBlock: maximum positive offset is MaxPositiveOffsetSafe
+//   - FinalizedBlock: maximum positive offset is MaxPositiveOffsetFinalized (most restrictive)
+func (b BlockNumberFinality) ValidateOffset() error {
```
joanestebanr: Mabe `Validate()` is a better name, because it's the common name for this kind of job and  validates the struct, what is real checked is just a implementation detail. Think that maybe you could check that the 'blockname' is valid

joanestebanr (approved):
Pending to pass coverage

goran-ethernal (approved):

types/block_finality.go
```diff
@@ -120,6 +123,44 @@ func (b BlockNumberFinality) IsLatest() bool {
...
+	switch b.Block {
+	case Latest:
+		maxOffset = MaxPositiveOffsetLatest
+	case Pending:
+		maxOffset = MaxPositiveOffsetPending
+	case Safe:
+		maxOffset = MaxPositiveOffsetSafe
+	case Finalized:
+		maxOffset = MaxPositiveOffsetFinalized
+	}
```
Stefan-Ethernal, line 152: Would it make sense to add the `maxOffset` field to the `BlockNumberFinality` [struct](https://github.com/agglayer/aggkit/blob/4656503450b9c23af2cc8302ef59ed6c82c4fdd6/types/block_finality.go#L41-L44)? The idea I have is in order to get rid off this `switch-case`.

It is instantiated [here](https://github.com/agglayer/aggkit/blob/4656503450b9c23af2cc8302ef59ed6c82c4fdd6/types/block_finality.go#L34-L37) and [here](https://github.com/agglayer/aggkit/blob/4656503450b9c23af2cc8302ef59ed6c82c4fdd6/types/block_finality.go#L49), other than unit tests, so hopefully it is feasible to make such change.

WDYT?

cmd/run.go
```diff
@@ -690,6 +693,10 @@ func runBridgeSyncL1IfNeeded(
 		return nil
 	}
 
+	if err := cfg.Validate(); err != nil {
+		log.Fatalf("invalid BridgeL1Sync config: %v", err)
+	}
```
Stefan-Ethernal, line 698: Don't you need to validate it for the `BridgeL2Syncer` as well?

cmd/run.go
```diff
@@ -606,6 +606,9 @@ func runReorgDetectorL1IfNeeded(
 		components) {
 		return nil, nil
 	}
+	if err := cfg.Validate(); err != nil {
+		log.Fatalf("invalid ReorgDetectorL1 config: %v", err)
+	}
```
Stefan-Ethernal, line 611: How about `ReorgDetectorL2` config validation?

cmd/run.go
rachit77: For L2s, we’re not certain whether they follow the same commitment levels as Ethereum. Moreover, different L2s might define these levels differently for instance, OP could have one set of levels, while zkEVM/Validium might use another, and any new L2 joining our network could introduce its own variations

So, I’d prefer not to hardcode those values

cmd/run.go
rachit77: same as above answer

vcastellm (approved):
LGTM

types/block_finality.go
rachit77: Thanks for the suggestion @Stefan-Ethernal!

I’d prefer to keep the existing `BlockNumberFinality` struct since I find it more intuitive. 
Adding a maxOffset field could introduce inconsistency if someone manually creates a struct with mismatched values in future.

Stefan-Ethernal (approved):

PR #1213, opened 2025-11-01
fix: record timestamp after sleep and recheck in wrapper

radik878:
Problem:
- RateLimit.Call() was appending a pre-sleep timestamp both in internal- and external-sleep flows.
- This mis-accounted requests and tightened the effective sliding window, causing unnecessarily strict throughput.
- The agglayer wrapper slept externally but the timestamp had already been recorded before the sleep.

Changes:
- In common/rate_limit.go, record the timestamp only when the request is actually allowed to proceed "now":
  * For blocking calls: sleep first, then recompute now, cleanup, and append.
  * For non-blocking calls: return the sleep duration without appending.
- In agglayer/rate_limit_wrapper.go, after external sleep, re-check and record using a loop to remain correct under concurrency.

Why:
- Aligns with the documented behavior: the system waits until the next interval before allowing more requests.
- Ensures precise sliding-window accounting and stable throughput, especially under concurrent load.

### DIFF
--- a/agglayer/rate_limit_wrapper.go
+++ b/agglayer/rate_limit_wrapper.go
@@ -61,6 +61,16 @@ func (r *RateLimitWrapper) applyRateLimit(methodName string) {
 			methodName,
 			rateLimitSleepTime.String(), rateLimiter.String())
 		time.Sleep(*rateLimitSleepTime)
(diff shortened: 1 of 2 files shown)


temaniarpit27:
Thanks a lot for raising the PR but at this point of time we are unable to accept external contributions
We will try to incorporate them in our upcoming releases

Issue #1225, opened 2025-11-03
feat: new_sync: Adapt l1infotree syncer to Multidownloader


joanestebanr:
# Feature Request
Adapt l1infotree syncer to Multidownloader

Issue #1229, opened 2025-11-03
feat: new_sync: compare performace


joanestebanr:
# Feature Request
Check the performance vs previous implementation for historical syncing

PR #1234, opened 2025-11-03
feat: multidownloader safe synchronizations

### LINKED ISSUE #1226: feat: new_sync: Implement safe syncing
# Feature Request
Implement safe syncing

joanestebanr:
## 🔄 Changes Summary
- Implements safe syncing. #1226
    -  It group registered syncers and start downloading in parallel logs + blocks
    - To download blocks use Batch Requests if availaible
- Partial `L1infotree` integration Multidownloader  (just downloader, but no reorg) #1225

- Internal refactors to take advantatge of existing objects or to avoid import cycles: 
    - `block_notifier.go` and `block_range` moved to `common`
    - `block_notifier_*` moved to `etherman`
    - `config/network.go` moved to `etherman/config`
    - `etherman/rollup_data_querier.go` moved to `etherman/querier` to avoid cycles
    - 

## 📋 Config Updates
- 🧾 **Diff/Config snippet**:  
- Added new config:
```
[L1Multidownloader]
	Enabled = false
	StoragePath = "{{PathRWData}}/l1_multidownloader.sqlite"
	BlockChunkSize = 10000
	MaxParallelBlockHeaderRetrieval = 30
	BlockFinality = "FinalizedBlock"
	WaitPeriodToCheckCatchUp = "10s"
[L2Multidownloader]
	Enabled = false
	StoragePath = "{{PathRWData}}/l2_multidownloader.sqlite"
	BlockChunkSize = 10000
	MaxParallelBlockHeaderRetrieval = 30
	BlockFinality = "FinalizedBlock"
	WaitPeriodToCheckCatchUp = "10s"
```

## ✅ Testing

## 🐞 Issues
- Closes #1226

## 📝 Notes
- This is a WIP, so there are some code commented that belong to next implementation steps
-


multidownloader/types/syncer_config.go
```diff
@@ -0,0 +1,177 @@
...
+type SyncerID = string
+
+type SyncerConfig struct {
+	// SyncerID is the unique identifier for the syncer
+	SyncerID SyncerID
+	// ContractAddr is list of contract addresses to sync
+	ContractsAddr []common.Address
+	// Starting block
+	FromBlock uint64
+	// Taget for final block
```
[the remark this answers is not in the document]
joanestebanr: fixed

multidownloader/config.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func (cfg *Config) String() string {
+	return "MultidownloaderConfig{" +
+		"Enabled:" + fmt.Sprintf("%t", cfg.Enabled) +
+		", BlockChunkSize:" + string(rune(cfg.BlockChunkSize)) +
+		", MaxParallelBlockHeaderRetrieval:" + string(rune(cfg.MaxParallelBlockHeaderRetrieval)) +
+		", BlockFinality:" + cfg.BlockFinality.String() +
+		", WaitPeriodToCheckCatchUp:" + cfg.WaitPeriodToCheckCatchUp.String() +
+		"}"
```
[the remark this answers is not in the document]
joanestebanr: factored using fmt.Sprintf

common/common.go
```diff
@@ -152,3 +154,16 @@ func MapSlice[T any, R any](in []T, f func(T) R) []R {
 	}
 	return out
 }
+
+const hexBase = 16
+
+func ParseHexUint64(hexStr string) (uint64, error) {
```
goran-ethernal: I think we should support the regular conversion of decimal values as well. I see this is used when parsing the block number from an ethereum node's `json rpc`, but there can be nodes that for lower blocks return decimal values (I've seen this in the old Edge days).

Take a look at this Polygon Edge function that supports both cases:
https://github.com/0xPolygon/polygon-edge/blob/0504fa79632b0a2797a7598d87f6743cb9a1cd01/helper/common/encoding.go#L14

etherman/batch_requests.go
```diff
@@ -0,0 +1,215 @@
...
+	"sync"
+
+	aggkitcommon "github.com/agglayer/aggkit/common"
+	aggkittypes "github.com/agglayer/aggkit/types"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/rpc"
+)
+
+// blockRawEth is eth_getBlockByNumber result structure
+type blockRawEth struct {
```
goran-ethernal, line 17: Looks like `types` package in `etherman` is a good place for this struct. WDYT?

etherman/batch_requests.go
```diff
@@ -0,0 +1,215 @@
...
+func retrieveBlockHeadersInBatchParallel(
+	logger aggkitcommon.Logger,
+	funcRetrieval func(map[uint64]struct{}) (map[uint64]*aggkittypes.BlockHeader, error),
+	blockNumbers map[uint64]struct{},
+	chunckSize, maxConcurrency int) (map[uint64]*aggkittypes.BlockHeader, error) {
+	results := make(map[uint64]*aggkittypes.BlockHeader)
+	errs := make([]error, 0)
+	var mu sync.Mutex
+	var wg sync.WaitGroup
+	sem := make(chan struct{}, maxConcurrency)
```
goran-ethernal: You don't need this semaphore pattern to limit active go routines. Newer `Go` versions support an `errorGroup` with limit set of active go routines. You just need to pass a context to the `retrieveBlockHeadersInBatchParallel`. Something like this:

```go
	g, _ := errgroup.WithContext(ctx)
	g.SetLimit(maxConcurrency) // Built-in concurrency limiting!

	for _, chunk := range chunks {
		g.Go(func() error {
			headers, err := funcRetrieval(chunk)
			if err != nil {
				return fmt.Errorf("RetrieveBlockHeadersInBatchParallel: %w", err)
			}

			mu.Lock()
			for blockNumber, header := range headers {
				results[blockNumber] = header
			}
			mu.Unlock()
			return nil
		})
	}

	if err := g.Wait(); err != nil {
		return results, err
	}
```

etherman/batch_requests.go
```diff
@@ -0,0 +1,215 @@
...
+	logger.Debugf("retrieveRPCBlockHeadersInParallel: Retrieved block headers for blocks %d",
+		len(blockNumbers))
+	if len(errs) > 0 {
+		return results, fmt.Errorf("retrieveRPCBlockHeadersInParallel: errors: %v", errs)
+	}
+	return results, nil
+}
+
+func splitBlockNumbersIntoChunks(blockNumbers map[uint64]struct{}, chunkSize int) []map[uint64]struct{} {
+	chunks := make([]map[uint64]struct{}, 0)
```
goran-ethernal: I think you can allocate the size of the `chunks` array by doing this:
```go
numChunks := (len(blockNumbers) + chunkSize - 1) / chunkSize
chunks := make([]map[uint64]struct{}, 0, numChunks)
currentChunk := make(map[uint64]struct{}, chunkSize) // even if the chunk doesn't have the entire chunk size, it is ok to set the capacity of the map to a greater value, there are no fantom elements added to it, just the ones we actually add
```
You can also remove the `count` variable and just use the `len(currentChunk)`:

```go
 for bn := range blockNumbers {
     currentChunk[bn] = struct{}{}
     if len(currentChunk) == chunkSize {
         chunks = append(chunks, currentChunk)
         currentChunk = make(map[uint64]struct{}, chunkSize)
     }
}
```

etherman/block_notifier/block_notifier_manager.go
```diff
@@ -0,0 +1,65 @@
...
+func NewBlockNotifierManager(logger aggkitcommon.Logger,
+	constructorFunc func(aggkittypes.BlockNumberFinality) (ethermantypes.BlockNotifier, error)) *BlockNotifierManager {
+	return &BlockNotifierManager{
+		blockNotifiers:  make(map[aggkittypes.BlockNumberFinality]ethermantypes.BlockNotifier),
+		constructorFunc: constructorFunc,
+		logger:          logger,
+	}
+}
+
+// TODO: You must only have real blockNotifiers for latest, safe, finalized...
```
goran-ethernal, line 32: Is this TODO still relevant?

etherman/block_notifier/block_notifier_polling.go
```diff
@@ -76,6 +77,13 @@ func (b *BlockNotifierPolling) String() string {
 	return res
 }
 
+func (b *BlockNotifierPolling) Initialize(ctx context.Context) error {
```
goran-ethernal, line 82: Why do we need this function? We didn't need it before in the `aggsender` when we used the block notifier.

etherman/block_notifier/block_notifier_polling.go
```diff
@@ -98,6 +106,9 @@ func (b *BlockNotifierPolling) Start(ctx context.Context) {
 		}
 	}
 }
+func (b *BlockNotifierPolling) BlockFinality() aggkittypes.BlockNumberFinality {
```
goran-ethernal: This is not used anywhere.

etherman/types/block_notifier.go
```diff
@@ -20,4 +20,6 @@ type BlockNotifier interface {
 	GetCurrentBlockNumber() uint64
 	String() string
 	Start(ctx context.Context)
+	BlockFinality() aggkittypes.BlockNumberFinality
```
goran-ethernal: Same. Not used anywhere.

etherman/batch_requests.go
joanestebanr: It's a private struct that it's only used in this file, for now... maybe if became public or are used my other parts make sense to move there

config/default.go
```diff
@@ -319,4 +319,20 @@ AgglayerBridgeL2Addr = "{{L2Config.BridgeAddr}}"
...
+	MaxParallelBlockHeaderRetrieval = 30
+	BlockFinality = "FinalizedBlock"
+	WaitPeriodToCheckCatchUp = "10s"
+
+[L2Multidownloader]
+	Enabled = false
+	StoragePath = "{{PathRWData}}/l2_multidownloader.sqlite"
+	BlockChunkSize = 10000
+	MaxParallelBlockHeaderRetrieval = 30
+	BlockFinality = "FinalizedBlock"
```
goran-ethernal: Currently `BridgeL2Sync` has `LatestBlock` set as block finality, do we want it set for the `L2Multidownloader` as well?

common/common.go
joanestebanr: good idea!, 9bee93f467bb35ac1977b5283bc3a9dda3fa31fb

etherman/batch_requests.go
joanestebanr: thank for the suggestion! 071584907b9a4262aa7208000effea48cc23ce56

etherman/block_notifier/block_notifier_manager.go
joanestebanr: yes, the idea is have a source of latest, safe,... and the rest of BlockFinality with offset be calculated using this to reduce the numbers of RPC calls

etherman/block_notifier/block_notifier_polling.go
joanestebanr: Becasue now the object catch the `blockNumber` at start, and to avoid race conditions I introduce the 'Initilaize` function that populate this number and you can call `GetCurrentBlockNumber`

etherman/block_notifier/block_notifier_polling.go
joanestebanr: fe19144790d50f48ea9e8fa63aeefcc2a74489a2

etherman/types/block_notifier.go
joanestebanr: fe19144790d50f48ea9e8fa63aeefcc2a74489a2

etherman/batch_requests.go
joanestebanr: good improvement! 81b1e04c5dc63805164353155026fe0df3d9360f

etherman/block_notifier/block_notifier_manager.go
goran-ethernal: I still do not understand completely. Does that mean we can not use block notifiers for block finality with offset?

multidownloader/config.go
```diff
@@ -0,0 +1,82 @@
...
+	BlockChunkSize uint32
+	// MaxParallelBlockHeaderRetrieval is the maximum number of parallel RPC calls to retrieve block headers
+	MaxParallelBlockHeaderRetrieval int
+	// BlockFinality is which block consider final (typically finalizedBlock)
+	BlockFinality aggkittypes.BlockNumberFinality
+	// WaitPeriodToCheckCatchUp is the wait applied when a request is still not available to check again
+	WaitPeriodToCheckCatchUp types.Duration
+}
+
+const defaultBlockChunkSize = 10000
```
goran-ethernal:
```suggestion
const (
    defaultBlockChunkSize = 10000
    defaultMaxParallelBlockHeaderRetrieval = 30
    defaultWaitPeriodToCheckCatchUp = time.Second * 10
)
```

multidownloader/evm_multidownloader.go
```diff
@@ -0,0 +1,607 @@
...
+)
+
+const safeMode = true
+
+// const unsafeMode = false
+
+type EVMMultidownloader struct {
+	log                  aggkitcommon.Logger
+	cfg                  Config
+	ethClient            aggkittypes.BaseEthereumClienter
```
goran-ethernal, line 35: Do we need these two separations to `ethClient` and `rpcClient` since `aggkittypes.EthClienter` already implements both interfaces?

multidownloader/evm_multidownloader.go
```diff
@@ -0,0 +1,607 @@
...
+	// TODO: Implement unsafe mode syncing
+	// err = dh.Sync(ctx, dh.StepUnsafe, "unsafe")
+	// if err != nil {
+	// 	return err
+	// }
+
+	return nil
+}
+
+func (dh *EVMMultidownloader) GetRPCServices() []jRPC.Service {
```
goran-ethernal, line 124: I don't see this being used (exposed) anywhere.

multidownloader/evm_multidownloader.go
```diff
@@ -0,0 +1,607 @@
...
+	}
+	return nil
+}
+
+func ethGetExtendendError(err error) string {
+	if err == nil {
+		return ""
+	}
+
+	jsonError, ok := err.(ethrpc.DataError) //nolint:errorlint
```
goran-ethernal: Maybe use `errors.As` to not have to define the `nolint` rule:
```go
var jsonError ethrpc.DataError
if !errors.As(err, &jsonError) {
    return ""
}
```
It handles wrapped errors as well.

multidownloader/evm_multidownloader.go
```diff
@@ -0,0 +1,607 @@
...
+		blocks, dh.cfg.MaxParallelBlockHeaderRetrieval)
+	if err != nil {
+		return false, fmt.Errorf("Safe/Step: cannot retrieve block headers (%d): %w", len(blockHeaders), err)
+	}
+
+	// Calculate new state (not set in memory until commit is successful)
+	dh.mutex.Lock()
+	newSyncedSegments := dh.syncedSegments.Clone()
+	newPendingSegments := dh.pendingSync.Clone()
+	dh.mutex.Unlock()
```
goran-ethernal, line 274: Hmmm, there is a lot of critical processing done after this lock, maybe it would be more safe to just hold the lock until we store the data in the `db`. This shouldn't take too much time to block any other routine. WDYT?

multidownloader/evm_multidownloader.go
```diff
@@ -0,0 +1,607 @@
...
+func extractSuggestedBlockRangeFromError(err error) *aggkitcommon.BlockRange {
+	if !isEthClientErrorTooManyResults(err) {
+		return nil
+	}
+	msg := ethGetExtendendError(err)
+	return extractSuggestedBlockRangeFromErrorMsg(msg)
+}
+
+func extractSuggestedBlockRangeFromErrorMsg(msg string) *aggkitcommon.BlockRange {
+	re := regexp.MustCompile(`\[([^\]]+)\]`)
```
goran-ethernal, line 385: Hopefully the different rpc providers have the same error formatting that includes the block number. I am not fond of relying on them to behave the same when it comes to this. Just a concern from my side, we do not have to change the current code.

multidownloader/evm_multidownloader.go
```diff
@@ -0,0 +1,607 @@
...
+
+// sync it's an internal function that executes the given stepFunc until it returns done=true or error
+func (dh *EVMMultidownloader) sync(ctx context.Context,
+	stepFunc func(ctx context.Context) (bool, error), name string) error {
+	dh.statistics.StartSyncing()
+
+	iteration := 0
+	dh.log.Infof("🚀🚀🚀🚀🚀🚀 start syncing %s ...", name)
+	// Execute steps until done or error
+	for done, err := stepFunc(ctx); !done; done, err = stepFunc(ctx) {
```
goran-ethernal, line 207: Seems like we might be missing a context cancellation propagation here if the main context is done. I do not see this in the step function as well.

multidownloader/evm_multidownloader.go
```diff
@@ -0,0 +1,607 @@
...
+// 	err = dh.pendingSync.UpdateToBlock(ctx, dh.blockNotifierManager)
+// 	if err != nil {
+// 		return false, fmt.Errorf("Unsafe/Step: cannot update ToBlock in pendingSync: %w", err)
+// 	}
+// 	return dh.pendingSync.Finished(), nil
+// }
+
+func getBlockNumbers(logs []types.Log) []uint64 {
+	blockNumbers := make(map[uint64]struct{})
+	result := make([]uint64, 0)
```
goran-ethernal, line 228: This can be quite inefficient if we have a lot of logs in a lot of blocks (which will happen when we start syncing from scratch). I think we can add at least some heuristic capacity to the slice, so that it doesn't resize non stop.

```go
blockNumbers := make(map[uint64]struct{}, len(logs)/10) // some heuristic capacity
result := make([]uint64, 0, len(logs)/10) // some heuristic capacity
```

WDYT?

multidownloader/evm_multidownloader.go
```diff
@@ -0,0 +1,607 @@
...
+	return result
+}
+
+func (dh *EVMMultidownloader) IsAvailable(query mdrtypes.LogQuery) bool {
+	dh.mutex.Lock()
+	defer dh.mutex.Unlock()
+	return dh.syncedSegments.IsAvailable(query)
+}
+
+func mapBlockHeadersToList(blocks map[uint64]*aggkittypes.BlockHeader) []*aggkittypes.BlockHeader {
```
goran-ethernal: I am not sure why do we need this function. We turn the map to a slice, and then in `SaveEthLogsWithHeaders` we turn it to a map again. Maybe just pass it as a map to `SaveEthLogsWithHeaders` and then convert it to a map you expect. WDYT?

config/default.go
joanestebanr: I agree, the L2 block finality must be `latest`

multidownloader/evm_multidownloader.go
joanestebanr: yes, because the simulated environment in e2e doesn't support RPC client, so you can pass ethClient but not rpcClient and then, instead of using batch call is using parallel calls to retrieve blocks

multidownloader/evm_multidownloader.go
joanestebanr: good catch! you're right it's missing exposing the end-points in `cmd.go`

joanestebanr: a2298ce9faa7450d2fd95e241cc7385e879dd9cb

multidownloader/evm_multidownloader.go
joanestebanr: I totally share your concern, for that reason there are a fallback mechanism that half the chunk if it's not able to infer from RPC response

multidownloader/evm_multidownloader.go
joanestebanr: The idea is to reduce the block time, because in parallel there are syncer asking for data:
- The only function that modify this fields `syncedSegments` and `pendingSync` is this one
- All DB modification is done using a Tx
- After commit it mutex again and set the new fields to the object
In the worst case, an syncer could ask for a BlockRange that is already at DB but the `syncedSegments` is still not updated and have to wait.

multidownloader/evm_multidownloader.go
joanestebanr: Good point! done: 073d8e8b0b30c9d931ad2e8d4c5fe4d54b488b76

common/block_range.go
```diff
@@ -88,3 +88,59 @@ func getBlockMinusOne(fromBlock uint64) uint64 {
...
+
+// Merge merges two BlockRanges into one encompassing BlockRange.
+func (b BlockRange) Merge(other BlockRange) BlockRange {
+	return NewBlockRange(
+		min(b.FromBlock, other.FromBlock),
+		max(b.ToBlock, other.ToBlock),
+	)
+}
+
+// Subtract a BlockRanges
```
vcastellm:
```suggestion
// Subtract two BlockRanges
```

multidownloader/evm_multidownloader.go
joanestebanr: It's a good optimitization, but I think that maybe it can be done in the future... right now it takes 15min to sync l1infotree from Sepolia. I think that is a good improvement and the idea is after fully implement everything do an optimization after executing tasks #1229

multidownloader/evm_multidownloader.go
joanestebanr: I prefer slices, because maps don't garantee the order, so I change everything to slice avoiding the need of transformations and removing `mapBlockHeadersToList`

multidownloader/storage/migrations/0001.sql
```diff
@@ -0,0 +1,41 @@
...
+    synced_from_block BIGINT NOT NULL,  -- Current synced from block
+    synced_to_block BIGINT NOT NULL,    -- Current synced to block
+    syncers_id TEXT NOT NULL,          -- Syncer identifier
+    PRIMARY KEY (contract_address)
+);
+
+CREATE TABLE reorg_chain (
+    chain_id INTEGER NOT NULL,
+    block_number BIGINT NOT NULL,
+     block_hash TEXT NOT NULL,
```
vcastellm:
```suggestion
    block_hash TEXT NOT NULL, 
```

multidownloader/storage/migrations/0001.sql
```diff
@@ -0,0 +1,41 @@
...
+    topics TEXT NOT NULL,                 -- list of hashes in JSON
+    data BLOB,                            -- 
+    block_number BIGINT NOT NULL REFERENCES block(block_number),
+    tx_hash TEXT NOT NULL,
+    tx_index INTEGER NOT NULL,
+    log_index INTEGER NOT NULL,      -- “index” is a reserved keyword
+    PRIMARY KEY (address, block_number, log_index)
+);
+
+CREATE TABLE block (
```
vcastellm: Better rename to `blocks`, maintaining consistency with `logs`

multidownloader/storage/migrations/0001.sql
```diff
@@ -0,0 +1,41 @@
...
+    contract_address TEXT NOT NULL,              -- Contract address       
+    target_from_block BIGINT NOT NULL,  -- Desired from block
+    target_to_block TEXT NOT NULL,    -- Desired to block
+    synced_from_block BIGINT NOT NULL,  -- Current synced from block
+    synced_to_block BIGINT NOT NULL,    -- Current synced to block
+    syncers_id TEXT NOT NULL,          -- Syncer identifier
+    PRIMARY KEY (contract_address)
+);
+
+CREATE TABLE reorg_chain (
```
vcastellm: Better raname to `reorgs`, current name has not good semantics IMO

vcastellm (commented):
See comments

etherman/block_notifier/block_notifier_manager.go
joanestebanr: There are going to be several `blockFinality` when all the syncer uses this *multidownloader*. 
Each `blockNotifier` implies RPC calls, to reduce that the idea behind as follow next example: 
- finality: `Finalized/+10` and `Finalized/+20` shares 1  blockNotifier  `Finalized/0` and then adapt the offset intead of having 2 `blockNofitier` that ask form the same block to RPC.

WDYT?

goran-ethernal: Yes. Makes sense 👍🏻

multidownloader/storage/migrations/0001.sql
joanestebanr: Finally I decide to delete this table, because is not in use in current implementation and make more sense to added in next interation when reorg are going to be implemented

multidownloader/evm_multidownloader.go
```diff
@@ -0,0 +1,484 @@
...
+			err)
+	}
+	committed = true
+	if err = tx.Commit(); err != nil {
+		return fmt.Errorf("Safe/Step: cannot commit tx: %w", err)
+	}
+	return nil
+}
+
+func ethGetExtendendError(err error) string {
```
vcastellm:
```suggestion
func ethGetExtendedError(err error) string {
```

multidownloader/evm_multidownloader.go
```diff
@@ -0,0 +1,484 @@
...
+func (dh *EVMMultidownloader) getFinalizedBlockNumber(ctx context.Context) (uint64, error) {
+	bn, err := dh.blockNotifierManager.GetCurrentBlockNumber(ctx, dh.cfg.BlockFinality)
+	if err != nil {
+		return 0, fmt.Errorf("Safe/Step: cannot get finalized block (%s): %w",
+			dh.cfg.BlockFinality.String(), err)
+	}
+	return bn, nil
+}
+
+func (dh *EVMMultidownloader) getNextQuery(ctx context.Context, chunck uint32, safe bool) (*mdrtypes.LogQuery, error) {
```
vcastellm:
```suggestion
func (dh *EVMMultidownloader) getNextQuery(ctx context.Context, chunk uint32, safe bool) (*mdrtypes.LogQuery, error) {
```

multidownloader/evm_multidownloader.go
```diff
@@ -0,0 +1,484 @@
...
+	blockNotifierManager ethermantypes.BlockNotifierManager
+	name                 string
+	syncersConfig        mdrtypes.SetSyncerConfig
+
+	mutex         sync.Mutex
+	isInitialized bool
+	// These are the  segments that we need to sync
+	pendingSync *mdrtypes.SetSyncSegment
+	// These are the segments that we have already synced
+	// when a syncer do a `FilterLogs`is used to check what is already synced
```
vcastellm:
```suggestion
	// when a syncer does a `FilterLogs`, it is used to check what is already synced
```

multidownloader/evm_multidownloader.go
```diff
@@ -0,0 +1,484 @@
...
+			Service: NewEVMMultidownloaderRPC(logger, dh),
+		},
+	}
+}
+func (dh *EVMMultidownloader) CheckDatabase(ctx context.Context) error {
+	chainID, err := dh.ChainID(ctx)
+	if err != nil {
+		return fmt.Errorf("Initialize: cannot get chainID: %w", err)
+	}
+	compatibilityStoragedChecker := compatibility.NewCompatibilityCheck(
```
vcastellm:
```suggestion
	compatibilityStoragChecker := compatibility.NewCompatibilityCheck(
```

multidownloader/evm_multidownloader.go
```diff
@@ -0,0 +1,484 @@
...
+		true,
+		func(ctx context.Context) (storage.DBRuntimeData, error) {
+			return storage.DBRuntimeData{NetworkID: chainID,
+				DataVersion: storage.DataVersionCurrent}, nil
+		},
+		compatibility.NewKeyValueToCompatibilityStorage[storage.DBRuntimeData](dh.storage, "multidownloader-"+dh.name),
+	)
+
+	err = compatibilityStoragedChecker.Check(ctx, nil)
```
vcastellm:
```suggestion
	err = compatibilityStoragChecker.Check(ctx, nil)
```

multidownloader/evm_multidownloader_syncers.go
```diff
@@ -0,0 +1,103 @@
+package multidownloader
+
+// This is the implementations of types.MultiDownloader used by syncers
```
vcastellm:
```suggestion
// This is the implementation of types.MultiDownloader used by syncers
```

multidownloader/evm_multidownloader_syncers.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	header, err := dh.ethClient.HeaderByNumber(ctx, big.NewInt(int64(number)))
+	if err != nil {
+		return nil, fmt.Errorf("EVMMultidownloader.BlockHeader: cannot get header for block number=%d: %w",
+			number, err)
+	}
+	return aggkittypes.NewBlockHeaderFromEthHeader(header), nil
+}
+
+// FilterLogs filters the logs. It get it from storage or waits until they are available
```
vcastellm:
```suggestion
// FilterLogs filters the logs. It gets it from storage or waits until they are available
```

multidownloader/evm_multidownloader_syncers.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	if block != nil {
+		return block, nil
+	}
+	// This is a fallback mechanism in case the block is not found in storage (must be on storage!)
+	dh.log.Debugf("EVMMultidownloader.HeaderByNumber: block number=%d not found in storage, fetching from ethClient",
+		number.Uint64())
+	ethBlock, err := dh.ethClient.HeaderByNumber(ctx, number) // Just to comply with the interface
+	if err != nil {
+		return nil, fmt.Errorf("EVMMultidownloader.HeaderByNumber: fails ethClient.HeaderByNumber(%d). Err:  %w",
```
vcastellm:
```suggestion
		return nil, fmt.Errorf("EVMMultidownloader.HeaderByNumber: fails ethClient.HeaderByNumber(%d). Err: %w",
```

multidownloader/evm_multidownloader_syncers_test.go
```diff
@@ -0,0 +1,224 @@
...
+
+func TestEVMMultidownloader_BlockNumber(t *testing.T) {
+	testData := newEVMMultidownloaderTestData(t, true)
+	testData.mockBlockNotifierManager.EXPECT().GetCurrentBlockNumber(mock.Anything, aggkittypes.LatestBlock).
+		Return(uint64(123456), nil)
+	num, err := testData.mdr.BlockNumber(t.Context(), aggkittypes.LatestBlock)
+	require.NoError(t, err)
+	require.Equal(t, uint64(123456), num)
+}
+func TestEVMMultidownloader_BlockHeader(t *testing.T) {
```
vcastellm, line 62:
```suggestion

func TestEVMMultidownloader_BlockHeader(t *testing.T) {
```

multidownloader/evm_multidownloader_test.go
```diff
@@ -0,0 +1,464 @@
...
+		mdr, err := NewEVMMultidownloader(logger, cfg, customName, ethClient, nil, db, nil)
+		require.NoError(t, err)
+
+		services := mdr.GetRPCServices()
+
+		require.Len(t, services, 1)
+		require.Equal(t, "multidownloader-"+customName, services[0].Name)
+	})
+}
+func TestEVMMultidownloader_Initialize(t *testing.T) {
```
vcastellm, line 318:
```suggestion

func TestEVMMultidownloader_Initialize(t *testing.T) {
```

multidownloader/evm_multidownloader_test.go
```diff
@@ -0,0 +1,464 @@
...
+		err := testData.mdr.Initialize(t.Context())
+		require.NoError(t, err)
+
+		// Second initialization should fail
+		err = testData.mdr.Initialize(t.Context())
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "already initialized")
+	})
+}
+func TestEVMMultidownloader_StepSafe(t *testing.T) {
```
vcastellm, line 346:
```suggestion
}

func TestEVMMultidownloader_StepSafe(t *testing.T) {
```

multidownloader/config.go
```diff
@@ -0,0 +1,84 @@
...
+type Config struct {
+	// Enabled indicates if the multidownloader is enabled
+	Enabled bool
+	// StoragePath is the path to the storage
+	StoragePath string
+	// BlockChunkSize is the number of blocks to query in each FilterLogs call
+	BlockChunkSize uint32
+	// MaxParallelBlockHeaderRetrieval is the maximum number of parallel RPC calls to retrieve block headers
+	MaxParallelBlockHeaderRetrieval int
+	// BlockFinality is which block consider final (typically finalizedBlock)
```
vcastellm:
```suggestion
	// BlockFinality is which block to consider final (typically finalizedBlock)
```

multidownloader/config.go
```diff
@@ -0,0 +1,84 @@
...
+	BlockFinality aggkittypes.BlockNumberFinality
+	// WaitPeriodToCheckCatchUp is the wait applied when a request is still not available to check again
```
vcastellm:
```suggestion
	// WaitPeriodToCheckCatchUp is the duration to wait before checking again if logs are not yet available
```

multidownloader/config.go
```diff
@@ -0,0 +1,84 @@
...
+
+const (
+	defaultBlockChunkSize                  = 10000
+	defaultMaxParallelBlockHeaderRetrieval = 30
+	defaultWaitPeriodToCheckCatchUp        = time.Second * 10
+)
+
+func NewConfigDefault(name string, basePathDB string) Config {
+	if basePathDB == "" {
+		basePathDB = "/tmp/aggkit/"
```
vcastellm, line 47: This should use https://pkg.go.dev/os#TempDir

multidownloader/evm_multidownloader.go
joanestebanr: I suppose that you mean `compatibilityStorageChecker` isn't? missing an 'e'

multidownloader/types/syncer_config_test.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"testing"
+
+	aggkittypes "github.com/agglayer/aggkit/types"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/require"
+)
+
+// Use the real BlockNumberFinality interface/type from your codebase.
+// Replace this import path with the actual one if different.
```
vcastellm: ??

multidownloader/types/syncer_config_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Ensure your BlockNumberFinality implementation is available for use in tests.
```
vcastellm: ??

multidownloader/types/syncer_config_test.go
```diff
@@ -0,0 +1,108 @@
...
+		ToBlock:       aggkittypes.LatestBlock,
+	})
+
+	configs := set.ContractConfigs()
+	require.Len(t, configs, 1)
+	cc := configs[0]
+	require.Equal(t, addr, cc.Address)
+	// FromBlock should be the minimum
+	require.Equal(t, uint64(5), cc.FromBlock)
+	// ToBlock should be the "less final" (minimum)
```
vcastellm:
```suggestion
	// ToBlock should be the earliest/most permissive (LatestBlock < FinalizedBlock)
```

multidownloader/types/syncer_config.go
```diff
@@ -0,0 +1,189 @@
...
+	SyncerID SyncerID
+	// ContractAddr is list of contract addresses to sync
+	ContractsAddr []common.Address
+	// Starting block
+	FromBlock uint64
+	// Target for final block
+	ToBlock aggkittypes.BlockNumberFinality
+}
+
+func NewSyncerConfig(data aggkittypes.SyncerConfig) SyncerConfig {
```
vcastellm: Could multiple syncers be registered for the same contract address? should we validate that?

multidownloader/evm_multidownloader.go
joanestebanr: I assume the same name as before

multidownloader/config.go
joanestebanr: In fact is for match the default path in `default.go` file: https://github.com/agglayer/aggkit/blob/4ae734f0c3b334309905f28ca6041c9721257f9d/config/default.go#L41

multidownloader/storage/storage.go
```diff
@@ -0,0 +1,400 @@
...
+	a.mutex.Lock()
+	defer a.mutex.Unlock()
+	for _, segment := range segments {
+		result, err := tx.Exec(query, segment.BlockRange.FromBlock,
+			segment.BlockRange.ToBlock, segment.ContractAddr.Hex())
+		if err != nil {
+			return fmt.Errorf("error updating synced blocks for contract %s: %w",
+				segment.ContractAddr.Hex(), err)
+		}
+		if err != nil {
```
vcastellm, line 349: The second `if err != nil` will never execute because the function returns in the first check. This is dead code. Remove the first error check or restructure the logic.

multidownloader/types/syncer_config_test.go
joanestebanr: removed!

cmd/run.go
```diff
@@ -596,6 +618,38 @@ func runReorgDetectorL1IfNeeded(
 	return rd, errChan
 }
 
+func runL1MultiDownloaderIfNeeded(
+	_ context.Context,
```
goran-ethernal: Should we remove the first two parameters since they are not used?

multidownloader/types/syncer_config.go
```diff
@@ -0,0 +1,189 @@
...
+
+	aggkitcommon "github.com/agglayer/aggkit/common"
+	aggkittypes "github.com/agglayer/aggkit/types"
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+type SyncerID = string
+
+type SyncerConfig struct {
```
goran-ethernal: We have the same struct in `multidownloader.go`.

multidownloader/types/syncer_config.go
joanestebanr: Yes, there are already a test for that: https://github.com/agglayer/aggkit/blob/4ae734f0c3b334309905f28ca6041c9721257f9d/multidownloader/types/syncer_config_test.go#L72

multidownloader/types/syncer_config.go
```diff
@@ -0,0 +1,189 @@
...
+					addresses = append(addresses, addr)
+					dups[addr] = struct{}{}
+				}
+			}
+		}
+	}
+	return addresses
+}
+
+func (f *SetSyncerConfig) Finalities() []aggkittypes.BlockNumberFinality {
```
goran-ethernal: This doesn't seem to be used anywhere.

multidownloader/types/syncer_config.go
```diff
@@ -0,0 +1,189 @@
...
+func convertContractMapToSlice(contractMap map[common.Address]*ContractConfig) []ContractConfig {
+	contractConfigs := make([]ContractConfig, 0, len(contractMap))
+	for _, cc := range contractMap {
+		contractConfigs = append(contractConfigs, *cc)
+	}
+	return contractConfigs
+}
+
+// Combine all filters into one, must check if the blockRange overlaps
+func (f *SetSyncerConfig) Combine(blockRange aggkitcommon.BlockRange) ethereum.FilterQuery {
```
goran-ethernal: This is not used anywhere.

multidownloader/types/sync_segment.go
```diff
@@ -0,0 +1,64 @@
...
+		RequiredBlockHeader: requiredBlockHeader,
+	}
+}
+
+func (s *SyncSegment) String() string {
+	return "SyncSegment{ contracts:" + s.ContractAddr.Hex() + " range:" + s.BlockRange.String() +
+		" blockHeader:" + fmt.Sprintf("%v", s.RequiredBlockHeader) + "}"
+}
+
+func (s SyncSegment) NewBlockRange(br aggkitcommon.BlockRange) SyncSegment {
```
goran-ethernal, line 38: This is not used anywhere.

multidownloader/types/set_sync_segment.go
```diff
@@ -0,0 +1,289 @@
...
+	maxFromBlock := uint64(0)
+	for _, segment := range f.segments {
+		if segment.BlockRange.FromBlock < minToBLock {
+			minToBLock = segment.BlockRange.FromBlock
+		}
+		if segment.BlockRange.ToBlock > maxFromBlock {
+			maxFromBlock = segment.BlockRange.ToBlock
+		}
+	}
+	bn := aggkitcommon.NewBlockRange(minToBLock, maxFromBlock)
```
goran-ethernal: Hmm, looking at this function it is not doing what its description is saying. It says it returns the total number of pending blocks to synchronize, but in fact, it returns a new block range spanning from the segment with lowest `FromBlock` to the highest `ToBlock`.

So for example if we have segments:
- `Contract A : 1-10 (10 blocks)`
- `Contract B: 100-110 (10 blocks)`
`TotalBlocks()` will return `1-110 (110)` blocks, even though it is syncing 20 blocks actually.

Maybe it is supposed to return the entire block range `1-110`, but the name and descriptions are misleading.

multidownloader/types/set_sync_segment.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+// GetByContract returns the SyncSegment for the given contract address
+func (s *SetSyncSegment) GetByContract(addr common.Address) *SyncSegment {
+	if s == nil {
+		return nil
+	}
+	for _, segment := range s.segments {
+		if segment.ContractAddr == addr {
+			return segment.Clone()
```
goran-ethernal: Correct me if I am wrong, but as far as I see through out the code, we do not change the segment once we pass it. We might not need this `Clone`.

multidownloader/types/set_sync_segment.go
```diff
@@ -0,0 +1,289 @@
...
+	f.segments = newSegments
+}
+
+func (f *SetSyncSegment) UpdateBlockRange(segment *SyncSegment, newBlockRange aggkitcommon.BlockRange) {
+	if f == nil || segment == nil {
+		return
+	}
+	for i, s := range f.segments {
+		if s.Equal(*segment) {
+			f.segments[i].BlockRange = newBlockRange
```
goran-ethernal:
```suggestion
			s.BlockRange = newBlockRange
```

multidownloader/types/set_sync_segment.go
```diff
@@ -0,0 +1,289 @@
...
+
+type SetSyncSegment struct {
+	segments []*SyncSegment
+}
+
+// String returns a string representation of the SetSyncSegment
+func (s *SetSyncSegment) String() string {
+	result := "SetSyncSegment: "
+	for i, segment := range s.segments {
+		result += fmt.Sprintf("SyncSegment[%d]=%s\n", i, segment.BlockRange.String())
```
goran-ethernal: Suggestion: use `strings.Builder` perhaps if we have a number of segments it is a bit more performant.

multidownloader/storage/storage.go
```diff
@@ -0,0 +1,396 @@
...
+		TxHash:      log.TxHash,
+		TxIndex:     log.TxIndex,
+		Index:       log.Index,
+	}
+}
+
+const SqliteBoolTrue = 1
+const SqliteBoolFalse = 0
+
+type BlockRow struct {
```
goran-ethernal: This doesn't look like it have to be exported.
```suggestion
type blockRow struct {
```

multidownloader/storage/storage.go
```diff
@@ -0,0 +1,396 @@
...
+	DBPath string
+}
+
+type MultidownloaderStorage struct {
+	dbtypes.KeyValueStorager
+	logger aggkitcommon.Logger
+	db     *sql.DB
+	cfg    MultidownloaderStorageConfig
+
+	mutex sync.RWMutex
```
goran-ethernal, line 32: You do not need a mutex for operations on the `db`, it already has built in concurrency handling.

multidownloader/storage/migrations/0001.sql
```diff
@@ -0,0 +1,34 @@
...
+    tx_index INTEGER NOT NULL,
+    log_index INTEGER NOT NULL,      -- “index” is a reserved keyword
+    PRIMARY KEY (address, block_number, log_index)
+);
+
+CREATE TABLE blocks (
+    block_number BIGINT NOT NULL,
+    block_hash TEXT NOT NULL,             
+    block_timestamp INTEGER NOT NULL,
+    block_parent_hash TEXT NOT NULL,
```
goran-ethernal: The `blockRaw` struct is defining the `BlockParentHash` as a nullable field:

```go
// BlockParentHash can be nil (the ethLogs doesn't include it)
BlockParentHash *common.Hash `meddler:"block_parent_hash,hash"`
```

But this is a non nullable in the `db` schema.

multidownloader/storage/migrations/0001.sql
```diff
@@ -0,0 +1,34 @@
+-- +migrate Down
+DROP TABLE IF EXISTS logs;
+DROP TABLE IF EXISTS blocks;
+DROP TABLE IF EXISTS sync_status;
+-- +migrate Up
+CREATE TABLE logs (
```
goran-ethernal, line 6: Should we maybe add an index for the logs table? It will contain a lot of rows, so it might help querying the block ranges. WDYT?

multidownloader/storage/storage.go
```diff
@@ -0,0 +1,396 @@
...
+	logs := make([]types.Log, 0)
+	dbRows := make([]*logAndBlockRow, 0)
+	sqlQuery := `
+	SELECT * FROM logs
+	LEFT JOIN blocks ON logs.block_number = blocks.block_number
+	WHERE address IN (?)
+	AND logs.block_number>=? AND logs.block_number<=?
+	ORDER BY logs.block_number ASC, log_index ASC
+	`
+	addrs := []string{}
```
goran-ethernal:
```suggestion
	addrs := make([]string, 0, len(query.Addrs))
```

multidownloader/storage/storage.go
```diff
@@ -0,0 +1,396 @@
...
+	defer a.mutex.Unlock()
+	// Save blocks headers
+	if err := a.saveBlocksNoMutex(tx, blockRows); err != nil {
+		return fmt.Errorf("saveLogsAndBlocks: error saving blocks: %w", err)
+	}
+
+	if err := a.saveLogsNoMutex(tx, logRows); err != nil {
+		return fmt.Errorf("saveLogsAndBlocks: error saving logs: %w", err)
+	}
+	// TODO: Sanity check logs match blockHash match with headers
```
goran-ethernal, line 269: Should we resolve this?

multidownloader/storage/storage_block.go
```diff
@@ -0,0 +1,142 @@
...
+		AreFinal: make(map[uint64]bool),
+	}
+}
+
+func (b *Blocks) Add(header *aggkittypes.BlockHeader, isFinal bool) {
+	b.Headers[header.Number] = header
+	b.AreFinal[header.Number] = isFinal
+}
+
+func (b *Blocks) Get(number uint64) (*aggkittypes.BlockHeader, bool, error) {
```
goran-ethernal, line 31: This is a confusing API in my opinion, having a bool and an error, especially, since I see that you are not using this `bool` (`isFinal`) return value anywhere. Maybe just remove the bool return parameter?

multidownloader/storage/storage_block.go
```diff
@@ -0,0 +1,142 @@
...
+		return nil, false, nil
+	}
+	header, isFinal, err := blocks.Get(blockNumber)
+	if err != nil {
+		return nil, false, err
+	}
+	return header, isFinal, nil
+}
+
+func (a *MultidownloaderStorage) GetBlockHeaderNotFinal(tx dbtypes.Querier,
```
goran-ethernal: This is not used anywhere.

multidownloader/storage/storage_block.go
```diff
@@ -0,0 +1,142 @@
...
+	header *aggkittypes.BlockHeader, isFinal bool) error {
+	blockRows := map[uint64]*BlockRow{
+		header.Number: newBlockRowFromAggkitBlock(header, isFinal),
+	}
+	a.mutex.Lock()
+	defer a.mutex.Unlock()
+	return a.saveBlocksNoMutex(tx, blockRows)
+}
+
+func (a *MultidownloaderStorage) UpdateIsFinal(tx dbtypes.Querier, blockNumbers []uint64) error {
```
goran-ethernal: This is only used in tests, so I suggest making this a private function, or just update the `IsFinal` flag directly in the test.

multidownloader/storage/storage_block.go
```diff
@@ -0,0 +1,142 @@
...
+		headers = append(headers, header)
+	}
+	return headers
+}
+
+func (b *Blocks) IsEmpty() bool {
+	return len(b.Headers) == 0
+}
+
+func (a *MultidownloaderStorage) SaveBlockAggkitBlock(tx dbtypes.Querier,
```
goran-ethernal:
```suggestion
func (a *MultidownloaderStorage) SaveAggkitBlock(tx dbtypes.Querier,
```

Also, this is only used in tests, so I suggest making this a private function, or just update the `db` directly in the test.

multidownloader/statistics.go
```diff
@@ -0,0 +1,87 @@
...
+	return s.timeTrackerTotal.Elapsed()
+}
+
+func (s *Statistics) StartDBOperation() {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	s.timeTrackerDB.Start()
+}
+
+func (s *Statistics) FinishDBOperation(_ error) {
```
goran-ethernal: Remove the error parameter since it is not used?

multidownloader/evm_multidownloader_syncers.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	return aggkittypes.NewBlockHeaderFromEthHeader(header), nil
+}
+
+// FilterLogs filters the logs. It gets it from storage or waits until they are available
+func (dh *EVMMultidownloader) FilterLogs(ctx context.Context, query ethereum.FilterQuery) ([]types.Log, error) {
+	dh.log.Debugf("EVMMultidownloader.FilterLogs: received query: %+v", query)
+	defer dh.log.Debugf("EVMMultidownloader.FilterLogs: finished query: %+v", query)
+	logQuery := mdrtypes.NewLogQueryFromEthereumFilter(query)
+	for !dh.IsAvailable(logQuery) {
```
goran-ethernal, line 52: This can wait forever if its not available, right? Maybe we should add some protection against this.

multidownloader/evm_multidownloader_syncers.go
```diff
@@ -0,0 +1,103 @@
...
+
+	block, _, err := dh.storage.GetBlockHeaderByNumber(nil, number.Uint64())
+	if err != nil {
+		return nil, fmt.Errorf("EVMMultidownloader.HeaderByNumber: cannot get BlockHeader number=%d: %w",
+			number.Uint64(), err)
+	}
+	if block != nil {
+		return block, nil
+	}
+	// This is a fallback mechanism in case the block is not found in storage (must be on storage!)
```
goran-ethernal: If it must be on storage, should we just return an error in this case, or at least, store this fallback result to `db`?

multidownloader/types/syncer_config_test.go
joanestebanr: removed

cmd/run.go
joanestebanr: 02ca5d5b24e71973c7732ad9ba8029ff5b04b5af

multidownloader/types/syncer_config.go
joanestebanr: good catch!

multidownloader/types/syncer_config.go
joanestebanr: Good catch! af6db2ffbc499d87025e21d968b0f9f49d3c167c

multidownloader/types/syncer_config.go
joanestebanr: 391df26d417d68de0bf1e728398a63a8cb59b9b5

multidownloader/types/sync_segment.go
joanestebanr: removed

multidownloader/types/set_sync_segment.go
joanestebanr: The idea is don't allow to modify the internal segment, so the interface clone the segment before returning it. So the caller can do whatever he wants with the pointer. 
Maybe make more sense to modify the signature and return a `SyncSegment` instead a `*SyncSegment`, WDYT?

multidownloader/storage/storage.go
joanestebanr: d49e981de0e722508d4972ae41daf2e029231f26

multidownloader/evm_multidownloader_syncers.go
joanestebanr: No, it can be that the required block in not in Multidownloader: currently the syncer as asking for empty blocks as synchronization points, these blocks are not synced by Multidownloader. That's the reason that is not an error,  nor even a warning

multidownloader/evm_multidownloader_syncers.go
joanestebanr: I agree, but it's not straight-forward. Because maybe you are asking for latest block and Multidownloader is syncing the first one, so it could take hours or days.

 The user can set a timeout in the context to prevent that 'forever' wait

multidownloader/statistics.go
joanestebanr: 9ccf2f4273c37e9d2467135199e6d47fed691af7

multidownloader/storage/storage_block.go
joanestebanr: It was used by the pre-implementation  of sync tip of chain, I would remove and reintroduce in next PR

multidownloader/storage/storage_block.go
joanestebanr: It's require, because, even now all blocks are `isFinal` = true. The current function can save this flag, so it's required to be retrieved also (is used in tests right now)

multidownloader/storage/storage.go
joanestebanr: not yet, I prefer to merge the changes and refine in the future with the full implementation

multidownloader/storage/migrations/0001.sql
joanestebanr: The  primary key is contains `address` and `block_number` and this generate automatically a index.
But there are also select only over `block_number`. So I added a index for block_number

multidownloader/storage/migrations/0001.sql
joanestebanr: You're right modified db schema

multidownloader/storage/storage.go
joanestebanr: In fact if you plan to write from several go-routines you need a mutex ([here](https://github.com/mattn/go-sqlite3/issues/1179)). Also, without writting I got very often `database table is locked` from the syncers go-rotuines. 

Right now, I think that only 1 go-routine is writting data, but in a long run the idea is that missing data that is required by a syncer to be stored in DB just in case that the syncer delete his database and need again the same data

multidownloader/types/set_sync_segment.go
joanestebanr: for sure

joanestebanr: 1f5eae911a4975f1b7232bbc590345978ea07aba

multidownloader/types/set_sync_segment.go
goran-ethernal: Sure. That is ok as well.

multidownloader/storage/storage.go
goran-ethernal: Take a look at this comment: https://github.com/mattn/go-sqlite3/issues/1179#issuecomment-1638083995.

It says the library is already safe for concurrent use. It would be really strange that the sqlite does not have this built in. Even the simple key-value db's like `leveldb` or `boltdb` have this built in.

multidownloader/types/set_sync_segment.go
joanestebanr: You're right 0b6061f59a4b4ea9b7587cb9e492612b09b070c3
Now in your example is 20 and next case: 
- A: 1-10
- B: 5-20
- C: 100-110

Total = 20 (A+B) and 10 (C)

multidownloader/types/set_sync_segment.go
joanestebanr: 54a85e3f62e24cba802e80518c64d729789328d6

goran-ethernal (approved):

vcastellm (approved):
LGTM

PR #1253, opened 2025-11-05
chore: reuse `common.HashLength` instead of magic number

Stefan-Ethernal:
## 🔄 Changes Summary
- Reuse the `common.HashLength` constant provided by `go-ethereum`, instead of using magic number 32 when dealing with raw hashes. This improves readability of code.
- Also remove unused `NewL2ReadOnly` constructor function.

## ⚠️ Breaking Changes
N/A

## 📋 Config Updates
N/A

## ✅ Testing

## 🐞 Issues
- Closes #[issue-number]


bridgesync/bridgesync.go
```diff
@@ -83,24 +82,6 @@ func NewL1(
 	)
 }
 
-func NewL2ReadOnly(
```
temaniarpit27, line 86: isnt it being used anywhere?

temaniarpit27 (approved):

goran-ethernal (approved):

bridgesync/bridgesync.go
Stefan-Ethernal: Nope, that's why I have removed it in the first place :slightly_smiling_face:

temaniarpit27: wasnt our plan to build something like this just for reading the l2bridgesyncer?

Stefan-Ethernal: I'm unaware of it. We can always bring it back in the future if necessary. WDYT?

temaniarpit27: +1
like was just wondering when and why did we add this

PR #1262, opened 2025-11-05
feat: check always last cert using the initial check 

### LINKED ISSUE #1215: bug: aggsender: in check status take in consideration that the state in agglayer and aggsender are not the same
# Bug Report
- A certificate have been forced from `InError` to `Settled` in **Agglayer**
- **Aggsender** must check this situation, and update the local state before trying to send a new certificate

Affected versions: `0.7.x`

joanestebanr:
## 🔄 Changes Summary
**Aggsender** test execute the same align code that at startup periodically to be aligned with **Agglayer**
This will allow to support the case that the last certificate `InError` is promoted manually to `Settled`

## 🐞 Issues
- Closes #1215

### DIFF
--- a/aggsender/aggsender.go
+++ b/aggsender/aggsender.go
@@ -269,7 +269,7 @@ func (a *AggSender) sendCertificates(ctx context.Context, returnAfterNIterations
 				a.status.SetLastError(err)
 				a.log.Errorf("error checking last certificate from agglayer: %v", err)
 			}
-			if !checkResult.ExistPendingCerts && checkResult.ExistNewInErrorCert {
+			if err == nil && !checkResult.ExistPendingCerts && checkResult.ExistNewInErrorCert {
 				if a.cfg.RetryCertAfterInError {
 					a.log.Infof("An InError cert exists. Sending a new one (%s)", a.cfg.CheckCertConfigBriefString())
 					_, err := a.sendCertificate(ctx)
@@ -279,7 +279,7 @@ func (a *AggSender) sendCertificates(ctx context.Context, returnAfterNIterations
 					}
 					a.checkSendCertificateStopCondition(err)
 				} else {
-					a.log.Infof("An InError cert exists but skipping send cert because RetryCertAfterInError is false")
+					a.log.Infof("An InError cert exists or fails check (%w) but skipping send cert because RetryCertAfterInError is false", err)
 				}
 			}
 			if returnAfterNIterations > 0 && iteration >= returnAfterNIterations {
--- a/aggsender/aggsender.go
+++ b/aggsender/aggsender.go
@@ -301,7 +301,7 @@ func (a *AggSender) sendCertificates(ctx context.Context, returnAfterNIterations
 						a.status.SetLastError(err)
 					}
 				} else {
-					a.log.Infof("epoch trigger: Skipping epoch %s because there are pending certificates",
+					a.log.Infof("Epoch trigger: Skipping epoch %s because there are pending certificates",
 						epoch.String())
 				}
 			}
--- a/aggsender/aggsender.go
+++ b/aggsender/aggsender.go
@@ -268,21 +268,20 @@ func (a *AggSender) sendCertificates(ctx context.Context, returnAfterNIterations
 			if err != nil {
 				a.status.SetLastError(err)
 				a.log.Errorf("error checking last certificate from agglayer: %v", err)
-			} else {
-				if err == nil && !checkResult.ExistPendingCerts && checkResult.ExistNewInErrorCert {
-					if a.cfg.RetryCertAfterInError {
-						a.log.Infof("An InError cert exists. Sending a new one (%s)", a.cfg.CheckCertConfigBriefString())
-						_, err := a.sendCertificate(ctx)
-						a.status.SetLastError(err)
-						if err != nil {
-							a.log.Error(err)
-						}
-						a.checkSendCertificateStopCondition(err)
-					} else {
-						a.log.Infof("An InError cert exists but skipping send cert because RetryCertAfterInError is false")
+			} else if err == nil && !checkResult.ExistPendingCerts && checkResult.ExistNewInErrorCert {
+				if a.cfg.RetryCertAfterInError {
+					a.log.Infof("An InError cert exists. Sending a new one (%s)", a.cfg.CheckCertConfigBriefString())
+					_, err := a.sendCertificate(ctx)
+					a.status.SetLastError(err)
+					if err != nil {
+						a.log.Error(err)
 					}
+					a.checkSendCertificateStopCondition(err)
+				} else {
+					a.log.Infof("An InError cert exists but skipping send cert because RetryCertAfterInError is false")
 				}
 			}
+
 			if returnAfterNIterations > 0 && iteration >= returnAfterNIterations {
 				a.log.Warnf("reached number of iterations, so we are going to return")
 				return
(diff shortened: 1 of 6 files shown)


aggsender/aggsender.go
```diff
@@ -263,7 +263,12 @@ func (a *AggSender) sendCertificates(ctx context.Context, returnAfterNIterations
 			iteration++
 			a.log.Debugf("Checking perodical certificates status (%s)",
 				a.cfg.CheckCertConfigBriefString())
-			checkResult := a.certStatusChecker.CheckPendingCertificatesStatus(ctx)
+
+			checkResult, err := a.certStatusChecker.CheckPeriodicallyCertificateStatus(ctx)
+			if err != nil {
+				a.status.SetLastError(err)
```
goran-ethernal, line 269: Should we continue the loop in case of an error?

aggsender/aggsender.go
```diff
@@ -284,20 +289,25 @@ func (a *AggSender) sendCertificates(ctx context.Context, returnAfterNIterations
 		case epoch := <-chEpoch:
 			iteration++
 			a.log.Infof("Epoch received: %s", epoch.String())
-			checkResult := a.certStatusChecker.CheckPendingCertificatesStatus(ctx)
+			checkResult, err := a.certStatusChecker.CheckPeriodicallyCertificateStatus(ctx)
+			if err != nil {
+				a.log.Errorf("Epoch trigger: error checking certificate status: %v", err)
+				a.status.SetLastError(err)
+				break
```
goran-ethernal: Same here, should we continue the loop, and not break it?

aggsender/statuschecker/cert_status_checker.go
```diff
@@ -100,11 +100,24 @@ func (c *certStatusChecker) CheckInitialStatus(
...
-// CheckPendingCertificatesStatus checks the status of pending certificates
+// CheckPeriodicallyCertificateStatus checks the status of pending certificates
+// and the last certificate from the aggregation layer.
+// It returns the status of pending certificates and any error encountered
+// while checking the last certificate from the aggregation layer.
+func (c *certStatusChecker) CheckPeriodicallyCertificateStatus(
+	ctx context.Context,
+) (types.CertStatus, error) {
+	// Don't need to check the returned value here
+	c.checkPendingCertificatesStatus(ctx)
```
goran-ethernal: Why do we need to call this function twice?

aggsender/aggsender.go
joanestebanr: I unify behaviour, in both cases doesn't break but avoid sending a certifiacte

aggsender/statuschecker/cert_status_checker.go
joanestebanr: 🤔 I think that you're right, the first call it's redudant, because the  `checkLastCertificateFromAgglayer` is going to update the state if required, good catch!

joanestebanr: 33ec7152a4263d3d94e78ee4ebe0ae65840237c3

aggsender/aggsender.go
joanestebanr: I unify behaviour, in both cases doesn't break but avoid sending a certifiacte

aggsender/aggsender.go
goran-ethernal: No, just put `continue` instead of the `break`, Maybe `agglayer` connection failed, so we need to try again. WDYT?

aggsender/aggsender.go
```diff
@@ -263,8 +263,12 @@ func (a *AggSender) sendCertificates(ctx context.Context, returnAfterNIterations
...
 			a.log.Debugf("Checking perodical certificates status (%s)",
 				a.cfg.CheckCertConfigBriefString())
-			checkResult := a.certStatusChecker.CheckPendingCertificatesStatus(ctx)
-			if !checkResult.ExistPendingCerts && checkResult.ExistNewInErrorCert {
+
+			checkResult, err := a.certStatusChecker.CheckPeriodicallyCertificateStatus(ctx)
+			if err != nil {
+				a.status.SetLastError(err)
+				a.log.Errorf("error checking last certificate from agglayer: %v", err)
+			} else if err == nil && !checkResult.ExistPendingCerts && checkResult.ExistNewInErrorCert {
```
goran-ethernal: You don't need to add the check `if err == nil`, it is already nil if it gets to this `else if` since the condition in the `if` is `err != nil`.

aggsender/statuschecker/cert_status_checker.go
goran-ethernal: It seems that you didn't remove the first call to `checkPendingCertificatesStatus` 😕

joanestebanr: Now it's done 36e896bc31e62d1e3f307303ea897475b3fcc392

aggsender/aggsender.go
joanestebanr: done

aggsender/aggsender.go
```diff
@@ -277,27 +281,33 @@ func (a *AggSender) sendCertificates(ctx context.Context, returnAfterNIterations
...
-				a.log.Infof("Skipping epoch %s because there are pending certificates",
-					epoch.String())
+				if !checkResult.ExistPendingCerts {
+					_, err := a.sendCertificateWithRetries(ctx)
+					if err != nil {
+						a.log.Errorf("Epoch trigger: error sending certificate: %v", err)
+						a.status.SetLastError(err)
+					}
+				} else {
+					a.log.Infof("epoch trigger: Skipping epoch %s because there are pending certificates",
```
Stefan-Ethernal: Nitpick, make the letters case consistent in the logs (either upper or lower case everywhere):
- https://github.com/agglayer/aggkit/blob/36e896bc31e62d1e3f307303ea897475b3fcc392/aggsender/aggsender.go#L294
- https://github.com/agglayer/aggkit/blob/36e896bc31e62d1e3f307303ea897475b3fcc392/aggsender/aggsender.go#L300

```suggestion
					a.log.Infof("Epoch trigger: Skipping epoch %s because there are pending certificates",
```

aggsender/aggsender.go
```diff
@@ -277,27 +281,33 @@ func (a *AggSender) sendCertificates(ctx context.Context, returnAfterNIterations
...
+						epoch.String())
+				}
 			}
 
 			if returnAfterNIterations > 0 && iteration >= returnAfterNIterations {
-				a.log.Warnf("reached number of iterations, so we are going to return")
+				a.log.Warnf("epoch trigger: reached number of iterations, so we are going to return")
```
Stefan-Ethernal:
```suggestion
				a.log.Warnf("Epoch trigger: reached number of iterations, so we are going to return")
```

aggsender/types/interfaces.go
```diff
@@ -187,7 +187,7 @@ type Logger interface {
 
 // CertificateStatusChecker is an interface defining functions that a CertificateStatusChecker should implement
 type CertificateStatusChecker interface {
-	CheckPendingCertificatesStatus(ctx context.Context) CertStatus
+	CheckPeriodicallyCertificateStatus(ctx context.Context) (CertStatus, error)
```
Stefan-Ethernal: I would simplify the naming and call it simply like this (as the function isn't aware whether it is called periodically or not, it simply checks the certificate status). WDYT?
```suggestion
	CheckCertificateStatus(ctx context.Context) (CertStatus, error)
```

Stefan-Ethernal (approved):

goran-ethernal (approved):

aggsender/types/interfaces.go
joanestebanr: There are another function called `CheckInitialStatus` so it make sense that this one point that is regularly called instead of called at startup... maybe it can be rename to `CheckPeriodicallyStatus` to be more symmetrical if you think that fit better. WDYT?

Stefan-Ethernal: I'm fine with both approaches, maybe `CheckPeriodicallyStatus` is better aligned, but as you wish...

joanestebanr: done 4b5c0a8eb3f672bf7631d46f1dac58bbb9ae205a

PR #1298, opened 2025-11-13
fix: handle reorg detection edgecase

temaniarpit27:
## 🔄 Changes Summary
- Handle an edgecase when we are not able to track previous block reorgs and we are on the current block. When polling new block, the blockhash changes after the last ticker, then we are not able to track the block hash of previous block and new block is added. Now we are always comparing the latest block header in reorg detection as well

### DIFF
--- a/reorgdetector/reorgdetector.go
+++ b/reorgdetector/reorgdetector.go
@@ -183,8 +183,6 @@ func (rd *ReorgDetector) detectReorgInTrackedList(ctx context.Context) error {
 			headers := hdrs.getSorted()
 			for _, hdr := range headers {
 				// Get the actual header from the network or from the cache
-				var err error
-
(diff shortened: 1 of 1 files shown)


rachit77 (approved):

reorgdetector/reorgdetector.go
```diff
@@ -186,19 +184,22 @@ func (rd *ReorgDetector) detectReorgInTrackedList(ctx context.Context) error {
 			for _, hdr := range headers {
 				// Get the actual header from the network or from the cache
 				var err error
```
Stefan-Ethernal: It seems like we can remove this declaration, but double check please
```suggestion
```

Stefan-Ethernal (approved):

joanestebanr (approved):

vcastellm (approved):

PR #1296, opened 2025-11-12
fix: add support zkevm claims pre-etrog

joanestebanr:
## 🔄 Changes Summary
- Add support to  pre-etrog claims (that doesn't have a direct leaf in L1InfoTree). It require to manually activate the feature, please check **Config Updates** section.
- Fix estimatedSize that must consided metadata as a Hash, not the length of bytes

## 🔨 Implementation
- First from the list os claims find the first claim 'postEtrog'  executing calls to RPC Filter.Logs
- For each pre-etrog claim set a new style 
- Set a proof a future GER

## ⚠️ Breaking Changes
- 🛠️ **Config**: [Optional: Changes to TOML config]
- 🔌 **API/CLI**: [Optional: Breaking interface changes]
- 🗑️ **Deprecated Features**: [Optional: Removed features]

## 📋 Config Updates
- To activate the feature require add next section to your configfile
```
	[AggSender.SupportLegacyZKEVM]
		Enabled = true
		L2BridgeAddr = "{{polygonBridgeAddr}}"
		RPCFilterChunkSize = 10000
```
The value `RPCFilterChunkSize` must be: 
- For **mainnet**: 1000
- For **bali**: 10000

## ✅ Testing
- 🖱️ **Manual**: 
- We tested in a shadowfork of zkevm Mainnet

## 🐞 Issues
- Closes #1302

PR #1304, opened 2025-11-14
feat: pick fixes from zkevm_migration_support-0_7_X

joanestebanr:
## 🔄 Changes Summary
- Fix estimatedSize that must consided metadata as a Hash, not the length of bytes
- Add a new RPC end-point for aggsender `aggsender_triggerCertitificate` to force a trigger certificate:
This is a useful feature for testing or event can be used in production to don't wait a epoch to force a trigger
```
curl -X POST http://localhost:5576/ "Content-Type: application/json"   -d '{"method":"aggsender_triggerCertitificate", "params":[], "id":1}'
```

## 🔗 Related PRs
- #1296

### DIFF
--- a/aggsender/aggsender.go
+++ b/aggsender/aggsender.go
@@ -192,6 +192,10 @@ func (a *AggSender) Info() types.AggsenderInfo {
 	return res
 }
 
+func (a *AggSender) ForceTriggerCertitificate() {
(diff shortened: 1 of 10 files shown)


vcastellm (approved):
LGTM

PR #1307, opened 2025-11-17
feat: pick fixes from zkevm_migration_support-0_7_X (#1304)

joanestebanr:
- Fix estimatedSize that must consided metadata as a Hash, not the length of bytes
- Add a new RPC end-point for aggsender `aggsender_triggerCertitificate` to force a trigger certificate:
This is a useful feature for testing or event can be used in production to don't wait a epoch to force a trigger
```
curl -X POST http://localhost:5576/ "Content-Type: application/json"   -d '{"method":"aggsender_triggerCertitificate", "params":[], "id":1}'
```
- #1296

## 🔗 Related PRs
- #1304

## 📝 Notes

### DIFF
--- a/aggsender/certificate_send_trigger.go
+++ b/aggsender/certificate_send_trigger.go
@@ -3,13 +3,14 @@ package aggsender
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	"github.com/agglayer/aggkit/agglayer"
 	"github.com/agglayer/aggkit/aggsender/config"
 	"github.com/agglayer/aggkit/aggsender/types"
 	aggkitcommon "github.com/agglayer/aggkit/common"
 	"github.com/agglayer/aggkit/log"
-	"github.com/agglayer/aggkit/sync"
+	aggkitsync "github.com/agglayer/aggkit/sync"
 	aggkittypes "github.com/agglayer/aggkit/types"
 )
 
@@ -150,6 +151,7 @@ func (r *epochBasedTrigger) ForceTriggerEvent() {
 type preconfTrigger struct {
 	log          aggkitcommon.Logger
 	l2BridgeSync types.L2BridgeSyncer
+	mutex        sync.Mutex
 	ch           chan types.CertificateTriggerEvent
 }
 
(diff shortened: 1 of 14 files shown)


aggsender/types/certificate_build_params.go
```diff
@@ -172,17 +172,8 @@ func (c *CertificateBuildParams) EstimatedSize() uint {
...
-		sizeBridges += agglayertypes.EstimatedBridgeExitSize
-		sizeBridges += float64(len(bridge.Metadata))
-	}
-
-	sizeClaims := float64(0)
-	for _, claim := range c.Claims {
-		sizeClaims += agglayertypes.EstimatedImportedBridgeExitSize
-		sizeClaims += float64(len(claim.Metadata))
-	}
+	sizeBridges := (agglayertypes.EstimatedBridgeExitSize + common.HashLength) * float64(len(c.Bridges))
```
Stefan-Ethernal, line 176: Consider adding a comment that indicate that `HashLength` is added because of the `Metadata` field, WDYT?

aggsender/certificate_send_trigger.go
```diff
@@ -181,19 +190,40 @@ func (r *preconfTrigger) Setup(ctx context.Context) {
 // notifications. Each value is a sync.Block (which implements CertificateTriggerEvent).
 // The returned channel will be closed when the provided context is canceled.
 func (r *preconfTrigger) TriggerCh(ctx context.Context) <-chan types.CertificateTriggerEvent {
-	ch := make(chan types.CertificateTriggerEvent)
 	syncSub := r.l2BridgeSync.SubscribeToSync("aggsender")
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
```
Stefan-Ethernal: What are we trying to protect with this mutex? If it is the channel, it should be thread-safe by default (and also it is a bit strange setting it to nil in case of cancellation)...

We should only take care not to write into already closed channel in the `ForceTriggerEvent` (which if necessary we can track using some field of `atomic.Bool` value).

joanestebanr: I agree, but I added the mutex because the tests where complaining about "data race".. I think that even the channel is thread-safe is not the member of the object 'ch' that requires to be protected to access it.
I remove the unneeded mutex

Stefan-Ethernal: The pattern with using mutex together with channels is not necessary AFAIK.

I have tried removing the `mutex` field altogether and I cannot see any data race warnings in the output. 
```go
stefan@Stefan:~/go/src/Polygon/aggkit$ go test -race -count=20 github.com/agglayer/aggkit/aggsender
ok      github.com/agglayer/aggkit/aggsender    56.281s
```
Therefore I'm still suggesting to remove the `mutex` field. WDYT?

Stefan-Ethernal (approved):
Generally LGTM

aggsender/certificate_send_trigger.go
joanestebanr: You're right is no longer complaining about data race! 
How do you suggest to fix this code to guarantee that after `if r.ch==nil` is not modified the value until line `r.ch <-...`?
```
func (r *preconfTrigger) ForceTriggerEvent() {
	blockNumber, err := r.l2BridgeSync.GetLastProcessedBlock(context.Background())
	if err != nil {
		r.log.Errorf("ForceTriggerEvent: Failed to get last processed block: %v", err)
		return
	}
	r.mutex.Lock()
	defer r.mutex.Unlock()
	if r.ch == nil {
		return
	}
	r.ch <- aggkitsync.Block{Num: blockNumber}
}
```

Stefan-Ethernal: You can check if the channel is closed, before trying to write to it, which is if I'm not mistaken enough to protect yourself from writing into closed channel

```go
_, ok := <-ch
if !ok {
    // Channel is closed and all buffered values have been received.
    fmt.Println("Channel is closed!")
}
```

joanestebanr: But if it's closed beetween the check and sending data?
```
check_channel()
#  other gorotuine close channel
r.ch <- myData
```
I still think that we are not protected against that, but anyway, I have removed the mutex! cdab031ed6141cf6982cbd9120ac85375cd03a08

goran-ethernal (approved):

PR #1316, opened 2025-11-20
feat: insert corrected claims and track invalid ones

### LINKED ISSUE #1299: handle DetailedClaimEvent for Inject Latest-N ger changes
`DetailedClaimEvent` will also be indexed in cases when reorgs occur on L1 as we might need to correct the data in DB for claims which happened on L2
While indexing `DetailedClaimEvent` we need to take care of the following things (refer https://github.com/agglayer/ADRs/issues/28 for cases):
1. In case B.1, `DetailedClaimEvent` will be emitted with corrected data, in this case we need to update the already existing claim with new data
2. In case B.2, `DetailedClaimEvent` will be emitted with corrected data but we need to enter this in a new row. It can be identified that if we have a unsetclaim event for the global index in event, we need to add it as a new row

Stefan-Ethernal:
## 🔄 Changes Summary
Implement upsertions of claims (in case there is no unset of the claim and fixed claim was "replayed"), which is needed for non-finalized GERs injection.

It also keeps track of invalid claims, which are being kept in the invalid_claim table. There are two cases:
- GER is invalid, but claim is correct (in that case we don't need to unset the claim)
- both GER and claims are invalid (in that case we need to unset the claim, and we insert the invalid claim in case when we index the unset claim event from contract)

## ⚠️ Breaking Changes
N/A

## 📋 Config Updates
N/A

## ✅ Testing

## 🐞 Issues
- Closes #1299 

## 🔗 Related PRs

### DIFF
--- a/bridgesync/migrations/bridgesync0010.sql
+++ b/bridgesync/migrations/bridgesync0010.sql
@@ -1,6 +1,7 @@
 -- +migrate Down
 DROP TABLE IF EXISTS set_claim;
 DROP TABLE IF EXISTS unset_claim;
+DROP TABLE IF EXISTS invalid_claim;
 
 ALTER TABLE claim ADD COLUMN from_address VARCHAR;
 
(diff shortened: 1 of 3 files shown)


joanestebanr (approved):

goran-ethernal (approved):

bridgesync/migrations/bridgesync0010.sql
```diff
@@ -24,4 +25,27 @@ CREATE TABLE set_claim (
 	PRIMARY KEY (block_num, block_pos)
 );
 
+CREATE TABLE invalid_claim (
+    id                      INTEGER PRIMARY KEY AUTOINCREMENT,
```
temaniarpit27: do we need this auto increment
block_num, block_pos should be unique

temaniarpit27 (approved):

bridgesync/migrations/bridgesync0010.sql
Stefan-Ethernal: Fixed https://github.com/agglayer/aggkit/pull/1316/commits/369b43d5da9fee056aad2405cde9bdcdef96ef2f

PR #1336, opened 2025-11-27
feat: preserve all claims, provide a way to compact them

### LINKED ISSUE #1330: preserve the block num and block pos for corrected claims
Once we emit the `DetailedClaimEvent`, as a correction measure to fix the claim calldata https://github.com/agglayer/agglayer-contracts/pull/568  (namely the proofs and exit roots), we should preserve all the other data of the original claim.

The `claim` table should be append only (no updates and no removals, unless there is a reorg). Drop the invali_claim table, it is unnecessary anymore. When fetching the claims, provide a way to say whether we would like:
- compacted claims - combine the data from the original claim (the first one historically, sorted by `block_num` and `block_pos`) and only rewrite the `proof_local_exit_root`, `proof_rollup_exit_root`, `mainnet_exit_root` and `rollup_exit_root` from the most recent one,
- full history - return all the claims we have indexed (for given block range, global index etc.)

goran-ethernal:
## 🔄 Changes Summary
This PR introduces a new `compacted` parameter to the `GetClaims` function, allowing callers to choose between receiving all claims or a compacted view where duplicate claims (same `global_index`) are merged.

### Problem

Previously, when multiple claims existed for the same `global_index` (e.g., due to GER updates on L1), the bridge would only return the oldest claim. This caused issues when:
- Users needed to see the complete history of all claim attempts
- The newest proofs were required for successful claim execution
- Debugging required visibility into all claim events
- Another table was needed to keep track of invalid claims
- It becomes tricky to update the `db` in cases of reorgs.

### Solution

We keep the record of all claims in the `claims` table even if there are claims with same `global_index`. This will make debugging possible issues easier, plus updating is minimal when it comes to reorgs, just remove everything that happens from the reorged block, without the need to update multiple tables or update old records.

#### BridgeSyncer API Changes

**Updated `GetClaims` signature:**
```go
// Before
func GetClaims(ctx context.Context, fromBlock, toBlock uint64) ([]Claim, error)

// After  
func GetClaims(ctx context.Context, fromBlock, toBlock uint64, compacted bool) ([]Claim, error)
```
**Behavior**:

`compacted=false`: Returns all claims in the block range, preserving complete history, meaning there can be multiple claims with same global index.
`compacted=true`: Returns compacted claims where duplicates are merged intelligently.

**Compaction logic**:

When `compacted=true`, claims with the same `global_index` are merged using this strategy:
- Preserve all fields from the oldest claim (`block_num`, `block_pos`, `metadata`, `amount`, `addresses`, etc.)
- Update only the proof fields from the newest claim (`proof_local_exit_root`, `proof_rollup_exit_root`, `mainnet_exit_root`, `rollup_exit_root`, `global_exit_root`)

**Updated `GetClaimsPaged` logic**

Before, the function returned all claims with the same global index.
Now, the function will return compacted claims always, since this function is only used by the `bridge service`.

Key Changes:
1. **Pagination happens first** on raw claims (ordered by block_num DESC, block_pos DESC)
2. **Compaction applies within each page** - but with a critical rule:
   - Claims are only returned if the newest claim with that global_index appears on the current page
    - If older claims appear on a page but the newest one is on a different page, they are excluded
3. Compaction logic (same as `GetClaims`):
    - Takes only the proof fields from the newest claim (`proof_local_exit_root`, `proof_rollup_exit_root`, `mainnet_exit_root`, `rollup_exit_root`, `global_exit_root`)

This ensures:
- ✅ Original claim metadata and transaction details are preserved
- ✅ Most recent proofs are available for successful claim execution
- ✅ Block ordering reflects the first appearance of each claim

#### Aggsender Changes
`FEP` (`AggchainProof`) mode - will call the `GetClaims` function with `compacted=false` because the `aggkit prover` needs the entire history of the claims, not just the latest correct one.
`PP` (`PessimisticProof`) mode - will cal the `GetClaims` function with `compacted=true` because it doesn't need the entire history of the claims, just the latest correct one.

## ⚠️ Breaking Changes
NA

## 📋 Config Updates
NA

## ✅ Testing

## 🐞 Issues
- Closes #1330

### DIFF
--- a/aggsender/flows/builder_flow_aggchain_prover.go
+++ b/aggsender/flows/builder_flow_aggchain_prover.go
@@ -122,7 +122,7 @@ func (a *AggchainProverBuilderFlow) CheckInitialStatus(ctx context.Context) erro
 	}
 
 	if err := a.baseFlow.VerifyBlockRangeGaps(
-		ctx, lastSentCertificate, startL2Block, startL2Block); err != nil {
+		ctx, lastSentCertificate, startL2Block, startL2Block, false); err != nil { // FEP does not use compacted claims
 		return fmt.Errorf("aggchainProverFlow - error verifying block range gaps on startup. Err: %w", err)
 	}
 
@@ -151,7 +151,7 @@ func (a *AggchainProverBuilderFlow) GenerateBuildParams(ctx context.Context,
 		return nil, fmt.Errorf("aggchainProverFlow - preParams is nil")
 	}
 
-	params, err := a.baseFlow.GenerateBuildParams(ctx, *preParams)
+	params, err := a.baseFlow.GenerateBuildParams(ctx, *preParams, false) // FEP mode does not use compacted claims
 	if err != nil {
 		return nil, fmt.Errorf("aggchainProverFlow - error generating build params: %w", err)
 	}
@@ -198,6 +198,7 @@ func (a *AggchainProverBuilderFlow) GetCertificateBuildParams(
 		bridges, claims, err := a.l2BridgeQuerier.GetBridgesAndClaims(
 			ctx, fromBlock,
 			toBlock,
+			false, // do not include compacted claims
 		)
 		if err != nil {
 			return nil, fmt.Errorf("aggchainProverFlow - error getting bridges and claims: %w", err)
@@ -251,7 +252,7 @@ func (a *AggchainProverBuilderFlow) GetCertificateBuildParams(
 			lastSentCert.CertType, typeCert)
 	}
 
-	buildParams, err := a.baseFlow.GetCertificateBuildParamsInternal(ctx, typeCert)
+	buildParams, err := a.baseFlow.GetCertificateBuildParamsInternal(ctx, typeCert, false) // FEP mode does not use compacted claims
 	if err != nil {
 		if errors.Is(err, errNoNewBlocks) {
 			// no new blocks to send a certificate
--- a/aggsender/flows/builder_flow_aggchain_prover.go
+++ b/aggsender/flows/builder_flow_aggchain_prover.go
@@ -198,7 +198,7 @@ func (a *AggchainProverBuilderFlow) GetCertificateBuildParams(
 		bridges, claims, err := a.l2BridgeQuerier.GetBridgesAndClaims(
 			ctx, fromBlock,
 			toBlock,
-			false, // do not include compacted claims
+			false, // do not compact claims
 		)
 		if err != nil {
 			return nil, fmt.Errorf("aggchainProverFlow - error getting bridges and claims: %w", err)
--- a/aggsender/flows/builder_flow_aggchain_prover.go
+++ b/aggsender/flows/builder_flow_aggchain_prover.go
@@ -122,7 +122,7 @@ func (a *AggchainProverBuilderFlow) CheckInitialStatus(ctx context.Context) erro
 	}
 
 	if err := a.baseFlow.VerifyBlockRangeGaps(
-		ctx, lastSentCertificate, startL2Block, startL2Block, false); err != nil { // FEP does not use compacted claims
+		ctx, lastSentCertificate, startL2Block, startL2Block); err != nil { // FEP does not use compacted claims
 		return fmt.Errorf("aggchainProverFlow - error verifying block range gaps on startup. Err: %w", err)
 	}
 
@@ -151,7 +151,7 @@ func (a *AggchainProverBuilderFlow) GenerateBuildParams(ctx context.Context,
 		return nil, fmt.Errorf("aggchainProverFlow - preParams is nil")
 	}
 
-	params, err := a.baseFlow.GenerateBuildParams(ctx, *preParams, false) // FEP mode does not use compacted claims
+	params, err := a.baseFlow.GenerateBuildParams(ctx, *preParams)
 	if err != nil {
 		return nil, fmt.Errorf("aggchainProverFlow - error generating build params: %w", err)
 	}
@@ -252,7 +252,7 @@ func (a *AggchainProverBuilderFlow) GetCertificateBuildParams(
 			lastSentCert.CertType, typeCert)
 	}
 
-	buildParams, err := a.baseFlow.GetCertificateBuildParamsInternal(ctx, typeCert, false) // FEP mode does not use compacted claims
+	buildParams, err := a.baseFlow.GetCertificateBuildParamsInternal(ctx, typeCert)
 	if err != nil {
 		if errors.Is(err, errNoNewBlocks) {
 			// no new blocks to send a certificate
--- a/bridgesync/processor.go
+++ b/bridgesync/processor.go
@@ -38,9 +38,6 @@ const (
 	// claimTableName is the name of the table that stores claim events
 	claimTableName = "claim"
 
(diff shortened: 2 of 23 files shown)


aggsender/flows/builder_flow_aggchain_prover.go
```diff
@@ -198,6 +198,7 @@ func (a *AggchainProverBuilderFlow) GetCertificateBuildParams(
 		bridges, claims, err := a.l2BridgeQuerier.GetBridgesAndClaims(
 			ctx, fromBlock,
 			toBlock,
+			false, // do not include compacted claims
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/1336/commits/1bf6cdfc73d0cfdbf338e295c367a43f51178016

bridgesync/processor_test.go
```diff
@@ -2980,3 +2879,1000 @@ func TestDeleteClaimReason_String(t *testing.T) {
 		})
 	}
 }
+
+//nolint:dupl
+func TestGetClaimsNew(t *testing.T) {
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/1336/commits/2f98dba9e8cc2d69ab3de60bd1ef7b3aba250a2a

bridgesync/processor.go
```diff
@@ -508,7 +505,92 @@ func (p *processor) GetBridges(
...
+	claimPtrs := []*Claim{}
+	if err = meddler.ScanAll(rows, &claimPtrs); err != nil {
+		p.log.Errorf("GetClaimsNew: meddler.ScanAll failed for block range [%d..%d]: %v", fromBlock, toBlock, err)
+		return nil, err
+	}
+
+	claimsIface := db.SlicePtrsToSlice(claimPtrs)
+	claims, ok := claimsIface.([]Claim)
+	if !ok {
+		p.log.Errorf("GetClaimsNew: failed to convert from []*Claim to []Claim for block range [%d..%d]", fromBlock, toBlock)
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/1336/commits/2f98dba9e8cc2d69ab3de60bd1ef7b3aba250a2a

bridgesync/processor.go
```diff
@@ -508,7 +505,92 @@ func (p *processor) GetBridges(
...
+
+	defer func() {
+		if cerr := rows.Close(); cerr != nil {
+			p.log.Errorf("error closing rows: %v", cerr)
+		}
+	}()
+
+	claimPtrs := []*Claim{}
+	if err = meddler.ScanAll(rows, &claimPtrs); err != nil {
+		p.log.Errorf("GetClaimsNew: meddler.ScanAll failed for block range [%d..%d]: %v", fromBlock, toBlock, err)
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/1336/commits/2f98dba9e8cc2d69ab3de60bd1ef7b3aba250a2a

bridgesync/processor.go
```diff
@@ -508,7 +505,92 @@ func (p *processor) GetBridges(
...
+		ORDER BY block_num ASC, block_pos ASC;
+	`
+
+	rows, err := p.db.QueryContext(dbCtx, query, fromBlock, toBlock)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			p.log.Debugf("no claims were found for block range [%d..%d]", fromBlock, toBlock)
+			return []Claim{}, nil
+		}
+		p.log.Errorf("GetClaimsNew: query failed for block range [%d..%d]: %v", fromBlock, toBlock, err)
```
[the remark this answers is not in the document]
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/1336/commits/2f98dba9e8cc2d69ab3de60bd1ef7b3aba250a2a

bridgesync/processor.go
```diff
@@ -75,6 +72,27 @@ func (d DeleteClaimReason) String() string {
 const (
 	// orderByBlockDesc is the default order by clause for block-based queries
 	orderByBlockDesc = "block_num DESC, block_pos DESC"
+
+	// compactedClaimsSelectSQL is the SELECT clause for compacted claims
+	// It combines metadata from the oldest claim with proofs from the newest claim
```
Stefan-Ethernal:
```suggestion
	// It combines metadata from the oldest claim with proofs and exit roots from the newest claim
```

bridgesync/processor.go
```diff
@@ -978,39 +1105,6 @@ func (p *processor) ProcessBlock(ctx context.Context, block sync.Block) error {
...
-
-				if len(existingClaims) > 0 {
-					_, err = tx.Exec(`DELETE FROM claim WHERE global_index = ?`, globalIndex.String())
-					if err != nil {
-						p.log.Errorf("failed to delete claims for global index %d: %v", globalIndex, err)
-						return err
-					}
-
-					for _, existingClaim := range existingClaims {
-						invalidClaim := NewInvalidClaim(&existingClaim, InvalidGERClaimCorrect.String())
```
Stefan-Ethernal, line 1005: We can remove this [enum](https://github.com/agglayer/aggkit/blob/eabc00da5a634a9b6750e658674134ce92133e08/bridgesync/processor.go#L54-L70), it is not used anymore.

bridgesync/processor.go
```diff
@@ -505,7 +523,71 @@ func (p *processor) GetBridges(
...
+	claims, ok := claimsIface.([]Claim)
+	if !ok {
+		p.log.Errorf("GetClaims: failed to convert from []*Claim to []Claim for block range [%d..%d]", fromBlock, toBlock)
+		return nil, errFailToConvertClaims
+	}
+
+	return claims, nil
+}
+
+func (p *processor) getClaims(ctx context.Context, fromBlock, toBlock uint64) ([]Claim, error) {
```
Stefan-Ethernal: We can probably change getClaims function, so that it receives query string so that this [part](https://github.com/agglayer/aggkit/blob/eabc00da5a634a9b6750e658674134ce92133e08/bridgesync/processor.go#L558-L585) reuses this function (or something similar) to avoid code repetitions. The idea is that both compacted and non compacted versions are relying on same `getClaimsInternal` function. WDYT?

Stefan-Ethernal (approved):
Generally LGTM, but leaving a couple of suggestions.

bridgesync/processor.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/1336/commits/1088f38b7b05ab7e8c686f6d73a0fb9f9948a88f

bridgesync/processor.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/1336/commits/1088f38b7b05ab7e8c686f6d73a0fb9f9948a88f

bridgesync/processor.go
goran-ethernal: Fix: https://github.com/agglayer/aggkit/pull/1336/commits/1088f38b7b05ab7e8c686f6d73a0fb9f9948a88f

temaniarpit27 (approved):

PR #1362, opened 2025-12-03
fix: compaction logic with unset claims consideration

goran-ethernal:
## 🔄 Changes Summary
This PR modifes the compaction logic of `claims` to take into consideration the `unset claim` events as well to do compaction logic.

The compaction logic for getting claims for a block range (`certificates`) works like this now:
- Find all claims with the same global_index across ALL blocks (not just in range)
- Check if there's an unset claim for that global_index
- If no unset claim AND the globally oldest claim is within the query range, return the compacted claim (take everything from the old claim, except the proof and exit root fields which are taken from the newest claim)
- Otherwise, return nothing for that global_index

For the `bridge service` logic when queried by page:
- Returns all claims on the page uncompacted if an `unset_claim` exists for that `global_index`.
- For claims without `unset_claims`, compacts them by combining oldest claim metadata with newest claim proofs, but only if the newest claim appears on the requested page
- Claims without `unset_claims` where the newest instance is not on the current page are excluded from results

The same logic was applied to functions:
- `GetClaims`
- `GetClaimsPaged`
- `GetClaimsByGlobalIndex`

## ⚠️ Breaking Changes
NA

## 📋 Config Updates
NA

## ✅ Testing

## 🔗 Related PRs
#1336

### DIFF
--- a/bridgesync/processor.go
+++ b/bridgesync/processor.go
@@ -59,6 +59,25 @@ const (
 	// orderByBlockDesc is the default order by clause for block-based queries
 	orderByBlockDesc = "block_num DESC, block_pos DESC"
 
+	// claimColumnsSQL is the list of all claim columns
+	claimColumnsSQL = `block_num,
+		block_pos,
+		tx_hash,
+		global_index,
+		origin_network,
+		origin_address,
+		destination_address,
+		amount,
+		proof_local_exit_root,
+		proof_rollup_exit_root,
+		mainnet_exit_root,
+		rollup_exit_root,
+		global_exit_root,
+		destination_network,
+		metadata,
+		is_message,
+		block_timestamp`
+
 	// compactedClaimsSelectSQL is the SELECT clause for compacted claims
 	// It combines metadata from the oldest claim with proofs and exit roots from the newest claim
 	compactedClaimsSelectSQL = `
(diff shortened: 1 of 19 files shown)


bridgesync/processor.go
```diff
@@ -59,6 +59,25 @@ const (
...
+		amount,
+		proof_local_exit_root,
+		proof_rollup_exit_root,
+		mainnet_exit_root,
+		rollup_exit_root,
+		global_exit_root,
+		destination_network,
+		metadata,
+		is_message,
+		block_timestamp`
```
Stefan-Ethernal, line 79: Is it doable to craft some unit test that would use the reflection to check if all the fields in the `Claim` object, annotated by `meddler` tag, are enumerated here? Also it would imply to change the field to string slice (so it is easier to manipulate in the test), and where necessary to use it, when building the query and injecting args to query, use `strings.Join(...)`. WDYT?

Stefan-Ethernal (approved):

bridgesync/processor.go
goran-ethernal: Yes, I had the same idea. I am trying  to write a test as we speak.

temaniarpit27 (approved):

bridgesync/processor.go
goran-ethernal: Implemented the test: https://github.com/agglayer/aggkit/pull/1362/commits/5b7110f438b1694f538426bf6ce9aab9ac74054e

PR #1372, opened 2025-12-04
fix: From field for bridge event syncronization 

### LINKED ISSUE #1346: bug: bridgesync: fullsync: multiples bridgeXXX are not going to be well synced
# Bug Report
NOTE: 
- `FromAddress`: the subcall `from`
- `TxnSender`: the whole tx `from`

The way that currently is synced the field `FromAddress` for `bridgeEvent` is going to fail in case of multiples `bridgeXXX` in the same transaction:
- Tx1: 
    - logs: 
       -  1. ClaimEvent 
       -  2. BridgeEvent
       -  3. BridgeEvent
    - internal calls
       - from: **Addr1**, to: BridgeAddr methoidID:  **claimAsset**
       - from: **Addr2**, to: BridgeAddr methoidID:  **bridgeAsset**
       - from: **Addr3**, to: BridgeAddr methoidID:  **bridgeAsset**

## The way that is checked before issue #1329 
It checked only the `to:` field so would assign **Addr1** to all of them: 
- 2. BridgeEvent `FromAddress` : **Addr1**  ❌
- 3. BridgeEvent `FromAddress` : **Addr1** ❌

## After fixing the issue #1329 
It checked only `to:` field, the `methodID` and the leaf type (that is not releveant for this example):
- 2.  BridgeEvent `FromAddress` : **Addr2**  ✔️
- 3. BridgeEvent `FromAddress` : **Addr2** ❌

Conclusion it require a better discriminator: 
- ? number of event vs number of calls? 
-

### LINKED ISSUE #1356: bug: bridgesync: detect possible rows affected by bug #1329 and update in background
# Feature Request

The issue #1329 affect to entries with more 1 bridge and 1 claim in the same transaction.

joanestebanr:
## 🔄 Changes Summary
- Cover all cases in MAINNET
- In case of find a event / calls that are not able to solve returns an error (avoid filling with wrong data)
- Upgrade in the background the wrong values in bridges From address
 

## ✅ Testing
- 🤖 **Automatic**: Unit Test
- 🖱️ **Manual**:Applied to mainnet sincronization with version `0.7.2` till block 23933203

## 🐞 Issues
- Closes #1356
- Closes #1346

## 🔗 Related PRs
- #1355


bridgesync/migrations/bridgesync0011.sql
```diff
@@ -0,0 +1,15 @@
+-- +migrate Down
+
+-- +migrate Up
+UPDATE bridge
+SET from_address = NULL
+WHERE tx_hash IN (
+    SELECT b.tx_hash
+    FROM bridge b
+    JOIN claim c ON b.tx_hash = c.tx_hash
```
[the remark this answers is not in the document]
joanestebanr: It's a small table and I prioritize clarity over performance

temaniarpit27 (approved):

bridgesync/migrations/bridgesync0011.sql
```diff
@@ -0,0 +1,15 @@
...
+);
+UPDATE bridge
+SET from_address = NULL
+WHERE tx_hash IN (
+    SELECT  tx_hash FROM bridge GROUP BY tx_hash HAVING COUNT(*) > 1
```
Stefan-Ethernal: Nitpick, extra space:
```suggestion
    SELECT tx_hash FROM bridge GROUP BY tx_hash HAVING COUNT(*) > 1 
```

bridgesync/downloader.go
```diff
@@ -115,8 +120,225 @@ func buildAppender(
 	return appender, nil
 }
 
+// Transaction represents the structure of a transaction returned by eth_getTransactionByHash
+type Transaction struct {
```
vcastellm, line 124: should we use go-ethereum types?

bridgesync/downloader.go
```diff
@@ -115,8 +120,225 @@ func buildAppender(
...
+
+func (t *Transaction) From() common.Address {
+	return common.HexToAddress(t.FromRaw)
+}
+
+func RPCTransactionByHash(client aggkittypes.EthClienter,
+	txHash common.Hash) (*Transaction, error) {
+	// Use client.Call to fetch transaction details using eth_getTransactionByHash
+	var tx Transaction
+	err := client.Call(&tx, "eth_getTransactionByHash", txHash.Hex())
```
vcastellm, line 146: Same here, wouldn't it be better to use https://github.com/ethereum/go-ethereum/blob/472e3a24ac17626cac07c52704ff7f5706260792/ethclient/ethclient.go#L275 ?

bridgesync/downloader.go
```diff
@@ -115,8 +120,225 @@ func buildAppender(
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch transaction by hash: %w", err)
+	}
+	return &tx, nil
+}
```
Stefan-Ethernal, line 151: This function is not needed, you can invoke, from `go-ethereum`. The `aggkittypes.EthClienter` already implements the `TransactionReader`:

```go
type TransactionReader interface {
   TransactionByHash(ctx context.Context, txHash common.Hash) (tx *types.Transaction, isPending bool, err error)
   ...
}
```
```suggestion
```

Also as @vcastellm pointed out, we can then remove the custom `Transaction` struct and reuse the one from go-ethereum.

bridgesync/downloader.go
joanestebanr: The same reason as previous comment

bridgesync/downloader.go
joanestebanr: This was there from  previous implementation, I just moved. 
But, after checking the code I understand that is using custom transaction because the way that go-ethereum calculate the `From` field from the ECDSA signature

bridgesync/downloader.go
joanestebanr: I explained in previous comment the reson: 
- Because it was the way that was done previously
- Performance reason

bridgesync/backfill_tx_sender_test.go
```diff
@@ -1322,14 +1317,23 @@ func TestBackfillTxnSender_processBatch_Comprehensive(t *testing.T) {
...
-			tx.To = "0x1234"
-		})
+		// mockClient.On("Call", mock.Anything, "eth_getTransactionByHash", mock.Anything).Return(nil).Run(func(args mock.Arguments) {
+		// 	tx, ok := args.Get(0).(*Transaction)
+		// 	if !ok {
+		// 		return
+		// 	}
+		// 	tx.FromRaw = testAddress
+		// 	tx.To = "0x1234"
+		// })
```
Stefan-Ethernal:
```suggestion
```

bridgesync/downloader.go
Stefan-Ethernal: > Because it was the way that was done previously

This is not the justification, if the code was already wrong and it takes minimal effort to make it better, why don't we take this opportunity

> Performance reason

This point I can understand to some extent, because of ECDSA recover to calculate the sender address, although AFAIK the from field in the transaction gets cached

I don't agree entirely on this, but your call.

bridgesync/downloader.go
```diff
@@ -115,8 +120,225 @@ func buildAppender(
...
+	for i, call := range foundCalls {
+		if i == 0 {
+			commonFrom = call.From
+		} else if call.From != commonFrom {
+			return common.Address{}, false
+		}
+	}
+
+	return commonFrom, true
+}
```
Stefan-Ethernal:
```suggestion
func haveCommonFromForCalls(calls []*Call) (common.Address, bool) {
	if len(calls) == 0 {
		return common.Address{}, false
	}

	commonFrom := calls[0].From
	for _, call := range calls[1:] {
		if call.From != commonFrom {
			return common.Address{}, false
		}
	}

	return commonFrom, true
}
```

Stefan-Ethernal (approved):

PR #1379, opened 2025-12-08
feat: index `BackwardLET` event

### LINKED ISSUE #1359: index `BackwardLET` event in bridge syncer
Handle the `BackwardLET` in the L2 bridge syncer.

### Function signature
```solidity
function backwardLET(
        uint256 newDepositCount,
        bytes32[_DEPOSIT_CONTRACT_TREE_DEPTH] calldata newFrontier,
        bytes32 nextLeaf,
        bytes32[_DEPOSIT_CONTRACT_TREE_DEPTH] calldata proof
    ) external virtual onlyGlobalExitRootRemover ifEmergencyState
```

### Event signature
```solidity
event BackwardLET(
        uint256 previousDepositCount,
        bytes32 previousRoot,
        uint256 newDepositCount,
        bytes32 newRoot
    );
```

### Behavior
Once we index the event:
- removing all the bridges that have the `depositCount` field greater than `newDepositCount` from the event in the `bridge` table, and moving them to a separate table. 
   - TBD reorgs handling, in a way that if certain bridge that was affected by `BackwardLET` later on the `BackwardLET` gets reorged,
- figure out how to update the `exitTree`, which is `AppendOnlyTree` https://github.com/agglayer/aggkit/blob/9ba13e37f74e7220018a36662eaa81a8da529e13/bridgesync/processor.go#L569. Should it become the `UpdateableTree`? 🤔 

### Open questions
1. (L2) reorgs handling, because it seems that bridges should be removed from bridge table,
2. `exitTree` should be `UpdatableTree`, since now we should be able to update its content (namely remove leafs)?

Stefan-Ethernal:
## 🔄 Changes Summary
Implement support for indexing of `BackwardLET` event in the bridge syncer. Also implemented `BackwardToIndex` method on the tree to roll back the Merkle tree to a specific index by deleting higher-indexed roots

### Index of `BackwardLET` event
Once we index the `BackwardLET` event, we remove all the bridges whose deposit_count is greater than the new deposit count from the event. We also store the removed bridges into newly introduced `bridge_archive` table, in order to be able to bring them back in case `BackwardLET` event gets reorged.

### Reorg of `BackwardLET` event
Once certain `BackwardLET` events get  reorged, we figure out which ranges of bridges, by deposit counts, are affected and we restore them back from the `bridge_archive` into the `bridge` table.

## ⚠️ Breaking Changes
N/A

## 📋 Config Updates
N/A

## ✅ Testing

## 🐞 Issues
- Closes #1359

## 🔗 Related PRs

## 📝 Notes


bridgesync/processor.go
```diff
@@ -1332,6 +1355,37 @@ func (p *processor) ProcessBlock(ctx context.Context, block sync.Block) error {
...
+			}
+
+			leafIndex := uint32(newDepositCountU64 - 1)
+			if err := p.exitTree.BackwardToIndex(ctx, tx, leafIndex); err != nil {
+				p.log.Errorf("failed to backward local exit tree to %d deposit count", leafIndex)
+				return err
+			}
+
+			// remove all the bridges whose deposit_count is greater than the one captured by the BackwardLET event
+			deleteBridges := fmt.Sprintf("DELETE from %s WHERE deposit_count > $1", bridgeTableName)
```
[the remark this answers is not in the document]
Stefan-Ethernal: The intention is to leave the bridge with the `NewDepositCount` in the storage, both on smart contracts and in the aggkit storage.

bridgesync/migrations/bridgesync0012.sql
```diff
@@ -1,6 +1,87 @@
...
+	);
+
+ALTER TABLE bridge ADD COLUMN source TEXT;
 ALTER TABLE bridge ADD COLUMN to_address VARCHAR;
 
+------------------------------------------------------------------------------
+-- Create bridge_archive table
+------------------------------------------------------------------------------
+CREATE TABLE IF NOT EXISTS bridge_archive (
+		deposit_count INTEGER PRIMARY KEY,
```
joanestebanr, line 33: It's possible that the same `deposit_count` to be backwarded twice? maybe the PRIMARY KEY must be another, WDYT?

bridgesync/migrations/bridgesync0012.sql
```diff
@@ -1,6 +1,87 @@
...
+		destination_network INTEGER NOT NULL,
+		destination_address VARCHAR NOT NULL,
+		amount TEXT NOT NULL,
+		metadata BLOB,
+		tx_hash VARCHAR,
+		block_timestamp INTEGER,
+		txn_sender VARCHAR,
+        from_address VARCHAR,
+        source TEXT,
+        to_address VARCHAR
```
joanestebanr, line 48: maybe make sense to add a reference to the `backward_let` that produce this? WDYT?

bridgesync/migrations/migrations.go
```diff
@@ -1,100 +1,70 @@
...
 )
 
-//go:embed bridgesync0001.sql
-var mig0001 string
-
-//go:embed bridgesync0002.sql
-var mig0002 string
+var (
+	//go:embed *.sql
+	migrationFS embed.FS
```
joanestebanr, line 16: really good improvement! 👏

bridgesync/migrations/bridgesync0012.sql
Stefan-Ethernal: Nope, this is not the valid use case, according to the `AgglayerBridgeL2` [implementation](https://github.com/agglayer/agglayer-contracts/blob/f27cc65f4b2ed5dcd4342c6d8f903738ac439418/contracts/sovereignChains/AgglayerBridgeL2.sol#L739-L741). The provided `newDepositCount` parameter must be strictly less than the `depositCount` stored on the `AgglayerBridgeL2` contract (meaning that single bridge can be backwarded only once). So I'd say we are safe with this regard and no need to change the primary key.

bridgesync/migrations/bridgesync0012.sql
Stefan-Ethernal: Adding a foreign key from `bridge_archive` to `backward_let` doesn’t make sense because the relationship between them is range-based, not row-based.

A single `backward_let` affects many `bridge_archive` rows, defined by a `deposit_count` range (`new_deposit_count` < `deposit_count` ≤ `previous_deposit_count`). Relational foreign keys can only model direct ownership (1-to-1 or many-to-1), not “belongs to a range”.

Also, `block_num` / `block_pos` in bridge_archive represent the original bridge block, while in `backward_let` they represent the block when the `backward_let` event was indexed; they’re different domains, so linking them would be semantically incorrect.

The relationship is fully derivable, deterministic, and easier to reason about by querying on `deposit_count` ranges. Persisting it via a FK would add redundancy, risk inconsistency during reorgs, and complicate deletes without providing real integrity benefits.

tree/tree.go
```diff
@@ -133,14 +133,11 @@ func (t *Tree) getRHTNode(tx dbtypes.Querier, nodeHash common.Hash) (*types.Tree
...
-		if err := meddler.Insert(tx, t.rhtTable, &nodes[i]); err != nil {
-			if sqliteErr, ok := db.SQLiteErr(err); ok {
-				if sqliteErr.ExtendedCode == db.UniqueConstrain {
-					// ignore repeated entries. This is likely to happen due to not
-					// cleaning RHT when reorg
-					continue
-				}
+	for _, node := range nodes {
+		if err := meddler.Insert(tx, t.rhtTable, &node); err != nil {
+			if sqliteErr, ok := db.SQLiteErr(err); ok && sqliteErr.ExtendedCode == db.UniqueConstrain {
```
temaniarpit27: is it `UniqueConstrain` or `UniqueConstraint`?

bridgesync/processor.go
```diff
@@ -1382,6 +1496,34 @@ func (p *processor) ProcessBlock(ctx context.Context, block sync.Block) error {
 				return err
 			}
 		}
+
+		if event.BackwardLET != nil {
+			newDepositCount, leafIndex, err := normalizeDepositCount(event.BackwardLET.NewDepositCount)
```
temaniarpit27, line 1536: why are we doing this here? isnt depositcount already uint32?

Stefan-Ethernal: Nope, on the smart contracts level, the `newDepositCount` is represented as `uint256` (https://github.com/agglayer/agglayer-contracts/blob/f27cc65f4b2ed5dcd4342c6d8f903738ac439418/contracts/sovereignChains/AgglayerBridgeL2.sol#L208), which gets translated to `big.Int` in the Golang bindings. That's why we need to do sanity checks and down cast it in case it is safe to do.

tree/tree.go
Stefan-Ethernal: Fixed https://github.com/agglayer/aggkit/pull/1379/commits/228956531b07a8ff756f2bbfc1bd4a391bf047b7

temaniarpit27 (approved):

rachit77 (approved):

PR #1397, opened 2025-12-12
feat: hash from rpc instead of calculated v2 1389

### LINKED ISSUE #1389: bug: some L2 doesn't match hash with calculated
# Bug Report
Reorg detector complain because the Hash of block in L2 doesn't match the calculated from fields

```
2025-12-10T11:21:56.652+0100	WARN	reorgdetector/reorgdetector.go:225	Reorg detected l2 for subscriber L2BridgeSyncer between blocks 34796290 and 34796290. currentHash: 0xb1916a642d65cf6182829c542f5e3ef7c498f242d640184c1738735721957b3f trackHash: 0xc3e563ddff3a91e43bf1af02eb1ada8f576f1612eddf95c10bac203c396ffee8	{"pid": 1484136, "version": "v0.1.0", "reorg-detector": "l2"}
github.com/agglayer/aggkit/reorgdetector.(*ReorgDetector).detectReorgInTrackedList.func1
	/home/jesteban/POLYGON/src/aggkit/aggkit-develop4/reorgdetector/reorgdetector.go:225
golang.org/x/sync/errgroup.(*Group).Go.func1
	/home/jesteban/go/pkg/mod/golang.org/x/sync@v0.18.0/errgroup/errgroup.go:93
2025-12-10T11:21:56.652+0100	WARN	sync/evmdriver.go:201	Reorg detected from block 34796290, stopping block processing...	{"pid": 1484136, "version": "v0.1.0", "syncer": "L2BridgeSyncer"}
github.com/agglayer/aggkit/sync.(*EVMDriver).Sync
	/home/jesteban/POLYGON/src/aggkit/aggkit-develop4/sync/evmdriver.go:201
github.com/agglayer/aggkit/bridgesync.(*BridgeSync).Start
	/home/jesteban/POLYGON/src/aggkit/aggkit-develop4/bridgesync/bridgesync.go:309
2025-12-10T11:21:56.652+0100	INFO	sync/evmdriver.go:177	cancellableCtx done, exiting block handler goroutine	{"pid": 1484136, "version": "v0.1.0", "syncer": "L2BridgeSyncer"}
2025-12-10T11:21:56.652+0100	ERROR	sync/evmdownloader.go:471	context is canceled getUnfilteredLogs, returning nil	{"pid": 1484136, "version": "v0.1.0", "syncer": "L2BridgeSyncer"}
```

## Environment (please complete the following information):
- environment:  `bali-63-outpost`
- version: `0.7.x`

## Additional context
- https://0xpolygon.slack.com/archives/C07HZQHN6RY/p1765204159899149
- Thaigo PR: [https://github.com/agglayer/aggkit/compare/develop...tnobayashi/experiments#diff-c722091b8b989e579ab20f2c90135[…]665d207ee0aaaab182afff4524c829bR513](https://github.com/agglayer/aggkit/compare/develop...tnobayashi/experiments#diff-c722091b8b989e579ab20f2c90135156e665d207ee0aaaab182afff4524c829bR513)

joanestebanr:
## 🔄 Changes Summary
- The eth client could use `Hash` field as blockHash in rpc Json response instead of calculating it from fields (default behaviour for **go-ethereum** package)

### Add a new function to `EthClienter` interface: 
```
CustomHeaderByNumber(ctx context.Context, number *BlockNumberFinality) (*BlockHeader, error)
```
- It requires a new name because can change `HeaderByNumber` because it returns a specific type of *go-ethereum* and `Hash()` function is not possible to override
- The new function use internal types, for blockNumber and the  block header struct

###  Refactor `BlockNumberFinality`
- It supports a number (so it can be used to retrieve a named block or a specific one)
-  Previous type `BlockNumber` have been renamed to `BlockName` to avoid confusions
- Removed `BlockNumberFinality.BlockHeader` now it's supported directly by `CustomHeaderByNumber`

### Removed object `NoopRPCClient`:
- If the RPC direct call is not availablee is handle by `DefaultEthClient`
- If you call to any direct call not supported returns an error (to avoid silently by pass errors)

### Unified eth clients configuration 
- The configuration struct `L2RPCClientConfig` have extended to L1, also. To keep the previous behaviour L1 client can on be mode `basic`

### Always create L1 client and use it for `RollupDataQuerier`
- The object `RollupDataQuerier` is always created (not related to components) so the L1 client need to be created always

## ⚠️ Breaking Changes
N/A
## 📋 Config Updates
- 🛠️ **Config**: Add config field `HashFromJSON` for rpc clients: 
```
[L2RPC] 
	HashFromJSON = true

	[L1NetworkConfig.RPC]
		HashFromJSON = true
```
If the param `HashFromJSON` is true use a direct  the Hash field instead of calculating, this require to pass a valid RPCClient with support for directs Calls (so the internals e2e that use simulated environment can't use this feature)

## ✅ Testing
- 🤖 **Automatic**: e2e, unittest
- **Manually**: run a bridgesyncer against network: bali-63-outpost

## 🐞 Issues
- Closes #1389

## Pending works
- Remove specific calls to `rpc.Call`, instead incorporate it to ethClienter:
    - bridgesyncer: `eth_getTransactionByHash`
    - bridgesyncer: `debug_traceTransaction`
- Rename multidownloader `HeaderByNumber` to `CustomHeaderByNumber`
-


types/block_finality.go
```diff
@@ -39,36 +39,81 @@ var (
...
+
+		if err == nil {
+			res, ok := result.(T)
+			if !ok {
+				return zeroValue, fmt.Errorf("type assertion failed during number conversion")
+			}
+			return res, nil
+		}
+	}
+	return zeroValue, fmt.Errorf("invalid block number format: %s", s)
```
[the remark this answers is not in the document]
joanestebanr: Removed multiples bases

cmd/run.go
```diff
@@ -887,17 +876,9 @@ func startPrometheusHTTPServer(c prometheus.Config) {
...
-	if err != nil {
-		log.Fatalf("failed to create retry handler: %w", err)
-	}
-
-	ethClient, err := aggkittypes.DialWithRetry(ctx, cfg.RPC.URL, retryHandler)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create Ethereum client for L1 using URL: %s. Err: %w", cfg.RPC.URL, err)
-	}
-	return ethermanquierier.NewRollupDataQuerier(ctx, cfg, ethClient,
+	return ethermanquierier.NewRollupDataQuerier(ctx, cfg, l1Client,
```
temaniarpit27, line 881: nit: ethermanquerier

log/log_nil.go
```diff
@@ -0,0 +1,50 @@
+package log
```
temaniarpit27, line 1: where are we using this?

temaniarpit27 (approved):

vcastellm (changes requested):
Still reviewing but I'm not completely sure about this solution, the differences on the hash calculation are because of op-geth differences in block header, etc. so it looks like the proper client should be used instead of coming up with our own trickery here. See https://github.com/ethereum-optimism/op-geth/tree/optimism/ethclient

Let me think a bit more on this

log/log_nil.go
joanestebanr: If the log is `nil` then create a new instance of `LoggerNil` [here](https://github.com/agglayer/aggkit/blob/feat/hash_from_rpc_instead_of_calculated_v2-1389/etherman/default_eth_client.go#L42)

cmd/run.go
joanestebanr: sorry, I can't see the difference!  😞

vcastellm (commented):
Looks like we should select the op-stack ethClient for L2 syncers based on op

etherman/default_eth_client.go
```diff
@@ -0,0 +1,172 @@
...
+		number = &aggkittypes.LatestBlock
+	}
+	// The number can have an offset, so maybe we need to resolve the blockName, apply offset to require the header
+	numberBigInt, err := c.resolveBlockNumber(ctx, number)
+	if err != nil {
+		return nil, err
+	}
+	var result *aggkittypes.BlockHeader
+	if c.HashFromJSON {
+		result, err = c.rpcGetBlockByNumber(ctx, numberBigInt)
```
vcastellm, line 106: This is called here too https://github.com/agglayer/aggkit/pull/1397/changes#diff-b1f974dfe4dd0deb115cab293014c56d3ceac09854aa5a5adeca534426c5bfd6R129 so we can save a call to the rpc if we collapse the two calls, there are several ways to do it

cmd/run.go
temaniarpit27: `i`

log/log_nil.go
temaniarpit27: in what cases, the logger will be nil?

joanestebanr: yes, you pass to the constructor function a nil as logger and internally it's filled  with a instance of `LoggerNil`

etherman/default_eth_client.go
joanestebanr: which call do you refer? `rpcGetBlockByNumber`? can you share again the link, it's not clear to the code that you mean

joanestebanr:
> Looks like we should select the op-stack ethClient for L2 syncers based on op

I suggest that after applying this PR that abstract the client we can do another PR that allow to choose a new mode = 'op-stack' and use the more convinient client, WDYT?

cmd/run.go
joanestebanr: ok, I understand, you think that is better to remove checking the error. I did in that way to it's more easy to find the source of the error when you see in logs

temaniarpit27 (approved):

vcastellm (approved):

PR #1405, opened 2025-12-16
feat: add `to_address` filed in bridge api response

### LINKED ISSUE #1398: feat: request to bridge service provide `To` field for bridges
# Feature Request

> Could you guys also add the "to" param in the response for the bridge endpoint on aggkit ?
> To refers to the smart contract address on which the specific transaction was executed. It isn't the bridge contract every time. It varies with multiple use cases...  Let us know when you guys can do this. We actually have noticed a bug on the Agglayer UI which can be fixed properly only if we have this "to" address param. Its not a blocker bug and we do have a temporary workaround too, but the right approach would be to get the to param from aggkit itself. Thank you.

https://0xpolygon.slack.com/archives/C08732X070V/p1765532406685169

rachit77:
## 🔄 Changes Summary
- This PR adds `to_address` filed in bridge api response
- No re-syncing is required as existing bridges can return zero address as `to_address`

## 🐞 Issues
- Closes #1398

### DIFF
--- a/bridgesync/downloader_test.go
+++ b/bridgesync/downloader_test.go
@@ -28,7 +28,7 @@ import (
 // mainnet:
 // case https://etherscan.io/tx/0x8db8e288d25102b64d8a37ad05769817d1b43f0384dd05da075d24d2cee9cb65 (bn: 19566985) -> fix
 // case: https://etherscan.io/tx/0x0b276867aa22d1c162c2700d35c500a124a6a953c7b24931a1d3efc63f7cd4ab  (bn: 22770713)
-func TestExtractTxnSenderAndFromExploratory(t *testing.T) {
+func TestExtractTxnAddressesExploratory(t *testing.T) {
 	t.Skip("Skipping exploratory test")
 	ctx := t.Context()
 	l1url := os.Getenv("L1URL")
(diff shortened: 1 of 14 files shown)


bridgesync/downloader_test.go
```diff
@@ -28,7 +28,7 @@ import (
 // mainnet:
 // case https://etherscan.io/tx/0x8db8e288d25102b64d8a37ad05769817d1b43f0384dd05da075d24d2cee9cb65 (bn: 19566985) -> fix
 // case: https://etherscan.io/tx/0x0b276867aa22d1c162c2700d35c500a124a6a953c7b24931a1d3efc63f7cd4ab  (bn: 22770713)
-func TestExtractTxnSenderAndFromExploratory(t *testing.T) {
+func TestExtractTxnAddressesExploratory(t *testing.T) {
```
temaniarpit27, line 31: lets remove this if we are skipping anyway

temaniarpit27 (approved):
pls check once if it works perfectly fine in internal bridges cases

Stefan-Ethernal (approved):

joanestebanr (approved):

rachit77:
> pls check once if it works perfectly fine in internal bridges cases

https://github.com/agglayer/aggkit/issues/1409
created a issue to add internal bridge tx

bridgesync/downloader_test.go
rachit77: i will let @joanestebanr take the decision since he added this test

joanestebanr: That is for debug, when do you need to check it manually you just comment the `skip` and run it. So from my point of view it's good to keep it (I use it very often)

joanestebanr (approved):

PR #1417, opened 2025-12-18
feat: allow to configure the finality to use for l1infotree

### LINKED ISSUE #1410: feat: aggsender: allow to configure the finality to use for l1infotree
# Feature Request
Aggsender must choose the latest l1inforoot based on a new finality param instead of using the reorg finality

joanestebanr:
## 🔄 Changes Summary
- **Aggsender** choose a l1infotree to include the bridges. This l1infotree used to be  the one that belong to more recent block synced by `l1infotreesync` <= finalized block
- This PR allow to configure the target block instead of finalized
- The aggsender target finality must be >= to l1infotree finality, if not it raise an error

## 📋 Config Updates
- 🧾 **Diff/Config snippet**: 
Add next configuration: 
```
[AggSender]
BlockFinalityForL1InfoTree = "FinalizedBlock"
[Validator]
BlockFinalityForL1InfoTree = "{{AggSender.BlockFinalityForL1InfoTree}}"
````

## ✅ Testing
- 🤖 **Automatic**: e2e

## 🐞 Issues
- Closes #1410

### CODE REVIEW

File: config/default.go [hunk 1]
```diff
@@ -215,6 +215,7 @@ MaxL2BlockNumber = 0
 StopOnFinishedSendingAllCertificates = false
 RequireCommitteeMembershipCheck = false
 AgglayerBridgeL2Addr = "{{L2Config.BridgeAddr}}"
+BlockFinalityForL1InfoTree = "FinalizedBlock"
```
goran-ethernal, line 218: Should we set this by default to `L1InfoTreeSync.BlockFinality`?

aggsender/query/l1info_tree_data_query.go
```diff
@@ -18,32 +18,47 @@ var _ types.L1InfoTreeDataQuerier = (*L1InfoTreeDataQuerier)(nil)
...
 		return nil, err
 	}
+	l1InfoTreeFinality := l1InfoTreeSyncer.Finality()
+	lessFinal, err := blockFinalityForL1InfoTree.LessFinalThan(l1InfoTreeFinality)
+	if err != nil {
+		return nil, fmt.Errorf("error comparing block finalities (target: %s and l1infotreeFinality: %s): %w",
+			blockFinalityForL1InfoTree.String(), l1InfoTreeFinality.String(), err)
+	}
+	if lessFinal {
+		return nil, fmt.Errorf("missconfigured block finality (%s) because l1infotreeSyncer has a lower finality (%s) so it "+
```
goran-ethernal: We can probably validate this on the config level. WDYT?

joanestebanr: I also think in that but it's an interaction of components, so if we validate in config level we are creating a behavioural coupling because we assume that l1infotree the way that works the syncer... in that way it's also validated in a more agnostic way (we don't assume how works l1infotreesyncer)

config/default.go
joanestebanr: I think so, by default must be `finalized` that is the current behaviour, in the same way that we are configuring l1infotree finality as `finalized`.

aggsender/query/l1info_tree_data_query.go
goran-ethernal: I was thinking adding this to the `aggsenderConfig.Validate()`, since it is already called before starting the `aggsender` in the `run.go`. I do understand it is coupling, but it is better to return an error on start, instead of keep firing this error non stop while running

config/default.go
goran-ethernal: Yeah, I was thinking like this: `BlockFinalityForL1InfoTree = "{{L1InfoTreeSync.BlockFinality}}"

joanestebanr: sorry, I misunderstood! I think that the l1infotree finality is going to be more close to latest than aggsender becuase it is used by other components also. That is the reason that I introduce a check in the configuration to be sure that the finality of aggsender are included in the one of l1infotreesyncer

aggsender/query/l1info_tree_data_query.go
joanestebanr: I agree, but the validation required that the l1infotreesyncer it's already constructed, so can't be done in `Validate()` and this error it's thrown it constrution stage, so on start.. so I think that there are no problem, isn't?

aggsender/query/l1info_tree_data_query.go
```diff
@@ -18,35 +18,50 @@ var _ types.L1InfoTreeDataQuerier = (*L1InfoTreeDataQuerier)(nil)
...
 	}
+	l1InfoTreeFinality := l1InfoTreeSyncer.Finality()
+	lessFinal, err := blockFinalityForL1InfoTree.LessFinalThan(l1InfoTreeFinality)
+	if err != nil {
+		return nil, fmt.Errorf("error comparing block finalities (target: %s and l1infotreeFinality: %s): %w",
+			blockFinalityForL1InfoTree.String(), l1InfoTreeFinality.String(), err)
+	}
+	if lessFinal {
+		return nil, fmt.Errorf("misconfigured block finality (%s) because l1infotreeSyncer has a lower finality (%s) so it "+
+			"never going to be fulfilled",
```
vcastellm: Check grammar

aggsender/query/l1info_tree_data_query.go
goran-ethernal: It should be constructed prior to aggsender, isn't it? As far as I see, we first run the `l1InfoTreeSyncer` if needed, and then we create the `major` components like `aggsender`. So I think it should be fine to validate in `run.go` on `createAggsender` and `createAggsenderValidator`. This would make code simpler. We wouldn't need to expand the syncer with `Finality()` function and no need for this check in `aggsender`.

joanestebanr: Yes, it's an option. Do you mean to pass the l1infotree configuration to  `createAggsender`  to be able to do the check of finalities? or use the `l1infotreesyncer.Finality()` and move the check from `NewL1InfoTreeDataQuerier` to `run.go`?

goran-ethernal: It was a misunderstanding. I thought this error was being returned in the `getTargetL1BlockNumber` function, but it is being returned in the constructor of `l1infotreequerier` which is an appropriate place for checking this. I will resolve the comment.

goran-ethernal (approved):

aggsender/query/l1info_tree_data_query.go
joanestebanr: changed to: `"block finality misconfiguration (%s): l1infotreeSyncer finality (%s) is lower; will never be fulfilled"`

a6b6f53736ccbdea6a1238e3d2ed6206b90c9ac0

rachit77 (approved):

vcastellm (approved):
LGTM

PR #1425, opened 2025-12-22
fix: check InError certs when no pending certs

vcastellm:
## 🔄 Changes Summary

Query storage for InError certificates if no pending certs and no transition detected, enabling retries and logging storage errors. Update tests to expect the additional storage query.

The fix is minimal, backward-compatible, and only adds an additional check when needed, ensuring the retry logic is more robust against transient failures.


goran-ethernal (approved):
LGTM. I would maybe add another test for this to check if the `ExistsNewInErrorCert` is as expected in these cases.

Stefan-Ethernal (approved):

Stefan-Ethernal:
@vcastellm Note that the `TestCheckPeriodicallyStatus` is failing due to these changes.

PR #1453, opened 2026-01-12
fix: migration file

rachit77:
## 🔄 Changes Summary
- Set default value for `to_address` field

## 🔗 Related PRs
- https://github.com/agglayer/aggkit/pull/1452


goran-ethernal (approved):
I think it is good enough. Although, on migrate down, I would still add the update query to update this column to an empty string just in case if someone want's to run that version, even though the chances for that are small.

rachit77:
> I think it is good enough. Although, on migrate down, I would still add the update query to update this column to an empty string just in case if someone want's to run that version, even though the chances for that are small.

this previous version of `to_address` is used by nobody, this feature will only go live with v0.8 release

Stefan-Ethernal (approved):

temaniarpit27 (approved):

PR #1457, opened 2026-01-13
feat: aggsender: send a new certificate as soon latest is settled

### LINKED ISSUE #1411: feat: aggsender: send a new certificate as soon latest is settled
# Feature Request
Next version of aggleyer doesn't require to submitted 1 cerfiticate per epoch. So the idea is, as soon as,  latest submitted certificate is `settled` send a new one

joanestebanr:
## 🔄 Changes Summary
- It allow to configure the trigger mode for certificates
- It introduce a new trigger that is *as soon as possible*: so after a succesful settle of a certificate try to generate a new one
- Reduce periodically logs from ` statuschecker/` that now are `debug` instead of `info`
- Fix local run scripts
- Add a new subscription to bridgesyncer for new bridges
- Update go-ethereum to v1.16.8  because `go-ethereum v1.16.7  abi vulnerability CVE 2026`

## ⚠️ Breaking Changes
- 🛠️ **Config**: Moved field `AggSender.EpochNotificationPercentage` to `AggSender.TriggerEpochBased`
   This field is no longer mandatory and it depens on the trigger mode choosen.

## 📋 Config Updates
New field `AggSender.TriggerCertMode` you can choose the trigger of new certificate: 
- `EpochBased`: this is the legacy mode that waits until reach a percentage of a epoch (you can configure here: `AggSender.TriggerEpochBased`)
- `ASAP`: this mode try to generate a new certificate after a sucessful settled certificate
- `NewBridge`: Each time that a new bridge is done in L2 it generate a certificate (if it's possible) (**experimental**)

#### Configuration for TriggerASAP
This trigger can be configured in `AggSender.TriggerASAP`: 
* **DelayBeetweenCertificates**: Delay between the moment the aggsender becomes idle and when it sends a new certificate trigger
* **MinimumNewCertificateInterval**: Minimum time that must elapse between certificate generation triggers, regardless of the trigger source
* **OnNewL2Bridge**: When enabled, the ASAP trigger will automatically generate certificate triggers when new bridge events are detected on L2

```
[AggSender]
TriggerCertMode = "Auto"
[AggSender.TriggerEpochBased]
	# Percentage of epoch completion to trigger certificate sending
	EpochNotificationPercentage = 50
[AggSender.TriggerASAP]
	DelayBeetweenCertificates = "1s"
	MinimumNewCertificateInterval = "5m"
	OnNewL2Bridge = false
```
## ✅ Testing
- 🤖 **Automatic**: unittest and e2e

## 🐞 Issues
- Closes #1411


arnaubennassar (approved):

aggsender/config/config.go
```diff
@@ -16,6 +17,49 @@ import (
 	ethCommon "github.com/ethereum/go-ethereum/common"
 )
 
+type TriggerASAPConfig struct {
+	// DelayBeetweenCertificates is the delay to wait before sending a new certificate after the previous one is settled
+	DelayBeetweenCertificates types.Duration `mapstructure:"DelayBeetweenCertificates"`
+	// MinimumNewCertificateInterval is the minimum interval between two new certificates triggers
+	MinimumNewCertificateInterval types.Duration `mapstructure:"MinimumNewCertificateInterval"`
```
goran-ethernal, line 24: Hmm, why do we need this config parameter?

aggsender/types/interfaces.go
```diff
@@ -381,11 +382,43 @@ type CertificateTriggerEvent interface {
 	fmt.Stringer
 }
 
+type CertificateSendTriggerMode string
```
goran-ethernal, line 386: Should these go to `types.go`?

aggsender/aggsender.go
```diff
@@ -341,10 +342,16 @@ func (a *AggSender) sendCertificates(ctx context.Context, returnAfterNIterations
 				a.log.Warnf("reached number of iterations, so we are going to return")
 				return
 			}
+
+			if !checkResult.ExistPendingCerts && !checkResult.ExistNewInErrorCert {
+				a.log.Debugf("No pending or InError certificates found, so aggsender is waiting for trigger")
+				a.certificateSendTrigger.OnAggsenderWaitingTrigger()
```
goran-ethernal: Hmm, not sure why do we need this. If we do have a periodic check configured, and we don't have any certificates to check, then I would just leave it running without this complex logic about waiting for the trigger for the next certificate. It seems unnecessary, and it brings more complexity, and even though this check is running on no certificates, I don't think it affects anything.

aggsender/config/config.go
joanestebanr: I introduce it for 2 reasons: 
- The main: if the config parameter `CheckStatusCertificateInterval` is 0 then there is no certificate check before the trigger, so it require a way to 'force' a trigger
- If the mechanism doesn't work as expected we have a save guard that is going to generate a certificate if you wait

aggsender/aggsender.go
joanestebanr: This condition is basically that, if there are no certificate and there are no `InError` certificate trigger a new certificate. This code is from the periodic check and it's the condition for a trigger. How do you suggest to reduce the complexity?

goran-ethernal: Hmm, so this is basically triggering the new certificate build when we do not have pending and in error certificates. Am i right? I might have misunderstood the code. In that case I wouldn't call this function `OnAggsenderWaitingTrigger` because it paints a picture like you are waiting here for something. Maybe something like `TriggerManually`.

joanestebanr: Yes, exactly... maybe make sense to call ir OnIdle() in the sense that  `aggsender` is not going to do anything else by his own. WDYT?

goran-ethernal: Or `TriggerOnIdle()` 😛

joanestebanr: But I think that Trigger is going to be too many `trigger`words: 
1) `a.certificateSendTrigger.TriggerOnIdle()`
vs
2) `a.certificateSendTrigger.OnIdle()`
I think that option 2 is less verbose, WDTY?

goran-ethernal (approved):

aggsender/aggsender.go
joanestebanr: 2dfda598657a154ffc6cf9e70dc4f0fa58dc3b21

PR #1496, opened 2026-02-24
feat: test remove GER tool

arnaubennassar:
🔄 Changes Summary

  - Added remove-ger CLI tool (tools/remove_ger/) to diagnose and recover from invalid Global Exit Root (GER) injection on L2. The tool classifies affected claims into categories (A, B.1, B.2) and runs interactive recovery steps (freeze bridge, remove GER, unset/re-set or force-emit claims, restore bridge).
  - Added three new bridgesync processor methods (GetClaimsByGER, GetBridgeByDepositCount, GetBridgesByContent) and corresponding bridgeservice REST API endpoints (/claims-by-ger, /bridge-by-deposit-count, /bridges-by-content) consumed by the tool.
  - Fixed a bug in bridgesync/downloader.go where ClaimEvent and DetailedClaimEvent were both persisted for the same transaction on the first block that emits both, causing duplicate claim records that broke aggsender unclaim filtering.
  - Added a comprehensive operator runbook (docs/remove_ger_runbook.md) covering detection, manual recovery steps, and tool usage.
  - Added bridgeservice client methods and updated Swagger/OpenAPI docs.
  - Added CI fix for govulncheck Python filter handling non-dict JSON values.

  ⚠️  Breaking Changes

  None.

  📋 Config Updates

  The remove-ger tool reads the standard aggkit-config.toml (using the same template rendering pipeline as the main binary) and requires an additional section:

  [RemoveGER]
    BridgeServiceURL = "http://localhost:8080"
    L2NetworkID = 1

    [RemoveGER.SovereignAdminKey]
      # Supports local keystore, AWS KMS, GCP KMS via signertypes.SignerConfig
      KeystoreFileLocation = "/path/to/keystore.json"
      KeystorePassword = "..."

  ✅ Testing

  - Automatic (E2E): Four E2E test scenarios against a dockerized OP-PP stack:
    - TestRemoveGER_NoProblematicClaims — injects fake GER with no associated claims
    - TestRemoveGER_CategoryA — fake GER with a dummy claim referencing no real bridge
    - TestRemoveGER_CategoryB1 — real bridge + claim with an invalid GER at the correct deposit_count
    - TestRemoveGER_CategoryB2 — real bridge + claim at wrong deposit_count (fake GER)
  - Automatic (unit): Unit tests added for all new bridgesync processor methods and bridgeservice HTTP handlers; coverage for new bridgeservice/client methods.
  - Manual:
    a. Build the tool: make build-tools
    b. Point --cfg at a live aggkit config with a [RemoveGER] section
    c. Run remove_ger --cfg aggkit-config.toml --ger <0x...> and follow interactive prompts
    d. Verify aggsender resumes normal certificate submission after recovery

  🐞 Issues

  Closes https://github.com/agglayer/pm/issues/253

  🔗 Related PRs

  - #1496 (prior remove-GER work that this PR builds on / cleans up)

  📝 Notes

  - The tool is intentionally interactive: it prints a diagnosis summary and asks for confirmation before executing any on-chain transactions.
  - SignerConfig from go_signer is used for the sovereign admin key, supporting local keystore, AWS KMS, and GCP KMS backends.
  - The duplicate-claim bug fix (bridgesync/downloader.go) is a correctness fix independent of the tool; it affects all sovereign chains where both ClaimEvent and DetailedClaimEvent are emitted.
  - SQL queries in bridgesync/processor.go are pre-built as compile-time constants to avoid dynamic SQL construction.


tools/remove_ger/config.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// RemoveGERConfig contains configuration specific to the remove-GER tool.
+type RemoveGERConfig struct {
+	// SovereignAdminPrivateKey is the private key with privileges to:
+	// - activateEmergencyState / deactivateEmergencyState on the L2 bridge
+	// - removeGlobalExitRoots on the L2 GER manager
+	// - unsetMultipleClaims / setMultipleClaims on the L2 bridge
+	// - forceEmitDetailedClaimEvent on the L2 bridge
+	SovereignAdminPrivateKey KeyConfig `mapstructure:"SovereignAdminPrivateKey"`
```
joanestebanr: I suggest to use the way to configure keys as the rest of aggkit:  `signertypes.SignerConfig`  (`signertypes "github.com/agglayer/go_signer/signer/types"`) this allows to store the key in AWS, ...

tools/remove_ger/cmd/main.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/agglayer/aggkit/tools/remove_ger"
+	"github.com/urfave/cli/v2"
+)
+
+func main() {
+	app := cli.NewApp()
+	app.Name = "remove-ger"
+	app.Usage = "Diagnose and recover from invalid GER injection on L2"
+	app.Flags = []cli.Flag{
```
joanestebanr, line 17: Maybe a `version` flag could be useful, WDYT?

tools/remove_ger/config.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+
+	"github.com/agglayer/aggkit/config"
+	"github.com/spf13/viper"
+	"github.com/urfave/cli/v2"
+)
+
+// Config extends the main aggkit config with fields specific to the remove-GER tool.
+type Config struct {
+	config.Config
```
joanestebanr: Instead of importing all aggkit config you can add specifically what you need. e.g:
```
type Config struct {
   L1NetworkConfig ethermanconfig.L1NetworkConfig
   Common ethermanconfig.CommonConfig
   RemoveGER RemoveGERConfig
}
```
WDYT?

bridgesync/downloader.go
```diff
@@ -475,6 +475,17 @@ func buildClaimEventHandler(ctx context.Context, agglayerBridge *agglayerbridge.
 			}
 		}
 
+		// Skip if a DetailedClaimEvent for the same transaction is already in this block's events.
```
joanestebanr: I suggest to move this block before calling `extractCallData` to save this calls. 
Also I see that exists a mechanism to avoid processing this event if `detailedClaimEvent` is available (check line 431: `boundaryBlock, err := querier.GetBoundaryBlockForClaimType(ctx, DetailedClaimEvent)`)

bridgeservice/bridge.go
```diff
@@ -210,6 +212,9 @@ func (b *BridgeService) registerRoutes() {
 		bridgeGroup.GET("/last-reorg-event", b.GetLastReorgEventHandler)
 		bridgeGroup.GET("/sync-status", b.GetSyncStatusHandler)
 		bridgeGroup.GET("/removed-gers", b.GetRemoveGEREventsHandler)
+		bridgeGroup.GET("/claims-by-ger", b.GetClaimsByGERHandler)
```
joanestebanr, line 215: swagger file must be updated with new end-points

bridgesync/downloader.go
joanestebanr: The old mechanism `GetBoundaryBlockForClaimType` will be fixed in another PR

tools/remove_ger/cmd/main.go
joanestebanr: The version introduced it's not working: always v0.1.0:
- In the build commandline need to pass the version as aggkit do: `-ldflags "all=$(LDFLAGS)"`

joanestebanr (approved):
The version is not working yet

PR #1502, opened 2026-03-02
feat: backward and forward LET testing + runbook + tool 

arnaubennassar:
🔄 Changes Summary
                                                                                                                                         
  - New backward_forward_let recovery tool (`tools/backward_forward_let`): CLI tool for diagnosis and recovery of rollup state after Agglayer and/or L1 become inconsistent with L2 state. Comes with full runbook documentation and E2E tests
  - Fix `bridgesync/processor.go` BackwardLET/ForwardLET bugs that caused the L2BridgeSyncer to halt or produce duplicate bridges during recovery
  - Rename `GetLastLocalExitRoot → GetInitialLocalExitRoot`
  - `bridgesync/processor.go` uses `GetInitialLocalExitRoot` insted of `EmptyLER` to avoid bugs for networks that have a non default value
  - EmptyLER moved from aggsender/types to bridgesync/types
  - New `aggsender_getCertificateBridgeExits RPC` on the aggsender JSON-RPC server, with matching client method: returns bridge exits for a given certificate height
  - flow_base.go: for settled certs, re-derives fromBlock using on-chain data via certQuerier instead of trusting the stored ToBlock, preventing divergence with the validator and accepting Agglayer as source of truth

  ⚠️  Breaking Changes

None

  📋 Config Updates

 None

  ✅ Testing

  - 🤖 Automatic:
    - `test/e2e/backwardforwardlet_test.go` fully test E2E the new feature
    - Unit tests for all new tool packages: diagnosis_test.go, recovery_test.go, helpers_test.go, override_test.go, send_cert_test.go
    - Updated unit tests in bridgesync/processor_test.go, aggsender/rpc/, aggsender/rpcclient/

  🐞 Issues

  - Closes https://github.com/agglayer/pm/issues/282


.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
+{
```
joanestebanr: I think that this file is local, must not be part of git (in my case I have my own configurations there that will be overwritten)

aggsender/config/config.go
```diff
@@ -150,6 +150,13 @@ type Config struct {
 	TriggerEpochBased TriggerEpochBasedConfig `mapstructure:"TriggerEpochBased"`
 	// TriggerASAP is the configuration for the ASAP trigger mode (TriggerCertMode==ASAP)
 	TriggerASAP TriggerASAPConfig `mapstructure:"TriggerASAP"`
+	// EnableDebugSendCertificate enables the debug RPC endpoint for sending arbitrary certificates.
```
joanestebanr: missing values in `config/default.go`

aggsender/rpc/aggsender_rpc.go
```diff
@@ -83,3 +108,95 @@ func (b *AggsenderRPC) GetCertificateHeaderPerHeight(height *uint64) (interface{
...
+	if exits == nil {
+		return nil, rpc.NewRPCError(rpc.NotFoundErrorCode,
+			fmt.Sprintf("certificate not found at height %d", resolvedHeight))
+	}
+	return exits, nil
+}
+
+// DebugSendCertificate sends an arbitrary certificate to AggLayer (test-only endpoint).
+// Requires EnableDebugSendCertificate=true in config and a valid Ethereum signature.
+func (b *AggsenderRPC) DebugSendCertificate(signedRequest DebugSendCertificateRequest) (interface{}, rpc.Error) {
```
joanestebanr: why do you need this end-point? you can:
- stop aggsender
- send directly the certificate to agglayer that is already signed with the privatekey of aggsender
- start aggsender that is going to ask for last cert and store in DB

aggsender/rpc/aggsender_rpc.go
```diff
@@ -83,3 +108,95 @@ func (b *AggsenderRPC) GetCertificateHeaderPerHeight(height *uint64) (interface{
...
+		b.logger.Warnf("debug: failed to marshal certificate for storage: %v", err)
+	} else {
+		jsonCertStr := string(jsonCert)
+		now := uint32(time.Now().Unix())
+		cert := types.Certificate{
+			Header: &types.CertificateHeader{
+				Height:           signedRequest.Certificate.Height,
+				CertificateID:    signedRequest.Certificate.CertificateID(),
+				NewLocalExitRoot: signedRequest.Certificate.NewLocalExitRoot,
+				Status:           agglayertypes.Pending,
```
joanestebanr: Several fields of the struct are empty: 
- PreviousLocalExitRoot
- FinalizedL1InfoTreeRoot
- L1InfoTreeLeafCount
- CertType
- BlockRange

aggsender/rpc/aggsender_rpc.go
```diff
@@ -83,3 +108,95 @@ func (b *AggsenderRPC) GetCertificateHeaderPerHeight(height *uint64) (interface{
...
+	jsonCert, err := json.Marshal(&signedRequest.Certificate)
+	if err != nil {
+		b.logger.Warnf("debug: failed to marshal certificate for storage: %v", err)
+	} else {
+		jsonCertStr := string(jsonCert)
+		now := uint32(time.Now().Unix())
+		cert := types.Certificate{
+			Header: &types.CertificateHeader{
+				Height:           signedRequest.Certificate.Height,
+				CertificateID:    signedRequest.Certificate.CertificateID(),
```
joanestebanr: CertifacateID must be the response of agglayer

aggsender/rpc/aggsender_rpc.go
arnaubennassar: If I do it this way, the DB is going to be missing bridge info. The idea is to simulate a scenario in which an aggsender sends a certificate that:
a) later on gets reorged (so we miss the bridges on the L2 and the bridgesyncdb)
b) the aggsender just alucinates and sends some incorrect values

.vscode/launch.json
arnaubennassar: I think it's very convinent to have this on git. This way you can just clone the repo and start debugging. If u have something useful I'd encourage you to share it

bridgesync/processor.go
```diff
@@ -1622,11 +1625,7 @@ func (p *processor) ProcessBlock(ctx context.Context, block sync.Block) error {
 		return sync.ErrInconsistentState
 	}
 
-	// Create a context with database timeout for the transaction
```
joanestebanr, line 1625: may be make sense to setup a timeout for process a block... WDYT?
If you remove from here also make sense to remove it from Reorg function

bridgesync/processor.go
```diff
@@ -1734,19 +1733,31 @@ func (p *processor) ProcessBlock(ctx context.Context, block sync.Block) error {
 				return err
 			}
 
-			// 1. archive and remove all the bridges whose
-			// deposit_count is greater than the one captured by the BackwardLET event
+			// 1. archive and remove all bridges at deposit_count >= newDepositCount.
```
joanestebanr: May be make sense to move all this if  to a function like `insertBackwardLET` because it's hard to see the logic in this huge function. WDTY?

bridgesync/processor.go
```diff
@@ -1877,6 +1901,15 @@ func (p *processor) sanityCheckLatestLER(tx dbtypes.Txer, ler common.Hash) error
 		lastRootHash = root.Hash
 	}
 
+	if ler == bridgesynctypes.EmptyLER {
```
joanestebanr: Why it need this case? the code do the same as next if `if lastRootHash != ler`... is to be able to emit a specific log? 
```
if lastRootHash != aggkitcommon.ZeroHash {
```
It's the same as
```
if lastRootHash != ler {
```

bridgesync/processor.go
```diff
@@ -1937,20 +1976,26 @@ func (p *processor) handleForwardLETEvent(tx dbtypes.Txer, event *ForwardLET, bl
...
+		switch len(archivedBridges) {
+		case 1:
 			archivedBridge := archivedBridges[0]
 			txnHash = archivedBridge.TxHash
 			txnSender = archivedBridge.TxnSender
 			// It copies the fromAddr pointer, which could be nil
 			fromAddrPtr = archivedBridge.FromAddress
-		} else if len(archivedBridges) > 1 {
-			p.log.Warnf("multiple archived bridges found that match forward LET leaf %s;"+
+		case 0:
```
joanestebanr, line 2007: We need to evaluate what happens with backfilling process in that case that txnSender and fromAddr are not set

bridgesync/processor.go
arnaubennassar: the problem with this timeout is that it makes it really hard to debug (when using the debugger, timeout is triggered if you add breakpoints)

bridgesync/processor.go
arnaubennassar: because of sanity check of previous root. If the DB is empty the root is returned as `0x000...0` while `EmptyLER` has a different value (the thing returned by the SC)

.vscode/launch.json
joanestebanr: I think it's a matter of taste: I prefer to run direclty in VScode the instace because it allow me to modify faster the code instead of connecting to the docker that for each change I must rebuild it.

By other hand, I used to have a lot of entries there to run against mainnet, testnets... with several config files... 

Maybe you can include this a README.md

bridgesync/processor.go
joanestebanr: I suggest that you can use ,for debug,  a config file with a really big value for the timeout instead of removing it

aggsender/rpc/aggsender_rpc.go
joanestebanr: But then implies that we need to backup aggsender DB because is no longer disposable

bridgesync/processor.go
joanestebanr: I think you have to consider use LERQuerier.GetLastLocalExitRoot()  to get the emptyLER

aggsender/db/aggsender_db_storage.go
```diff
@@ -286,6 +288,23 @@ func (a *AggSenderSQLStorage) GetLastSettledCertificate() (*types.CertificateHea
 	return &certificateHeader, nil
 }
 
+// GetCertificateBridgeExits returns the bridge exits for the signed certificate at the given height.
+// Returns nil if no certificate exists at that height or the certificate has no signed certificate data.
+func (a *AggSenderSQLStorage) GetCertificateBridgeExits(height uint64) ([]*agglayertypes.BridgeExit, error) {
```
joanestebanr: Make sense to move that to RPC, because storage don't known the content of SignedCertificate

aggsender/flows/flow_base.go
```diff
@@ -116,6 +119,21 @@ func (f *baseFlow) NextCertificateBlockRange(ctx context.Context,
...
+	// fromBlock even when the stored ToBlock is stale (e.g. set to 0 by the debug endpoint).
+	if lastSentCertificate != nil && lastSentCertificate.Status.IsSettled() && f.certQuerier != nil {
+		agglayerCert := converters.ConvertAggsenderCertHeaderToAgglayer(
+			lastSentCertificate, f.l2BridgeQuerier.OriginNetwork())
+		derivedToBlock, err := f.certQuerier.GetLastSettledCertificateToBlock(ctx, agglayerCert)
+		if err != nil {
+			return aggkitcommon.BlockRangeZero, 0,
+				fmt.Errorf("error deriving toBlock for settled cert %s: %w", lastSentCertificate.ID(), err)
+		}
+		previousToBlock = derivedToBlock
```
joanestebanr, line 139: I suggest to add a log.Infof if this values are not equal

joanestebanr (commented):
- Review SonarCloud issue
- Update PR description: the description is pointing to the previous implementation

joanestebanr (approved):

PR #1569, opened 2026-03-31
fix: unstack aggsender unset claims

arnaubennassar:
## 🔄 Changes Summary
  - Fix aggsender recovery when invalid claims and their matching unclaims are both present in
  the same candidate certificate but the unclaims happen in a different order.
  - Add operator-facing diagnostics for blocking invalid claims, including the claim block,
  global index, token, and amount.
  - Warn when `MaxCertSize` or `MaxL2BlockNumber` would prevent the matching unclaim from
  fitting into the same certificate, and suggest the config change needed to unblock aggsender.

  ## ⚠️ Breaking Changes
None

  ## 📋 Config Updates
None

  ## ✅ Testing
  - 🤖 **Automatic**:
    - `go test ./aggsender/flows`
  - 🖱️ **Manual**:
    - Reproduced the network-12 failure pattern in tests using the real claim/unclaim block
  ordering.
    - Verified aggsender no longer cuts the certificate when all invalid claims are also unset
  within the same candidate certificate.
    - Verified new logs for:
      - missing matching unclaim
      - matching unclaim present, aggsender can proceed
      - limiter prevents the unclaim from fitting into the certificate

  ## 🐞 Issues
  - Closes #N/A

  ## 🔗 Related PRs
  - `feat/fix-removeger-tool-docs`: hardening in `tools/remove_ger` used to clean the invalid
  GER incident and confirm the network state before fixing aggsender recovery behavior.

  ## 📝 Notes
  - RCA: aggsender could remain stuck even after the network was cleaned because the previous
  logic required later invalid claims to be unclaimed no later than earlier ones. That was
  stricter than the final imported-bridge-exit construction, which only requires claim/unclaim
  pairs to cancel out within the same certificate.
  - This PR keeps the core safety requirement intact: an invalid claim is only considered
  recoverable when its matching unclaim is also inside the same candidate certificate.


aggsender/flows/flow_base.go
```diff
@@ -656,19 +662,26 @@ func (f *baseFlow) adjustCertificateIfNonFinalizedClaims(
...
 	}
-	if cutBlock != 0 {
-		newToBlock := cutBlock - 1
+	if assessment != nil && assessment.reason == invalidClaimAssessmentReasonRecoverable {
+		f.log.Infof(
+			"found invalid claim with matching unclaim in current cert, aggsender can proceed. "+
+				"%s, unclaim_block=%d, cert_range=%d-%d",
+			formatClaimForLogs(*assessment.culpritClaim), assessment.culpritUnclaim, certParams.FromBlock, certParams.ToBlock)
+	}
+	if assessment != nil && assessment.cutBlock != 0 {
```
joanestebanr, line 675: Reduce complexity adding an no action required if: 
```
if assessment == nil {
  return certParams, nil
}
```

aggsender/flows/flow_base.go
```diff
@@ -239,11 +239,17 @@ func (f *baseFlow) VerifyBuildParams(ctx context.Context, fullCert *types.Certif
 // size is expressed in bytes
 func (f *baseFlow) LimitCertSize(
 	certParams *types.CertificateBuildParams) (*types.CertificateBuildParams, error) {
+	originalCert := certParams
 	currentCert := certParams
 	var err error
 	maxCertSize := f.cfg.MaxCertSize
 	for {
 		if maxCertSize == 0 || currentCert.EstimatedSize() <= maxCertSize {
+			if originalCert != nil && currentCert != nil && currentCert.ToBlock < originalCert.ToBlock {
```
joanestebanr, line 248: originalCert must be always != nil, same for currentCert

joanestebanr:
Close in favour https://github.com/agglayer/aggkit/pull/1573

PR #1574, opened 2026-04-10
fix: determine claim syncer starting block from earliest settled reference when DB is empty

joanestebanr:
## 🔄 Changes Summary
- When the claim syncer DB is empty on startup, the starting block is now determined using the **earliest** block among the three settled references (bridge exit block, imported bridge exit block, FEP L2 block) instead of the latest.
- Added `EarliestBlock()` method to `SettledBlocks` that excludes `LastSettledL2BlockNum` when it is 0 (not found).
- Added `GetBlockNumbersFromCertHeader()` to `certificateQuerier` to query all three block sources independently, recording errors per source without aborting the others.
- When the imported bridge exit block cannot be resolved from local storage, falls back to RPC log scanning (`GetLatestBlockNumByGlobalIndexFromRPC`) using chunked `FilterLogs` backwards from the finalized block.
- Early-exit in `SetClaimSyncerNextRequiredBlock` when the claim syncer DB already has processed blocks (idempotent startup).
- Removed `ClaimQuerier` interface and `GetBoundaryBlockForClaimType` method from storage, processor, and all related interfaces/mocks. Deduplication between `ClaimEvent` and `DetailedClaimEvent` is now handled entirely by the `DetailedClaimEvent` handler in-block, removing the need for DB lookups during event processing.

## ⚠️ Breaking Changes
None

## 📋 Config Updates
None

## ✅ Testing
  - 🤖 **Automatic**:
    - `make test-unit`
  - 🖱️ **Manual**:
    - Deploy with empty claim syncer DB against a network that has settled certificates with imported bridge exits and verify the syncer starts from the correct earliest block.

## 🐞 Issues
  - Closes #N/A

## 🔗 Related PRs
  None

## 📝 Notes
`GetLatestBlockNumByGlobalIndexFromRPC` auto-detects the RPC max-range limit from the first error message and retries in backward chunks until a matching claim log is found or the full range is exhausted.


AGENTS.md
```diff
@@ -26,3 +26,31 @@
 - Each component has: config, types, mocks subdirectories
 - Tests use testify; files named `*_test.go`. Use `require` (not `assert`) when creating unit tests
 - Database migrations in `migrations/` subdirectories
+
+
+## GitHub
+
+- To create a new PR use the following template:
+```
+## 🔄 Changes Summary
```
arnaubennassar: reference the template file instead of duplicating

aggsender/query/certificate_query.go
```diff
@@ -41,6 +41,50 @@ func NewCertificateQuerier(
 	}
 }
 
+// GetBlockNumbersFromCertHeader queries each of the three block number sources independently
+// using the provided certificate header.
+// A failure in one source is recorded in the corresponding error field without
+// stopping the other queries.
+func (c *certificateQuerier) GetBlockNumbersFromCertHeader(ctx context.Context,
```
arnaubennassar: by reading the name of the func and the comment, is not clear what this function does. The return type suggests "SettledBlocks" not just "BlockNumbers". Also, what are the "three block number sources"

EDIT: after going further in the review and looking at `types.SettledBlocks` this is easier to understand. I'd sugges renaming the func to `GetSettledBlocksFromCertHeader`

arnaubennassar (approved):

AGENTS.md
joanestebanr: `AGENTS.md` now references `.github/PULL_REQUEST_TEMPLATE.md` instead of duplicating the template content.

PR #1622, opened 2026-05-21
fix(exit-certificate): F-01 implement ERC-20 balance storage patching for SC-locked exits

joanestebanr:
### Fix F-01: `ensureERC20Balance` storage patching for SC-locked ERC-20 exits
- Implements `ensureERC20Balance` in Step G, which was a stub that always returned an error without actually patching Anvil storage.
- The function now patches `_balances[account]` via `hardhat_setStorageAt` using a two-layout detection strategy, verifying `balanceOf` after each attempt:
  1. **OZ v4 non-upgradeable**: `_balances` mapping at storage slot 0
  2. **OZ v5 upgradeable**: `_balances` inside the namespaced `ERC20Storage` struct at `ERC20StorageLocation = 0x52c63247e1f47db19d5ce0460030c497f067ca4cebf71ba98eeadabe20bace00`
- Adds a package-level `erc20NamespacedStorageLocation` constant documenting the OZ v5 storage namespace derivation.

### Refactor: remove `lbtFile` config option
- `lbtFile` was an escape hatch to skip Step 0 by providing a pre-generated LBT. Step 0 now always runs, so the field is removed.
- Merges `RunStepCWithEntries` back into `RunStepC` (simpler API, no config dependency).
- Updates `resolveOrGenerateLBT`, `loadWrappedTokensFromLBT`, `runSingleC`, `runSingleB`, `runSingleF`, `runSingleG` accordingly.

### Kurtosis script improvements
- Adds two helper scripts for the Kurtosis test environment.
- Embeds a pre-generated exit address keypair (`0xe25f5B65E4976025f670e52b790a9746F27A3DB6`) in `configuration_based_on_kurtosis.sh` so the exit address is stable across runs without requiring Foundry at script runtime. The private key and an encrypted keystore (password: `test`) are written to `tmp/` on first execution.

## ⚠️ Breaking Changes
- `lbtFile` config field removed. Any existing config files using it will have the field silently ignored (unknown JSON fields are not errors, but Step 0 will now always run).

## 📋 Config Updates
- `lbtFile` removed from `Config` and `rawConfig`. Step 0 is no longer skippable via config.

## ✅ Testing
- 🤖 **Automatic**: Existing unit tests pass (`ok github.com/agglayer/aggkit/tools/exit_certificate`)
- 🖱️ **Manual**:
  1. Run `tools/exit_certificate/scripts/reproduce_sc_locked.sh` against a live Kurtosis `aggkit` enclave
  2. The script deploys a `TokenHolder` smart contract on L2, transfers wrapped ERC-20 tokens to it, then drives the exit-certificate tool from steps 0→G
  3. Before fix: Step G failed with `ERC20InsufficientBalance` (`0xe450d38c`) when replaying the SC-locked `BridgeExit`
  4. After fix: Step G completes and emits a valid `NewLocalExitRoot`

## 🐞 Issues
- Fixes: #1624 

## 🔗 Related PRs
- N/A

## 📝 Notes
- `TokenWrapped` (wTTK) deployed by `AgglayerBridge` uses OZ v5 upgradeable storage, so the slot-0 attempt is a no-op. The second candidate (namespaced storage) is the one that matches and patches the balance correctly.
- The two-layout approach is safe: a failed slot write leaves the token balance unchanged and the loop moves to the next candidate. If neither layout works the function returns a descriptive error.

### DIFF
--- a/tools/exit_certificate/CLAUDE.md
+++ b/tools/exit_certificate/CLAUDE.md
@@ -62,7 +62,7 @@ All checks run regardless of individual failures. A combined error lists every f
 
 ### Step 0 — Generate LBT
 
-- **Trigger:** runs unless `lbtFile` is set and the file exists.
+- **Trigger:** always runs as part of the full pipeline.
 - **Does:** scans L2 bridge `NewWrappedToken` events, fetches `totalSupply` per token at `targetBlock`, computes unlocked native balance.
 - **Output:** `step-0-lbt.json` (`[]LBTEntry`)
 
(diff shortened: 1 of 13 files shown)


web3security (approved):
F01 fix LGTM

PR #1631, opened 2026-05-28
feat: split long running UTs

### LINKED ISSUE #321: [FEATURE] Split unittest between short and log
There are merged unittest that are 'pure' unittest and another that are integration tests (takes, longer or depend on dockers, external services).
The goal is to have a Makefile target to run the shorter ones. This will allow as to integrate un visual studio to run it each time we save a file and detect broken test / coverage is a more agile way

arnaubennassar:
## 🔄 Changes Summary
- Split unit tests that are "pure" vs the ones that have deps that make them slow

## ⚠️ Breaking Changes
None

## 📋 Config Updates
None

## ✅ Testing
tested that things keep working... UTs still take 20s to run

## 🐞 Issues
- Closes #321

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,12 +102,8 @@ build-docker-debug: ## Builds a debug docker image (dlv headless on :40000, no o
 	docker build -t aggkit:local-debug -f ./Dockerfile.debug .
 
 .PHONY: test-unit
-test-unit: ## Runs the unit tests
+test-unit: ## Runs the short unit tests
 	trap '$(STOP)' EXIT; MallocNanoZone=0 go test -count=1 -short -race -p 1 -covermode=atomic -coverprofile=coverage.out -timeout 15m ./...
-
-.PHONY: test-unit-short
-test-unit-short: ## Runs only the fast unit tests (skips integration tests). Suitable for editor on-save runs.
-	go test -count=1 -short -timeout 2m ./...
 
 .PHONY: test-e2e
 test-e2e: ## Runs the e2e tests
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -105,6 +105,10 @@ build-docker-debug: ## Builds a debug docker image (dlv headless on :40000, no o
 test-unit: ## Runs the unit tests
 	trap '$(STOP)' EXIT; MallocNanoZone=0 go test -count=1 -short -race -p 1 -covermode=atomic -coverprofile=coverage.out -timeout 15m ./...
```
joanestebanr, line 106: Makefile target `test-unit` also runs the short version of tests (`-short`)

Makefile
```diff
@@ -105,6 +105,10 @@ build-docker-debug: ## Builds a debug docker image (dlv headless on :40000, no o
...
 
+.PHONY: test-unit-short
+test-unit-short: ## Runs only the fast unit tests (skips integration tests). Suitable for editor on-save runs.
```
joanestebanr: Maybe make sense to use the same params as `test-unit`? or we want to force that short ones can run in parallel? WDYT: 
```
trap '$(STOP)' EXIT; MallocNanoZone=0 go test -count=1 -short -race -p 1 -covermode=atomic -coverprofile=coverage.out -timeout 15m ./...
````
But decreasing the timeout to prevent that any no-short  test

joanestebanr (commented):
I suggest that CI also runs the short test individually for this reason: 
- If unit-test fails will be so easy to known if are the long or short ones
- To prevent a non-short test from being mistakenly classified as short.

joanestebanr (dismissed):
LGTM, I would suggest to add a Makefile target "test-unit-all" to run the all tests, shorts and logs

PR #1644, opened 2026-06-08
feat: autoclaim L1 -> L2

### LINKED ISSUE #1652: L1 -> L2 Autoclaim

arnaubennassar:
## 🔄 Changes Summary
- Adds an opt-in `autoclaim` Aggkit component for L1 -> L2 bridge claims. It discovers eligible exits from `l1bridgesync`, persists request/cursor/attempt state, prepares claim proofs with `l1infotreesync` plus a per-claimer `l2gersync`, submits `claimAsset` / `claimMessage` transactions through `EthTxManager`, and tracks requests through confirmation or failure.
- Adds per-destination claimers with named policies: `allow-all`, `api-approve`, `no-message`, and `basic-filter`.
- Adds optional Auto Claim REST endpoints under `/autoclaim/v1` for request listing, inspection, manual approval, and manual rejection, with generated Swagger docs.
- Wires Auto Claim into startup through the new `autoclaim` component, config loading/validation/defaults, docs, mocks, and the OP e2e environment.
- Hardens the bridge-service L2 deposit L1 info tree lookup used by `/l1-info-tree-index`: if a verified batch LER is missing from the local L2 exit tree, the lookup now avoids mixing L2 bridge-sync block numbers with L1 verified-batch block numbers and falls back to a bounded linear scan.

## ⚠️ Breaking Changes
- 🛠️ **Config**: No breaking migration. Auto Claim adds a new optional `[AutoClaim]` config section and remains disabled by default.
- 🔌 **API/CLI**: Adds a new optional `autoclaim` component selector and optional `/autoclaim/v1` API. Existing bridge service API paths are unchanged.

## 📋 Config Updates
- 🧾 **Diff/Config snippet**: Adds defaults for the new disabled-by-default Auto Claim section:

```toml
[AutoClaim]
Enabled = false
StoragePath = "{{PathRWData}}/autoclaim.sqlite"
Claimers = []

[AutoClaim.API]
Enabled = false
Host = "0.0.0.0"
Port = 5579

[AutoClaim.L1ToL2Watchdog]
Enabled = true
StartBlock = 0
PollInterval = "3s"
RetryAfterErrorPeriod = "1s"
EtrogL1UpgradeBlock = 0

[AutoClaim.L2ToLxWatchdog]
Enabled = false
```

- Enabling Auto Claim requires both selecting the `autoclaim` component and setting `[AutoClaim].Enabled = true`.
- Each enabled `[[AutoClaim.Claimers]]` entry configures one EVM destination network with its own RPC URL, bridge address, policy, gas settings, retry settings, and `EthTxManager` storage/signing config.
- Auto Claim owns `AutoClaim.StoragePath`; each claimer also has an independent `EthTxManager.StoragePath` and isolated per-claimer `l2gersync` / L2 reorg detector databases under the Auto Claim storage directory.
- L2 -> Lx Auto Claim remains unsupported and must stay disabled.

## ✅ Testing
- 🤖 **Automatic**: Added unit coverage across `autoclaim/api`, `claimer`, `claimtx`, `config`, `policy`, `proof`, `runtime`, `sender`, `simulator`, `storage`, `types`, and `watchdog`.
- 🤖 **Automatic**: Added Go e2e coverage for `TestAutoClaimL1ToL2AllowAll`, `TestAutoClaimL1ToL2APIApprove`, and `TestAutoClaimL1ToL2BasicFilter`.
- 🤖 **Automatic**: Current PR checks show `lint`, `test-unit`, Go e2e, Docker image build, CodeQL, SonarCloud, and the configured multi/single-chain e2e workflows passing. `Run govulncheck` is currently failing.
- 🖱️ **Manual**: No separate manual testing was run while refreshing this PR description.

## 🐞 Issues
- Close #1652 

## 🔗 Related PRs
- None listed.

## 📝 Notes
- Auto Claim currently supports L1 -> L2 only. L2 -> Lx discovery is reserved for future work.
- Request IDs are derived from `origin_network:destination_network:deposit_count`; `origin_network` is the bridged token's origin network, not necessarily the chain where the bridge was initiated.
- GER readiness is checked during proof preparation by each claimer's own `l2gersync`, which supports both legacy and sovereign L2 GER manager behavior.
- The optional `api-approve` policy requires the Auto Claim API so operators can approve or reject requests in `manual-approval-required` status.
- This branch currently skips five known flaky remove-GER e2e entrypoint tests: `TestRemoveGER_NoProblematicClaims`, `TestRemoveGER_CategoryA`, `TestRemoveGER_CategoryB1`, `TestRemoveGER_CategoryB2`, and `TestGenerateInvalidGER`. This is being worked on #1643 
- Bridge-service changes are a scoped regression fix for L2-origin bridge proof lookup. The previous fallback could call `GetFirstVerifiedBatchesAfterBlock` with `GetLastRoot().BlockNum`, which is an L2 bridge-sync block number even though the L1 info tree query expects an L1 block number. The new path only falls back on `db.ErrNotFound`, keeps the original L1 verified-batch bounds, and skips missing LERs with a linear scan when binary search cannot safely continue.


joanestebanr:
@claude

docs/autoclaim.md
```diff
@@ -0,0 +1,471 @@
...
+## Configuration
+
+Minimal L1 to L2 configuration:
+
+```toml
+[AutoClaim]
+Enabled = true
+StoragePath = "/var/lib/aggkit/autoclaim.sqlite"
+
+[AutoClaim.API]
```
joanestebanr, line 248: "Perhaps it would make more sense to use the same REST endpoint (host + port) as Bridge, allowing both services to coexist while sharing the same host and port.

Take a look to config: 
```
[REST]
Host = "0.0.0.0"
Port = 5577
ReadTimeout = "5m"
WriteTimeout = "5m"
MaxRequestsPerIPAndSecond = 10
```

autoclaim/types/types.go
```diff
@@ -0,0 +1,389 @@
...
+}
+
+// NewBridgeExitFromSync converts a bridge sync record into an Auto Claim bridge exit.
+func NewBridgeExitFromSync(bridge bridgesync.Bridge) BridgeExit {
+	return NewBridgeExitFromSyncWithEtrog(bridge, 0)
+}
+
+// NewBridgeExitFromSyncWithEtrog converts a bridge sync record using Etrog-upgrade awareness.
+func NewBridgeExitFromSyncWithEtrog(bridge bridgesync.Bridge, etrogL1UpgradeBlock uint64) BridgeExit {
+	preEtrog := isPreEtrogBridge(bridge, etrogL1UpgradeBlock)
```
joanestebanr: This transformation is not done in bridgeservice? It sounds to me like this code already exists somewhere

docs/autoclaim.md
```diff
@@ -0,0 +1,471 @@
...
+The optional REST API is not needed to submit claims. Enable it only when operators need request inspection or use the
+`api-approve` policy, which requires manual decisions through the API.
+
+## Configuration
+
+Minimal L1 to L2 configuration:
+
+```toml
+[AutoClaim]
+Enabled = true
```
joanestebanr: Maybe it's redundant... if you don't want AutoClaim  remove from component list... it would make more sense to call it DryRun or something like that, that run normally but skip the send `claim` tx

autoclaim/runtime/runtime.go
```diff
@@ -0,0 +1,688 @@
...
+	StartEthTxManager func(context.Context, EthTxManager)
+	NewPolicy         func(
+		autoclaimcfg.PolicyName,
+		autoclaimcfg.PolicyConfig,
+		...policy.RegistryOption,
+	) (autoclaimtypes.Policy, error)
+	NewTargetClaimReader func(common.Address, aggkittypes.BaseEthereumClienter) (autoclaimtypes.TargetClaimReader, error)
+	// NewGERSyncer builds the per-claimer destination-L2 GER syncer and returns a start function that
+	// runs its reorg detector and sync loop (the start function blocks and is meant to run in a goroutine).
+	NewGERSyncer func(
```
joanestebanr, line 126: The new L2GerSyncer can't set individually BlockFinality or InitialBlockNum. They inherit the configuration from the instance in config file

joanestebanr (commented):
Another point is that the API for bridgeservice and for autoclaim is very similar. But there are a different that can be a bit misleading: the first page_number: `bridgeservice` use 1 and `autoclaim` use 0

docs/autoclaim.md
arnaubennassar: Addressed by the API redesign: the public read endpoints now live in the bridge service on the shared `[REST]` host/port (`GET /bridge/v1/autoclaim/bridges[/:id]`), registered only when the `autoclaim` component runs. Only the admin `approve`/`reject` endpoints stay on the separate `[AutoClaim.API]` (its own host/port), so operators can keep admin controls off the public surface.

autoclaim/types/types.go
arnaubennassar: Good catch — the pre-Etrog/global-index computation is now a single shared helper `bridgesync.GlobalIndexForBridge`, called by both `bridgeservice.NewBridgeResponse` and autoclaim's `NewBridgeExitFromSyncWithEtrog` (the duplicated `isPreEtrogBridge` predicates were unified).

docs/autoclaim.md
arnaubennassar: Done — removed the redundant top-level `[AutoClaim] Enabled` (running is gated solely by the `--components` list) and added `DryRun`: it runs the full pipeline but skips submitting the claim tx, ending matching requests in a new terminal `dry-run` status.

autoclaim/runtime/runtime.go
arnaubennassar: Added optional per-claimer `BlockFinality` and `InitialBlockNum` keys on `[[AutoClaim.Claimers]]`, applied in `newGERSyncer` over the shared `[L2GERSync]` values (inherited when unset).

arnaubennassar:
## Review comments addressed

Pushed `2883c4f` addressing the open threads.

**API redesign (the main one).** Split endpoints by audience so operators can expose request status without exposing admin controls:
- **Public, read-only** inspection moved into the **bridge service**: `GET /bridge/v1/autoclaim/bridges[/{id}]` on the shared `[REST]` host/port, registered only when the `autoclaim` component runs (the bridge service gets an optional read-only `AutoClaimQuerier`, wired in `run.go` by opening the autoclaim storage once and sharing the handle).
- **Admin** `approve`/`reject` stay on the standalone `[AutoClaim.API]` (own host/port, can be firewalled). Shared DTOs + query parsing extracted to `autoclaim/apitypes`. This also answers the "share host/port with Bridge" comment.

**`Enabled` → `DryRun`.** Removed the redundant top-level `[AutoClaim] Enabled` (running is gated by the `--components` list); added `DryRun`, which runs the full pipeline but skips submitting the claim tx, ending requests in a new terminal `dry-run` status.

**Per-claimer GER config.** `[[AutoClaim.Claimers]]` now accepts optional `BlockFinality` / `InitialBlockNum` overrides, applied over the shared `[L2GERSync]` defaults in `newGERSyncer`.

**Dedup.** The pre-Etrog/global-index computation is now a single helper `bridgesync.GlobalIndexForBridge`, used by both `bridgeservice` and `autoclaim`.

**Admin API hardening** (flagged by the bot): max-length validation on `decider`/`decider_id`/`reason` and an `io.LimitReader` body cap.

Docs, swagger, and tests updated; the resolved CodeQL/bot threads were already outdated. Bot "suggested/optional" items (errgroup startup, metrics, etc.) left for follow-ups.

PR #1654, opened 2026-06-15
feat(exit-certificate): subtract genesis preload from EOA balances in Step B

joanestebanr:
## 🔄 Changes Summary
- **Step B1** now always subtracts each EOA's genesis (block 0) ETH balance from its live balance, so genesis-preloaded ETH (never bridged in) no longer inflates the certificate.
- `checkGenesisBalances` returns the set of EOAs preloaded at genesis (no longer aborts on any non-zero genesis balance); keeps the diagnostic breakdown, reserves its error for RPC failures.
- New `subtractGenesisBalances` removes the genesis amount per EOA, drops EOAs reduced to 0, and caps to 0 + flags any EOA whose genesis balance exceeds its live balance.
- Subtraction runs before building EOA balances/accumulations, so all outputs are genesis-adjusted.

## ⚠️ Breaking Changes
- 🔌 **Behavior**: `ignoreGenesisBalance` semantics changed. It no longer gates "any genesis preload aborts"; the preload is always subtracted, and the flag now only governs the case where the genesis balance exceeds the live balance (abort by default vs cap-to-0-and-continue).

## 📋 Config Updates
- 🧾 No new config keys. `options.ignoreGenesisBalance` (default `false`) meaning updated as above.

## ✅ Testing
- 🤖 **Automatic**: `go test ./tools/exit_certificate/...` (added `TestSubtractGenesisBalances`, `TestRunStepB1GenesisPreloadSubtracted`, `TestRunStepB1GenesisPreloadExceedsBalanceAborts`). `golangci-lint run ./tools/exit_certificate/...` → 0 issues.

## 🐞 Issues
- Related: agglayer/pm#345

## 📝 Notes
- Targets the integration branch `feature/exit-certificate-tool` (the tool is not yet in `develop`).


joanestebanr:
another PR do the same

PR #1662, opened 2026-06-22
feat: autoclaim plan

### LINKED ISSUE #1652: L1 -> L2 Autoclaim

arnaubennassar:
## 🔄 Changes Summary
- Adds an opt-in `autoclaim` Aggkit component for L1 → L2 bridge claims. It discovers eligible exits from `l1bridgesync`, persists request/cursor/attempt state, prepares claim proofs with `l1infotreesync` plus a per-claimer `l2gersync`, submits `claimAsset` / `claimMessage` transactions through `EthTxManager`, and tracks requests through confirmation or failure.
- Adds per-destination claimers with named policies: `allow-all`, `api-approve`, `no-message`, and `basic-filter`.
- **API split**: Public read endpoints (`GET /autoclaim/v1/bridges`, `GET /autoclaim/v1/bridges/:id`) are registered on the same public REST server as the bridge service (`[PublicREST]`) when the `autoclaim` component is running, under their own `/autoclaim/v1` prefix — not nested under the bridge service's `/bridge/v1` paths. Admin write endpoints (`POST /autoclaim/v1/bridges/:id/approve`, `POST /autoclaim/v1/bridges/:id/reject`) are served on a separate admin REST server (`[AdminREST]`), gated by `[AutoClaim.API].Enabled`. Shared DTOs live in `autoclaim/apitypes`.
- Adds a `DryRun` flag: when `true`, the full pipeline (discovery, policy, proof preparation) runs but the claim transaction is not submitted; requests end in the new terminal `dry-run` status.
- Allows per-claimer `BlockFinality` and `InitialBlockNum` overrides for the destination-L2 GER syncer; empty/zero falls back to the shared `[L2GERSync]` values.
- Extracts the pre-Etrog global-index computation into `bridgesync.GlobalIndexForBridge`, removing the duplication between bridge service and autoclaim.
- Hardens the admin REST API: max-length validation on `decider`, `decider_id`, and `reason` fields; `io.LimitReader` body cap (1 MiB).
- Hardens the multidownloader `CheckValidBlock` path with a new `checkValidBlockAgainstL1` fallback for blocks absent from local storage (legacy checkpoint blocks after upgrade, pruned blocks) — validated against the L1 canonical block before being accepted or rejected.
- Wires Auto Claim into startup through the new `autoclaim` component, config loading/validation/defaults, docs, mocks, and the OP e2e environment.

## ⚠️ Breaking Changes
- 🛠️ **Config: `[REST]` split into `[PublicREST]` and `[AdminREST]`**: The single top-level `[REST]` section (previously used for the bridge service's HTTP server) is replaced by two sections — `[PublicREST]` (bridge service + Auto Claim public read endpoints) and `[AdminREST]` (Auto Claim admin endpoints only). Existing configs with a `[REST]` section must rename it to `[PublicREST]`; there is no automatic migration. A leftover `[REST]` section is ignored and logs a deprecation warning at startup (warn-only for now because the e2e infra still generates configs with `[REST]`).
- 🛠️ **Config**: The new `[AutoClaim]` section is opt-in — Auto Claim only runs if `autoclaim` is added to the component list. No migration needed since this section did not previously exist.
- 🔌 **API/CLI**: Adds an optional `autoclaim` component that, when enabled, registers new public read routes (`GET /autoclaim/v1/bridges`, `GET /autoclaim/v1/bridges/:id`) on the `[PublicREST]` server alongside the existing bridge service routes, and new admin write routes (`POST /autoclaim/v1/bridges/:id/approve`, `POST /autoclaim/v1/bridges/:id/reject`) on a brand-new `[AdminREST]` server (gated by `[AutoClaim.API].Enabled`). Existing bridge service paths under `/bridge/v1` are unchanged.

## 📋 Config Updates
- 🧾 **Diff/Config snippet**: New `[AutoClaim]` section. Add `"autoclaim"` to the component list to enable:

```toml
[AutoClaim]
DryRun = false
StoragePath = "{{PathRWData}}/autoclaim.sqlite"

[AutoClaim.API]
Enabled = false

[AutoClaim.L1ToL2BridgeDetector]
Enabled = true
StartBlock = 0
PollInterval = "3s"
RetryAfterErrorPeriod = "1s"
EtrogL1UpgradeBlock = 0

[AutoClaim.L2ToLxBridgeDetector]
Enabled = false

# One entry per destination network. Add as many [[AutoClaim.Claimers]] blocks as needed.
[[AutoClaim.Claimers]]
Enabled = true
ID = "my-l2"
NetworkType = "evm"
NetworkID = 1
URLRPC = "http://l2-rpc:8545"
BridgeAddr = "0x0000000000000000000000000000000000000000"
PolicyName = "allow-all"   # allow-all | api-approve | no-message | basic-filter
GasOffset = 0
WaitPeriod = "5s"
RetryAfter = "30s"
MaxRetries = 3
# Optional: override the shared [L2GERSync] values for this claimer's destination-L2 GER syncer.
# BlockFinality = "LatestBlock"
# InitialBlockNum = 0

[AutoClaim.Claimers.Policy]
AllowMessageClaims = false   # only relevant for "basic-filter" policy
AllowedOrigins     = []      # empty = all origins allowed
AllowedTokens      = []      # empty = all tokens allowed
ManualFallback     = false   # fall back to "api-approve" when filter rejects
MaxGas             = 0       # 0 = unlimited

[AutoClaim.Claimers.EthTxManager]
StoragePath = "{{PathRWData}}/autoclaim-claimer-my-l2-ethtx.sqlite"
# PrivateKeys, ForcedGas, MaxGasPrice, RetryInterval, WaitTxToBeMined, etc.
# See [EthTxManager] in the main config for the full list of fields.
```

The HTTP servers are configured via the top-level `[PublicREST]` and `[AdminREST]` sections. `[PublicREST]` replaces the old `[REST]` section (same fields, serves the bridge service plus, when enabled, Auto Claim's public read routes). `[AdminREST]` is brand new — a separate server that only exists to host Auto Claim's admin write routes:

```toml
[PublicREST]
Host = "0.0.0.0"
Port = 5577
ReadTimeout = "5m"
WriteTimeout = "5m"
MaxRequestsPerIPAndSecond = 10

[AdminREST]
Host = "0.0.0.0"
Port = 5579
ReadTimeout = "5m"
WriteTimeout = "5m"
MaxRequestsPerIPAndSecond = 10
```

- Enabling Auto Claim requires adding `"autoclaim"` to the process component list; `DryRun = false` (default) means claims will be submitted once a claimer is configured.
- Each `[[AutoClaim.Claimers]]` entry configures one EVM destination network with its own RPC URL, bridge address, policy, gas settings, retry settings, `EthTxManager` storage/signing config, and optional `BlockFinality`/`InitialBlockNum` GER-sync overrides.
- Auto Claim owns `AutoClaim.StoragePath`; each claimer also has an independent `EthTxManager.StoragePath` and isolated per-claimer `l2gersync`/reorg-detector databases under the Auto Claim storage directory.
- L2 → Lx Auto Claim remains unsupported and must stay disabled.

## ✅ Testing
- 🤖 **Automatic**: Unit coverage added across `autoclaim/api`, `autoclaim/apitypes`, `claimer`, `claimtx`, `config`, `policy`, `proof`, `runtime`, `sender`, `simulator`, `storage`, `types`, and `watchdog`. Coverage target ≥ 80% new code.
- 🤖 **Automatic**: Go e2e coverage for `TestAutoClaimL1ToL2AllowAll`, `TestAutoClaimL1ToL2APIApprove`, and `TestAutoClaimL1ToL2BasicFilter`.
- 🤖 **Automatic**: `lint`, `test-unit`, Go e2e, Docker image build, CodeQL, SonarCloud, and multi/single-chain e2e workflows all pass. `Run govulncheck` is currently failing (pre-existing, not introduced by this PR).

## 🐞 Issues
- Close #1652

## 🔗 Related PRs
- None.

## 📝 Notes
- Auto Claim currently supports L1 → L2 only. L2 → Lx discovery is reserved for future work.
- Request IDs are derived from `origin_network:destination_network:deposit_count`; `origin_network` is the bridged token's origin network, not necessarily the chain where the bridge was initiated.
- GER readiness is checked during proof preparation by each claimer's own `l2gersync`, which supports both legacy and sovereign L2 GER manager behavior.
- The `api-approve` policy requires the Auto Claim admin REST API to be enabled so operators can approve or reject requests in `manual-approval-required` status.
- The `dry-run` terminal status is distinct from `policy-rejected`: it means the full pipeline ran successfully but transaction submission was intentionally skipped.
- This branch currently skips five known flaky remove-GER e2e tests: `TestRemoveGER_NoProblematicClaims`, `TestRemoveGER_CategoryA`, `TestRemoveGER_CategoryB1`, `TestRemoveGER_CategoryB2`, and `TestGenerateInvalidGER`. Tracked in #1643.
- The `checkValidBlockAgainstL1` multidownloader fallback only activates for blocks at or below the finalized height; blocks above finality are still treated as inconsistencies and trigger a retry.


joanestebanr (commented):
Update PR description with a  full config example (missing important fields like `Claimers`). Also fix outdated comments like references to `[AutoClaim].Enabled = true`

config/default.go
```diff
@@ -151,13 +151,20 @@ ReadTimeout = "2s"
 WriteTimeout = "2s"
 MaxRequestsPerIPAndSecond = 10
 
-[REST]
```
joanestebanr, line 154: - Add a rule run into `deprecatedFieldsOnConfig` to warn about the renaming
- Add it to PR description 'breaking changes'

config/default.go
```diff
@@ -151,13 +151,20 @@ ReadTimeout = "2s"
...
+[PublicAPI]
 Host = "0.0.0.0"
 Port = 5577
 ReadTimeout = "5m"
 WriteTimeout = "5m"
 MaxRequestsPerIPAndSecond = 10
 
+[AdminAPI]
```
joanestebanr: Add to PR description

cmd/run.go
```diff
@@ -216,6 +239,39 @@ func start(cliCtx *cli.Context) error {
...
+		}
+		if acRuntime != nil {
+			acRuntime.PublicAPI.RegisterRoutes(publicServer.Engine())
+			if acRuntime.AdminAPI != nil {
+				acRuntime.AdminAPI.RegisterRoutes(adminServer.Engine())
+			}
+		}
+	}
+
+	go publicServer.Start(ctx)
```
joanestebanr: Don't make sense to run always both service, only if have any end-point to server. WDYT?

common/httpserver.go
```diff
@@ -0,0 +1,105 @@
...
+// Start starts the HTTP server and blocks until ctx is done, then shuts down gracefully.
+func (s *HTTPServer) Start(ctx context.Context) {
+	srv := &http.Server{
+		Addr:         s.cfg.Address(),
+		Handler:      s.engine,
+		ReadTimeout:  s.cfg.ReadTimeout.Duration,
+		WriteTimeout: s.cfg.WriteTimeout.Duration,
+	}
+
+	go func() {
```
joanestebanr, line 69: How will it behave if there's a port conflict or if it can't listen for some reason? Maybe it should return an error?

joanestebanr, line 69: There are a double 'go', in `run.go` is launching it with `go publicServer.Start(ctx)`

joanestebanr:
@claude

config/default.go
arnaubennassar: Done — added `AutoClaim.L1ToL2Watchdog` and `AutoClaim.L2ToLxWatchdog` entries to `deprecatedFieldsOnConfig` in `config/config.go`.

common/httpserver.go
arnaubennassar: Fixed — the naked `go publicServer.Start(ctx)` in `run.go` is now replaced with an explicit `go func() { if err := ...; err != nil { log.Errorf(...) } }()` wrapper. The inner goroutine in `Start()` remains for the graceful-shutdown listener, and `Start()` itself blocks on `ListenAndServe` as before.

config/default.go
arnaubennassar: Added to PR description `Breaking Changes` section.

cmd/run.go
arnaubennassar: Fixed — servers now only start when at least one component has registered routes on them. `publicHasRoutes` is set when bridge service or autoclaim public API registers routes; `adminHasRoutes` is set when autoclaim admin API registers routes.

common/httpserver.go
arnaubennassar: Fixed — `Start()` now returns `error` instead of panicking. Port-bind failures are propagated to the caller, which logs them via `log.Errorf`.

arnaubennassar:
@joanestebanr Updated the PR description with a full `[[AutoClaim.Claimers]]` example showing all fields (ID, NetworkType, NetworkID, URLRPC, BridgeAddr, PolicyName, Policy sub-section, EthTxManager sub-section, BlockFinality/InitialBlockNum overrides). Also updated the `Breaking Changes` section to document the watchdog field renames as deprecated (they now emit startup warnings) and removed the now-inaccurate `Claimers = []` placeholder from the config snippet.

joanestebanr:
> @joanestebanr Updated the PR description with a full `[[AutoClaim.Claimers]]` example showing all fields (ID, NetworkType, NetworkID, URLRPC, BridgeAddr, PolicyName, Policy sub-section, EthTxManager sub-section, BlockFinality/InitialBlockNum overrides). Also updated the `Breaking Changes` section to document the watchdog field renames as deprecated (they now emit startup warnings) and removed the now-inaccurate `Claimers = []` placeholder from the config snippet.

Still inacurate PR description: 

> API split: Public read endpoints (GET /bridge/v1/autoclaim/bridges, GET /bridge/v1/autoclaim/bridges/:id) are served by the bridge service when the autoclaim component is running. Admin write endpoints (POST /autoclaim/v1/bridges/:id/approve, POST /autoclaim/v1/bridges/:id/reject) remain on the standalone admin API (configurable via [AutoClaim.API]). Shared DTOs live in autoclaim/apitypes.

I think that the end-points are: `autoclaim/v1/bridges` isn't?

> 🛠️ Config: No breaking migration. The former [AutoClaim].Enabled flag is replaced by [AutoClaim].DryRun. Auto Claim is disabled by default; enabling it requires adding autoclaim to the component list.

There are a breaking change: [REST] -> [PublicAPI]

> 🔄 Config field renames (deprecation warnings): The watchdog sub-sections are renamed: AutoClaim.L1ToL2Watchdog → AutoClaim.L1ToL2BridgeDetector and AutoClaim.L2ToLxWatchdog → AutoClaim.L2ToLxBridgeDetector. Configs using the old names will emit a deprecation warning at startup — update your config files to the new names.

This change is internal to this PR, There's no need to include it.

> 🔌 API/CLI: Adds optional /bridge/v1/autoclaim/bridges routes to the bridge service (present only when autoclaim component runs) and an optional /autoclaim/v1 admin API. Existing bridge service paths are unchanged.

No longer apply

> Config Updates

Missing new PublicAPI / AdminAPI

I also miss in the description the important change to add an extra REST API Admin

joanestebanr:
Another suggestion, to keep the same format as before I suggest to rename `PublicAPI` to `PublicREST` and `AdminAPI` to `AdminREST`. 
The point is that there are JSON-RPC end-points and REST, and the former field was `REST`

config/config.go
```diff
@@ -239,6 +242,14 @@ var (
 			FieldNamePattern: "AggSender.EpochNotificationPercentage",
 			Reason:           aggsenderEpochPercentageHint,
 		},
+		{
+			FieldNamePattern: "AutoClaim.L1ToL2Watchdog",
```
joanestebanr: This can't be a deprecated field because never exists a version with this fields, you can remove it

config/config.go
```diff
@@ -239,6 +242,14 @@ var (
 			FieldNamePattern: "AggSender.EpochNotificationPercentage",
 			Reason:           aggsenderEpochPercentageHint,
 		},
+		{
```
joanestebanr, line 247: Still missing deprecation of "REST" section

config/config.go
arnaubennassar: Removed both entries — you're right, `AutoClaim.L1ToL2Watchdog`/`L2ToLxWatchdog` never shipped, so there's nothing to deprecate. Pushed in 562941a.

config/config.go
arnaubennassar: Added — `REST` is now in `deprecatedFieldsOnConfig` with a hint pointing to `PublicREST`/`AdminREST`, plus a test case. Pushed in 562941a.

cmd/run.go
```diff
@@ -216,6 +241,53 @@ func start(cliCtx *cli.Context) error {
...
+			if acRuntime.AdminREST != nil {
+				acRuntime.AdminREST.RegisterRoutes(adminServer.Engine())
+				adminHasRoutes = true
+			}
+		}
+	}
+
+	if publicHasRoutes {
+		go func() {
+			if err := publicServer.Start(ctx); err != nil {
```
joanestebanr: The issue still persists where, if the port is already in use, the service starts anyway and only logs an error. This is a case that could easily go unnoticed, because you would expect the process to terminate if it cannot start the endpoint.

arnaubennassar: Good catch — switched both to `log.Fatalf` so the process terminates on a listen failure (port conflict, etc.) instead of silently continuing with a dead REST server. Matches the existing pattern used for `l1MultiDownloader.Start`. Pushed in 8284650.

common/httpserver.go
arnaubennassar: Reworked in dee164ad: `Start()` now binds the listener synchronously and serves in the background, so the call site in `run.go` no longer wraps it in a goroutine at all — startup just does `if err := publicServer.Start(ctx); err != nil { log.Fatalf(...) }`.

cmd/run.go
arnaubennassar: You were right that Fatalf-in-a-goroutine still let startup proceed past the failure. In dee164ad the listener is created synchronously inside `Start()` (`net.Listen` before returning), so a port conflict now aborts startup immediately and deterministically — the "listening on" log only appears after a successful bind.

config/config.go
arnaubennassar: FYI: the fatal `[REST]` deprecation added in 562941a broke e2e (kurtosis-cdk still generates configs with a `[REST]` section), so in dee164ad it's a warn-only deprecation instead — the section is ignored and logs a startup warning pointing to `PublicREST`/`AdminREST`. Making it fatal once the e2e templates migrate is tracked in #1702.

joanestebanr:
@claude

joanestebanr (approved):

PR #1675, opened 2026-06-29
fix(exit-certificate): AET-04: fail-fast when a token balanceOf batch errors in Step B

### LINKED ISSUE #1679: exit_tool:  SP001: When any wrapped token's balanceOf batch fails inside step_b.fetchAllTokenBalances(), the error is silently discarded
When any wrapped token's balanceOf batch fails inside step_b.fetchAllTokenBalances(), the error is silently discarded. The affected token is absent from the returned balance map, causing Step C to treat its entire LBT supply as SC-locked value. The final certificate sends the whole token supply to exitAddress rather than distributing it across individual EOA holders, excluding those holders from the certificate entirely.

joanestebanr:
## 🔄 Changes Summary
- **Fix (High):** `fetchAllTokenBalances` (Step B1) silently discarded per-token RPC errors (`log.Warnf` + `return`), leaving the failed token absent from the balance map. Step C then treated its entire LBT supply as SC-locked and Step D routed the whole supply to `exitAddress`, excluding the real EOA holders from the certificate.
- Made it **fail-fast**, consistent with its sibling helpers (`classifyAddresses`, `fetchETHBalances`): `fetchAllTokenBalances` now returns an `error` and uses `errgroup.WithContext` (limited to `tokenConcurrency`) to cancel the remaining scans on the first failure. `RunStepB1` propagates it and aborts the pipeline.
- Added a doc comment explaining why the error must not be dropped.

## ⚠️ Breaking Changes
- None. No config, API, or CLI changes. Behaviour change only: a persistent per-token RPC failure now aborts the pipeline (after the existing RPC retries) instead of producing a misrouted certificate.

## 📋 Config Updates
- None.

## ✅ Testing
- 🤖 **Automatic:** Added `TestFetchAllTokenBalancesFailFast` (a failing `balanceOf` batch must return an error, never omit the token). Updated `TestFetchAllTokenBalances` to the new signature. Full package suite passes (`go test ./tools/exit_certificate/`) and `golangci-lint` reports 0 issues.

## 🐞 Issues
- Closes #1679

## 🔗 Related PRs
- N/A

## 📝 Notes
- Chose fail-fast as the default without adding an opt-out flag (e.g. `ignoreTokenBalanceFetchError`) to keep the config surface minimal; can add one if a knowingly-degraded test mode is desired.

### DIFF
--- a/tools/exit_certificate/step_b.go
+++ b/tools/exit_certificate/step_b.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/agglayer/aggkit/log"
 	"github.com/ethereum/go-ethereum/common"
+	"golang.org/x/sync/errgroup"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


joanestebanr:
AET-04 Silent Token Balance Fetch Failure In fetchAllTokenBalances

PR #1683, opened 2026-06-29
fix(exit-certificate): AET-10 `step E`: use configured l2NetworkId for zkevm pending-bridges `dest_net`

### LINKED ISSUE #1681: exit_tool: SP003: step_e.fetchZkevmPendingBridges() hardcodes dest_net=1 in the HTTP query string when fetching pending bridges
step_e.fetchZkevmPendingBridges() hardcodes dest_net=1 in the HTTP query string when fetching pending bridges from the zkevm bridge service, instead of using the configured l2NetworkId. For any L2 chain whose network ID is not 1, the function queries deposits targeting a different chain entirely, causing Step E's bridge service cross-check to silently pass or spuriously fail against the wrong deposit set.

joanestebanr:
## 🔄 Changes Summary
- `step_e.fetchZkevmPendingBridges()` hardcoded `dest_net=1` in the `GET /pending-bridges` query string. For any L2 chain whose network ID is not `1`, Step E's bridge service cross-check queried deposits targeting a different chain, causing the check to silently pass or spuriously fail against the wrong deposit set.
- The function now takes a `destNet` parameter and the caller passes `cfg.L2NetworkID`.

## ⚠️ Breaking Changes
- None.

## 📋 Config Updates
- None.

## ✅ Testing
- 🤖 **Automatic**: `TestFetchZkevmPendingBridges` now asserts the request's `dest_net` query param matches the configured network ID (locking the fix). All `tools/exit_certificate` unit tests pass.

## 🐞 Issues
- Closes #1681

## 📝 Notes
- The aggkit bridge service path already filtered by `cfg.L2NetworkID` (`destination_network`); only the zkevm path was affected.

### DIFF
--- a/tools/exit_certificate/step_e.go
+++ b/tools/exit_certificate/step_e.go
@@ -429,7 +429,7 @@ func checkBridgeServicePendingBridges(ctx context.Context, cfg *Config, unclaime
 		label = "zkevm bridge service"
 		log.Infof("Querying zkevm bridge service for pending bridges (url=%s, l2NetworkID=%d)", baseURL, cfg.L2NetworkID)
 		var fetchErr error
-		svcCounts, fetchErr = fetchZkevmPendingBridges(ctx, baseURL, leafTypeAsset)
+		svcCounts, fetchErr = fetchZkevmPendingBridges(ctx, baseURL, cfg.L2NetworkID, leafTypeAsset)
 		if fetchErr != nil {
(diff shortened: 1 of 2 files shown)


joanestebanr:
AET-10 Hardcoded dest_net=1 In ZkEVM Bridge Service Query Bypasses Conﬁgured L2 Network ID

PR #1684, opened 2026-06-29
fix(exit-certificate): AET-12 :`step E` prevent uint64 overflow panic in decodeABIString

### LINKED ISSUE #1680: exit_tool: SP002: A uint64 integer overflow in step_e.decodeABIString()
A uint64 integer overflow in step_e.decodeABIString() allows a maliciously crafted L1 ERC-20 token whose name() function returns an oversized ABI length field to bypass the function's bounds guard and trigger a process-fatal panic, permanently preventing the exit certificate tool from generating a certificate for any chain where such a token has an unclaimed L1-to-L2 deposit.

joanestebanr:
## 🔄 Changes Summary
- Fix a uint64 integer overflow in `step_e.decodeABIString()` (SP002). A maliciously crafted L1 ERC-20 whose `name()`/`symbol()` returns an oversized ABI length field could make `64+strLen` wrap past `MaxUint64` to a small value, bypassing the bounds guard and triggering a slice out-of-range **panic** that aborts the exit certificate tool for any chain with an unclaimed L1→L2 deposit of such a token.
- Reject length fields that don't fit in `uint64` (`big.Int.IsUint64()`) and compare `strLen` against the remaining byte count without the overflowing addition (`strLen > uint64(len(data))-twoABIWords`, safe since `len(data) >= 64`).

## ⚠️ Breaking Changes
- None.

## 📋 Config Updates
- None.

## ✅ Testing
- 🤖 **Automatic**: Added `TestDecodeABIString` with regression cases — `MaxUint64` length field (the overflow case) and a length field larger than `uint64` — asserting no panic and `""` return, plus valid/empty/too-short/length-exceeds-data cases. Full `go test ./tools/exit_certificate/...` passes; `go vet` and `gofmt` clean.
- 🖱️ **Manual**: n/a (deterministic decode logic covered by unit test).

## 🐞 Issues
- Closes #1680

## 📝 Notes
- The guard now also rejects length fields exceeding `uint64` range, which `big.Int.Uint64()` would have silently truncated.

### DIFF
--- a/tools/exit_certificate/step_e.go
+++ b/tools/exit_certificate/step_e.go
@@ -348,11 +348,18 @@ func decodeABIString(data []byte) string {
 	if len(data) < twoABIWords {
 		return ""
 	}
-	strLen := new(big.Int).SetBytes(data[32:64]).Uint64()
-	if 64+strLen > uint64(len(data)) {
+	lenField := new(big.Int).SetBytes(data[abiWordBytes:twoABIWords])
+	// Reject lengths that don't fit in uint64; the payload can never be that large anyway.
+	if !lenField.IsUint64() {
 		return ""
 	}
-	return string(data[64 : 64+strLen])
+	strLen := lenField.Uint64()
+	// Compare against the remaining bytes without computing twoABIWords+strLen, which would
+	// overflow uint64 for a maliciously oversized length field and wrap past this guard.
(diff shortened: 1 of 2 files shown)


joanestebanr:
AET-12 uint64 Addition Overﬂow In decodeABIString Bypasses Bounds Guard And Crashes The Process

PR #1692, opened 2026-07-01
feat: add bridgeservicefinder package

### LINKED ISSUE #1666: Sequencer URL monitor

arnaubennassar:
## 🔄 Changes Summary
- Adds a new `bridgeservicefinder` package that resolves and serves the bridge service URL for every network attached to a rollup manager, exposing `New(...)`, `Start(ctx)`, and `GetURL(networkID)`. Not yet wired into `cmd/run.go` or the app config — it's intended to be consumed by other components in the future.
- URL resolution follows three sources in strict priority: (1) static config map, (2) on-chain `aggchainMetadata` key `"BRIDGE_SERVICE_URL"` on the aggchain contract, (3) `trustedSequencerURL()` with the port substituted to the hardcoded bridge REST port `5577`.
- Networks are enumerated from the real rollup manager (`RollupCount`/`RollupIDToRollupData`) rather than a static list; `networkID == rollupID` for L2s, with network 0 (L1) only servable via config.
- A lightweight, self-contained polling loop (`FilterLogs` over a configurable block-finality window) watches `SetTrustedSequencerURL` and `AggchainMetadataSet` events and updates the cache live, respecting source priority (config is never overridden) and a health-gating rule (only replace a cached URL with a new one if it's reachable, unless the previous URL was itself unreachable).
- Adds selector/event-compatible mock Solidity contracts (`test/contracts/rollupmanagermock`, `test/contracts/aggchainrollupmock`) + generated bindings, verified against the real `agglayermanager`/`aggchainbase`/`polygonrollupbaseetrog` bindings via a smoke test.

## ⚠️ Breaking Changes
- 🛠️ **Config**: None.
- 🔌 **API/CLI**: None.

## 📋 Config Updates
- 🧾 **Diff/Config snippet**: None — no app config or wiring is added in this PR. The package is standalone and not yet started by any binary.

## ✅ Testing
- 🤖 **Automatic**: Extensive unit tests on a go-ethereum simulated backend (`bridgeservicefinder/*_test.go`) covering each source in isolation, all priority combinations, config-immunity to on-chain updates, metadata-over-sequencer precedence, multi-network initial-cache builds, unreachable-service handling at `Start`, live updates via both event types, and health-gating edge cases. A smoke test (`test/contracts/bridgeservicefinder_mocks_test.go`) confirms the mock contracts are selector/event-compatible with the real bindings.
- 🖱️ **Manual**: `make build`, `make lint` (0 issues), `make test-unit`, and `make generate-mocks` (no diff) all verified clean.

## 🐞 Issues
- Closes #1666 

## 🔗 Related PRs
- None

## 📝 Notes
- The `aggchainMetadata`/`trustedSequencerURL` calls are treated as "source not available" (graceful fallthrough) on any contract-call error, since go-ethereum bindings don't reliably distinguish "method absent/reverted" from other RPC errors — this favors graceful degradation over strict error classification.
- Live-update seeding starts from the current finalized block at `Start()` time (no historical backfill), since the initial cache build already covers the current state.
- Intentionally not wired into `cmd/run.go`/`config/*` in this PR — a future component will construct and start the finder itself.


bridgeservicefinder/cache.go
```diff
@@ -0,0 +1,49 @@
+package bridgeservicefinder
```
joanestebanr, line 1: Internally to implement a cache is used  `github.com/jellydator/ttlcache/v3` maybe you can use the same?

bridgeservicefinder/bridgeservicefinder_test.go
```diff
@@ -0,0 +1,189 @@
+package bridgeservicefinder
```
joanestebanr, line 1: It's a bit confussing: usually the unitest have the same name as the file that is checking. There are no finder.go, by other hand the work `mocks` also it's confusing

bridgeservicefinder/bridgeservicefinder.go
```diff
@@ -0,0 +1,299 @@
...
+	// HealthChecker probes resolved URLs. Defaults to the HTTP /health checker built from cfg.
+	HealthChecker HealthChecker
+	// HTTPClient is passed to the default HealthChecker. Defaults to a client with the configured
+	// health-check timeout. Ignored when HealthChecker is supplied.
+	HTTPClient *http.Client
+	// LogFilterer is the eth-client surface the (S4) listener uses. Stored now for the later live-
+	// update step; not exercised by Start's initial cache build. Defaults to EthClient.
+	LogFilterer LogFilterer
+	// Logger is the logger used by the finder. Defaults to log.WithFields("module", moduleName).
+	Logger *log.Logger
```
joanestebanr: better use the interface `common.Logger`

joanestebanr (commented):
- check the naming of unittest that doesn't follow the rest of code
- the review is not finished yet

bridgeservicefinder/bridgeservicefinder.go
```diff
@@ -0,0 +1,299 @@
...
+	if url, ok := f.cfg.URLs[rollupID]; ok && url != "" {
+		f.cache.set(rollupID, cacheEntry{url: url, source: SourceConfig})
+		f.logger.Debugf("network %d resolved from config: %s", rollupID, url)
+
+		return
+	}
+
+	data, err := f.rollupManager.RollupIDToRollupData(&bind.CallOpts{Context: ctx}, rollupID)
+	if err != nil {
+		f.logger.Warnf("failed to read rollup data for network %d, skipping: %v", rollupID, err)
```
joanestebanr: maybe must return an error? you would have a degrated behaviour that would be difficult to find out... the same for the rest of warnings.

This function appears to be called only during startup. If a rollup is added after the service has started, it won't be recognized, and the system will require a restart.

bridgeservicefinder/bridgeservicefinder.go
arnaubennassar: Done in fbd3c0d — switched the finder, health checker and listener to the `common.Logger` interface instead of the concrete `*log.Logger`.

bridgeservicefinder/bridgeservicefinder_test.go
arnaubennassar: Good catch — renamed `finder_mocks_test.go` -> `bridgeservicefinder_test.go` to match the file under test (`bridgeservicefinder.go`, which is where `New`/`Start` live). Done in fbd3c0d.

bridgeservicefinder/cache.go
arnaubennassar: I looked at reusing `ttlcache/v3` (as in `agglayer/agglayer_client_cache.go`) but it's a mismatch for this cache: entries here have no TTL and no capacity/eviction — they live for the process lifetime and are mutated in place by the event listener (source-priority + health-gating). We'd have to store everything with `ttlcache.NoTTL`, which effectively turns it into a plain concurrent map while adding a dependency and losing the typed `cacheEntry` (url/source/healthy) semantics that `get`/`set` and `probeAll` rely on. So I've kept the small purpose-built `RWMutex` map. Happy to reconsider if you'd still prefer the shared type.

bridgeservicefinder/bridgeservicefinder.go
arnaubennassar: Both addressed in d499dc34.

**1. Hard errors vs warnings.** `buildInitialCache`/`resolveNetwork` now distinguish a genuine RPC/transport failure (reading rollup data, building the reader, or a non-fall-through resolution error) — which is returned and aborts `Start` — from a benign `ErrNoSourceAvailable` (this rollup simply exposes no bridge service yet), which is still logged + skipped. So a broken L1 endpoint fails loudly instead of silently dropping networks, while a legitimately source-less rollup keeps its address in the routing table so a later event can populate it.

**2. Rollups added after startup.** The listener now also watches the rollup manager for `CreateNewRollup` / `CreateNewAggchain` / `AddExistingRollup` and registers the announced network live (resolves its URL, caches it, adds its contract to the watched set), so new rollups are served without a restart. The manager address is always watched, so even a fleet that had zero rollups at `Start` can discover its first ones. Discovery runs on the polling goroutine, so its failures are logged rather than aborting the loop.

Added the three lifecycle events to `RollupManagerMock` (+ regenerated binding), a selector-compatibility smoke test against the real `agglayermanager` binding, and discovery integration tests (immediate-resolve + no-source-then-healed-by-event). `make build`/`lint`/`test-unit` (incl. `-race`) all clean.

joanestebanr (approved):
LGTM

PR #1698, opened 2026-07-03
fix(exit-certificate): AET-01 replay external native tokens as ERC-20 in Step G

### LINKED ISSUE #1693: bug: AET-01 Detailed Findings External Native-Token Exits Are Replayed As The Local Gas Token
The exit tool misclassiﬁes bridge exits for external native tokens (tokens whose origin is another network’s native
asset) as the local gas token during Step G replay. This causes the certiﬁcate’s NewLocalExitRoot to be computed from
incorrect bridge leaves, while the tool’s internal consistency check still passes because both the shadow-fork replay
and the oﬀchain tree builder make the same misclassiﬁcation.
Step G’s isNativeBridgeExit() checks only OriginTokenAddress == zero without requiring the origin network to match
the gas-token network. This conﬂates two distinct assets:
• the local native gas token, e.g. (originNetwork=0, originTokenAddress=0x0);
• another network’s native asset represented on this L2 as a wrapped ERC-20, e.g.
originTokenAddress=0x0).

joanestebanr:
## 🔄 Changes Summary
- Step G misclassified bridge exits for **external native tokens** (another network's native asset, represented on this L2 as a wrapped ERC-20 with `OriginTokenAddress=0x0` but a non-local `OriginNetwork`) as the **local gas token**, computing `NewLocalExitRoot` from the wrong leaves. The internal consistency check still passed because both the shadow-fork replay and the off-chain tree builder made the same misclassification.
- `isNativeBridgeExit` (`step_g2.go`): removed the standalone `OriginTokenAddress == 0x0` condition. An exit is now native only when `TokenInfo == nil` or its origin identity (network **and** address) matches the gas token's.
- `buildLiteTreeFromCertificate` (`step_g_events.go`): the off-chain tree builder now selects the leaf origin via `isNativeBridgeExit` instead of an address-only check, keeping it in lock step with the shadow-fork replay's native/ERC-20 decision.
- `resolveTokenAddresses` already delegated to `isNativeBridgeExit`, so external native tokens now correctly resolve their L2 wrapped address instead of being skipped.

## ⚠️ Breaking Changes
- None.

## 📋 Config Updates
- None.

## ✅ Testing
- 🤖 **Automatic**: Added AET-01 regression case to `TestIsNativeBridgeExit` (`(OriginNetwork=5, OriginTokenAddress=0x0)` must not be native). Full package suite passes: `go test ./tools/exit_certificate/`. `golangci-lint run ./tools/exit_certificate/...` clean.

## 🐞 Issues
- Closes #1693

## 📝 Notes
- The tool only supports ETH-gas-token chains (Step CHECK #7), so in practice `gasToken` is `(network=0, addr=0x0)`; the fix is general regardless of gas token identity.

### DIFF
--- a/tools/exit_certificate/step_g2.go
+++ b/tools/exit_certificate/step_g2.go
@@ -972,11 +972,17 @@ func logReplayProgress(done, total int, start time.Time) {
 		done, total, float64(done)/float64(total)*percentMultiplier, rate, eta)
 }
 
+// isNativeBridgeExit reports whether a bridge exit is the local native gas token — the only asset
+// bridged via the native path (bridgeAsset with token=0x0 and msg.value). An exit is native only when
+// its origin identity matches the gas token's (network AND address), or when TokenInfo is nil (which
+// defaults to the gas token). A zero OriginTokenAddress alone is NOT sufficient: another network's
+// native asset is represented on this L2 as a wrapped ERC-20 with OriginTokenAddress=0x0 but a
+// non-local OriginNetwork, and must be replayed as an ERC-20 — matching on address alone would
+// misclassify it as the gas token and compute the LER from the wrong leaves.
 func isNativeBridgeExit(
 	ti *agglayertypes.TokenInfo, gasTokenNetwork uint32, gasTokenAddress common.Address,
 ) bool {
 	return ti == nil ||
-		ti.OriginTokenAddress == (common.Address{}) ||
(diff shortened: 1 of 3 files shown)


joanestebanr:
@codex review

joanestebanr:
@claude

joanestebanr:
@codex review

PR #1691, opened 2026-07-01
feat: add op-pp-2chains env and L2→L2 bridge test

arnaubennassar:
## 🔄 Changes Summary
- Adds a two-network OP-PP e2e environment (`test/e2e/envs/op-pp-2chains`) generated from a kurtosis-cdk snapshot, running two OP-PP L2 chains (chain IDs 20201/20202) on a shared L1 and agglayer instance.
- Generalizes the e2e loader to expose `L2B` (second L2) alongside `L2` (first L2/L2A) without breaking the existing single-chain `op-pp` env.
- Implements `BridgeL2ToL2` helper in `bridge_utils.go` that bridges a MintableERC20 token L2A → L2B and asserts the destination ERC20 balance after claim.
- Adds `TestBridgeL2ToL2` exercising the full L2→L2 bridge flow end-to-end.
- Matrices the CI workflow (`test-go-e2e.yml`) over `[op-pp, op-pp-2chains]` with `fail-fast: false`; bumps job timeout to 60 min for GER-propagation headroom.
- Fixes `BridgeL1ToL2` / `BridgeL1ToL2WithResult` to use the **actual** injected `L1InfoTreeIndex` from `GetInjectedL1InfoLeaf` for claim proofs (the API returns the first GER injected with index ≥ requested; using a stale lower index produced a GER hash absent from L2's `GlobalExitRootMap`, causing `ClaimAsset` to revert).
- Fixes two `claimsync` bugs: concurrent block-0 insert race in `L2ClaimSyncer` startup and incorrect finality parameter in the EVMDownloader safe-zone check.

## ⚠️ Breaking Changes
- None

## 📋 Config Updates
- None

## ✅ Testing
- 🤖 **Automatic**: `TestBridgeL2ToL2` added and passing in CI under `AGGKIT_E2E_ENV=op-pp-2chains`. Both `op-pp` and `op-pp-2chains` matrix jobs pass.
- 🖱️ **Manual**: `AGGKIT_E2E_ENV=op-pp-2chains go test -v -timeout 30m -run TestBridgeL2ToL2 ./test/e2e/` passes from a cold Docker start (~250 s total including post-test L1↔L2 health checks).

## 🐞 Issues
- N/A

## 🔗 Related PRs
- kurtosis-cdk snapshot branch: `feat/op-pp-2chains-snapshot` @ `627340aa`

## 📝 Notes
[...]

PR #1705, opened 2026-07-07
feat: auto claim L2->L1 and L2->L2 support

### LINKED ISSUE #1665: L2 -> Ly Autoclaim

arnaubennassar:
## 🔄 Changes Summary

- Extends the existing Auto Claim service (L1→L2 only, from #1662) with automatic claiming for the other two
  bridge directions:
  - **L2→L1**: bridge exits initiated on a rollup, destined for L1 (destination network `0`).
  - **L2→L2**: bridge exits initiated on one rollup, destined for another rollup.
  - **L1→L2 behavior is unchanged in shape** — same detector, same claim path — but its proof-readiness gate now
    goes through the destination bridge service instead of a local GER syncer (see below); regression-tested end
    to end (see Testing).
- The node running Auto Claim does **not** sync any source L2 directly. A new `bridgedetector.L2ToLx` detector
  watches `l1infotreesync` for each source rollup's local exit root (LER) advancing — via
  `VerifyBatchesTrustedAggregator`, which the rollup manager emits for **both** zkEVM and pessimistic/aggchain
  verifications — then fetches the new bridges from that source rollup's **own bridge service** via a new
  `bridgeservicefinder`-resolved REST call. Discovery is hardened per review:
  - The block-window cursor is held just before the verify row of a source skipped for a transient reason
    (unresolved bridge-service URL, source not synced yet), so such sources are durably retried on later polls
    even if they never publish another LER.
  - The destination filter is partitioned into batches of at most `bridgeservice.MaxNetworkIDs` per
    claim-candidates request (more than five enabled destinations previously produced an HTTP 400).
  - A `StartL1Block` predating the first L1 info tree leaf falls back to full history instead of failing every
    poll.
- A new bridge service endpoint, `GET /bridge/v1/claim-candidates`, serves exactly that: bridges in a deposit-count
  range, filtered by destination network. Backed by a new deposit-range+destination query in `bridgesync`
  (`GetBridgesInDepositRange`). It does **not** return a per-bridge Merkle proof — see "leaf→LER proof" below.
- **GER-injection gate via the destination bridge service, replacing per-claimer `l2gersync`.** Auto Claim no
  longer runs a per-claimer `l2gersync` instance (no isolated per-claimer SQLite database, no per-claimer L2 reorg
  detector, no per-claimer L2 RPC client dedicated to GER tracking). Instead, during proof preparation, readiness
  for an L2-destination claimer is gated by calling that **destination** network's own aggkit bridge service `GET
  /bridge/v1/injected-l1-info-leaf` (resolved through `bridgeservicefinder.Finder`). This applies uniformly to
  **both directions**, including L1→L2. A claimer whose destination is L1 (`NetworkID = 0`) has no such gate — it
  is ready as soon as `l1infotreesync` has the relevant leaf, since the GER already exists in the L1 GER manager by
  construction.
- **Leaf→LER proof fetched only at claim time.** The `L2ToLx` detector does not fetch or store the leaf→LER
  Merkle proof at discovery time. The new rollup-origin proof preparer (`autoclaim/proof.RollupPreparer`) builds
  the LER→rollup-exit-root proof locally from `l1infotreesync` and **always** fetches a fresh leaf→LER proof from
  the source's bridge service (`GET /bridge/v1/claim-proof`) when preparing a claim — the proof is never cached.
  `SourceAwarePreparer` dispatches every claimer between this and the existing L1-origin preparer based on each
  request's `source_network`, not the claimer's destination.
- Claim submission (`claimtx`, `claimer.IsClaimed`, global index derivation) is now keyed by `source_network`
  instead of always assuming L1 origin. The claim ABI is byte-identical across L1 and L2 v2 bridge contracts, so
  `claimAsset`/`claimMessage` packing is reused unchanged.
- A destination network `0` (L1) claimer is now a legal, fully-supported configuration — like every other claimer,
  it has no `l2gersync`, and additionally it is the only claimer type with no GER-injection gate at all.
- Request identity changed from `origin_network:destination_network:deposit_count` (origin = bridged token's
  origin, which isn't unique across multiple source networks) to `source_network:destination_network:deposit_count`
  (the network the bridge exit was initiated on — this is also what the claim global index encodes).
- **claimsync** (carried from #1691, adjusted per review): the standalone claim syncer's `EVMDownloader` uses
  `rd.GetFinalizedBlockType()` again like every other syncer, keeping its blocks tracked by the reorg detector;
  the op-pp-2chains env instead sets `[ReorgDetectorL2] FinalizedBlock = "LatestBlock"` (op-reth there never
  advances the `finalized` tag). `InsertBlock` treats a duplicate insert as a successful no-op only when the
  existing row has the same hash; a mismatch now errors.
- **Security**: upgraded `google.golang.org/grpc` → v1.82.1 (GO-2026-6061) and `golang.org/x/text` → v0.39.0
  (GO-2026-5970). `.github/workflows/govulncheck.yml` is unchanged relative to `develop`.

## ⚠️ Breaking Changes

- 🛠️ **Config**: `[AutoClaim.L2ToLxBridgeDetector]` is no longer a permanently-disabled placeholder — it is a
  real, independently-enableable detector. A new `[AutoClaim.BridgeServiceFinder]` section (embeds
  `bridgeservicefinder.Config`) is required when `L2ToLxBridgeDetector.Enabled = true` **or when any enabled
  claimer has an L2 destination (`NetworkID != 0`), in either direction** — this includes a pure L1→L2 setup with
  the L2ToLx detector disabled. `[[AutoClaim.Claimers]]` now accepts `NetworkID = 0` as a valid L1-destination
  claimer (requires the L2ToLx detector to be enabled).
- 🛠️ **Config**: `RetryAfterErrorPeriod` and `MaxRetryAttemptsAfterError` on **both**
  `[AutoClaim.L1ToL2BridgeDetector]` and `[AutoClaim.L2ToLxBridgeDetector]` are **removed** — they were validated
  but never consumed. A failed detector poll is logged and retried on the next `PollInterval` tick.
- 🛠️ **Config**: Per-claimer `[[AutoClaim.Claimers]].BlockFinality` and `.InitialBlockNum` are **removed** — they
  configured the now-deleted per-claimer `l2gersync`. Autoclaim no longer consumes `[L2GERSync]`/
  `[ReorgDetectorL2]` at all (those sections remain, unchanged, for the node-global `l2gersync` that backs the
  bridge service and Aggoracle).
- ⚙️ **Operational**: L1→L2 auto claiming now **requires** `[AutoClaim.BridgeServiceFinder]` to be configured
  (with `RollupManagerAddr`) and the destination network's own bridge service to be reachable, whenever the
  destination is an L2. Previously an L1→L2 claimer only needed the shared `[L2GERSync]`/`[ReorgDetectorL2]`
  settings and its own RPC client; there was no dependency on a remote bridge service for GER readiness.
- 🔌 **API**: `GET /bridge/v1/claim-candidates` returns `{"claim_candidates": [{"bridge": {...}}], "count": N}` —
  no per-bridge Merkle proof. `to_ler`/`from_ler` define the deposit-count range as request parameters.
- 🔌 **API**: `GET /bridge/v1/injected-l1-info-leaf` now returns `404 Not Found` (previously `500 Internal Server
  Error`) when no injected global exit root covers the requested L1 info tree index yet. The Go client
  (`bridgeservice/client.Client.GetInjectedL1InfoLeaf`) surfaces this as `client.ErrNotFound` instead of a generic
  wrapped error.
- 🔌 **API**: Auto Claim request IDs (`GET /autoclaim/v1/bridges/{id}`) changed format from
  `origin_network:destination_network:deposit_count` to `source_network:destination_network:deposit_count`. For
  every pre-existing L1→L2 request these are numerically identical (`source_network` is always `0` for L1→L2) —
  but any old client that assumed "origin network" semantics for a *wrapped* token bridged L1→L2 would have seen a
  different (bug-fixed) key than before; see Migration.
- 🗄️ **Storage**: `autoclaim_request` gains `source_network`, `ler`, and `verify_block_num` columns and the
  uniqueness constraint moves to `(source_network, destination_network, deposit_count)`; a new
  `autoclaim_ler_cursor` table is added (see Migration).
- No breaking changes to `claimAsset`/`claimMessage` calldata or the admin approve/reject endpoints.

## 📋 Config Updates

```toml
[AutoClaim.L2ToLxBridgeDetector]
Enabled = true
StartL1Block = 0            # 0 = full history; else initial from_ler derived via GER at this block
PollInterval = "3s"

[AutoClaim.BridgeServiceFinder]
RollupManagerAddr = "0x..."       # required when L2ToLxBridgeDetector.Enabled = true, OR when any
                                   # enabled claimer (either direction) has an L2 destination
PollInterval = "30s"
# BlockFinality, BlockChunkSize, HealthCheckPath, HealthCheckTimeout, RequireAllHealthyOnStart all default.

[AutoClaim.BridgeServiceFinder.URLs]
# Static override map, source network ID -> bridge service base URL.
# Required to reach network 0 (L1), which is never enumerated on-chain:
# 0 = "http://static-override-l1:5577"

# Every claimer uses this same [[AutoClaim.Claimers]] shape, including an L1-destination one
# (NetworkID = 0, BridgeAddr = the L1 bridge contract, URLRPC = an L1 RPC endpoint). There are no
# BlockFinality/InitialBlockNum keys anymore (removed with the per-claimer l2gersync); an
# L2-destination claimer's readiness gate is served by [AutoClaim.BridgeServiceFinder] + the
# destination's own bridge service instead.
[[AutoClaim.Claimers]]
Enabled = true
NetworkID = 0
...
```

Full reference: `docs/autoclaim.md` (Configuration section, "Top-level keys" and "Claimer keys" tables).

## ✅ Testing

**Unit tests** (`make test-unit`, full repo green), coverage for every package this PR touched or added,
including new tests for the review fixes (detector skip-retry/destination-batching/missing-baseline behavior,
`InsertBlock` hash-mismatch rejection):

| Package | Coverage |
| --- | --- |
| `autoclaim/api` | 90.0% |
| `autoclaim/apitypes` | 96.9% |
| `autoclaim/bridgedetector` | 66.4% |
| `autoclaim/claimer` | 84.4% |
| `autoclaim/claimtx` | 89.5% |
| `autoclaim/config` | 98.4% |
| `autoclaim/policy` | 97.0% |
| `autoclaim/proof` | 81.6% |
| `autoclaim/runtime` | 60.6% |
| `autoclaim/sender` | 81.4% |
| `autoclaim/simulator` | 95.0% |
| `autoclaim/storage` | 74.0% |
| `autoclaim/storage/migrations` | 100.0% |
| `autoclaim/types` | 100.0% |
| `bridgeservice` | 92.9% |
| `bridgeservice/client` | 92.5% |
| `bridgesync` | 77.5% |
| `bridgesync/migrations` | 88.1% |
| `bridgesync/types` | 92.3% |
| `l1infotreesync` | 72.0% |
| `config` | 87.5% |

- 🤖 **Automatic (e2e, dockerized)**:
  - Single-chain `op-pp` environment, one full green run
    (`TestAutoClaimL1ToL2AllowAll`, `TestAutoClaimL1ToL2APIApprove`, `TestAutoClaimL1ToL2BasicFilter`,
    `TestAutoClaimL2ToL1AllowAll`):
    - The three `L1ToL2` tests are regressions of the existing direction, now exercising the
      destination-bridge-service GER gate instead of the old per-claimer `l2gersync`.
    - `TestAutoClaimL2ToL1AllowAll` is the **new L2→L1 direction**: `L2ToLx` detector, `RollupPreparer`, a
      `NetworkID = 0` (L1-destination) claimer, end to end against a pessimistic-proof rollup.
  - Two-chain `op-pp-2chains` environment (`AGGKIT_E2E_ENV=op-pp-2chains`), `TestAutoClaimL2ToL2AllowAll` run
    twice consecutively with `-count=1`, both green. Exercises the L2ToLx detector + `bridgeservicefinder`
    resolving both source and destination networks, the destination-bridge-service GER-injection gate, and the
    claim-time leaf-proof fetch path end to end.

## 🐞 Issues

- Closes #1665

## 🔗 Related PRs

- Builds on #1662 (`feat: autoclaim plan`, the original L1→L2 Auto Claim implementation).
- Depends on #1691 (`feat: add op-pp-2chains env and L2→L2 bridge test`) and #1692
  (`feat: add bridgeservicefinder package`) — **both merged**; this branch was rebased onto `develop` after they
  landed and the PR retargeted accordingly.

## 📝 Notes

- **Migration**: `autoclaim/storage/migrations/autoclaim0002.sql` adds `source_network` to `autoclaim_request`
  (existing rows default to `0`, correct since they're all L1-origin), replaces the `UNIQUE(origin_network,
  destination_network, deposit_count)` constraint with `UNIQUE(source_network, destination_network,
  deposit_count)` (the real claim identity — it's what the global index encodes), re-keys every pre-existing
  `autoclaim0001` row's `request_key` to the new `source:destination:deposit_count` format (a no-op value change
  for existing rows, since `source_network = origin_network = 0` for all of them), adds `ler`/`verify_block_num`
  columns, and adds a new `autoclaim_ler_cursor` table. This re-key is safe: `autoclaim0001` (introduced by
  #1662) has not shipped in any tagged release or release branch.
- **Pessimistic verifications**: no dedicated `VerifyPessimisticStateTransition` indexing is needed — the rollup
  manager emits `VerifyBatchesTrustedAggregator` (carrying the new LER) on every pessimistic verification as
  well, in all deployed contract versions, and `l1infotreesync` already indexes that event.
- **Docs**: `docs/autoclaim.md`, `docs/bridge_service.md`, and `docs/common_config.md` are fully rewritten/updated
  for all three directions and both proof-pipeline changes (architecture diagrams, sequence diagrams, config
  reference, API fields, storage tables, and the e2e test catalog).
- **API**: `GET /autoclaim/v1/bridges` gained a `source_network` (uint32) response field and list-filter query
  param, and a `ler` (hex string, empty for L1-origin requests) response field. `ler` records the source LER
  observed at detection time (used to pick the covering L1 info tree leaf); the leaf→LER proof itself is never
  part of the stored request or the API response, since it is always fetched fresh at claim time.
- **Review round (2026-07-29)**: all reviewer and Codex findings addressed in the commits after the rebase —
  claimsync finality + `InsertBlock` hash check, removal of the redundant pessimistic-event indexing, L2ToLx
  detector hardening (skip-retry, destination batching, missing baseline leaf), removal of the dead detector
  retry config keys, and the grpc/x-text security upgrades. `govulncheck.yml` was intentionally left identical
  to `develop`.


arnaubennassar:
@claude

arnaubennassar:
Thanks for the thorough review. I went through each finding against the actual code. Disposition below; commit 421ba230 addresses the one actionable item.

### 🔴 #1 Migration safety — not applied (false positive, suggested fix is incorrect)
The re-key `'0:' || destination_network || ':' || deposit_count` matches `DeriveRequestKey(source, dest, deposit)` (`autoclaim/types/types.go:324`), and `source = 0` is correct **by construction**: autoclaim0001 had only the L1→L2 detector, whose source is always L1 (network 0). There is no `source_network` column in the autoclaim0001 schema to verify against — so the assumption cannot be checked at migration time.

The suggested guard `WHERE origin_network != 0` is actively wrong: `origin_network` is the *token* origin network, which is non-zero for wrapped tokens in perfectly valid L1→L2 claims. That check would falsely `ABORT` legitimate migrations. The backport concern is a process risk, not a code defect; the existing comment already documents the invariant.

### 🟡 #2 COUNT(*) per page — by design
`GetBridgesInDepositRange` needs the total for offset validation (`calculateOffset`) and to return the total count to the API caller. This matches the existing paginated-query pattern used throughout `processor.go`.

### 🟡 #3 `maxLeafSelectionAdvances = 256` — already justified
The constant is documented at `rollup_preparer.go:20-23` (common case advances zero times; the bound only guards a pathological run). Making it configurable is scope creep with no demonstrated need.

### 🟡 #4 Cursor advancement — already intentional
As the finding itself notes, idempotent enqueue makes this safe (at-least-once). This is the intended semantics.

### 🟡 #5 Rate limiting — out of scope
The fetcher already honors `RetryAfterErrorPeriod` / `MaxRetryAttemptsAfterError`. Adding a rate limiter is a separate enhancement, not a defect in this PR.

### 🟡 #6 `rows.Close()` error surfacing — convention-consistent
Logging (not surfacing) the deferred `rows.Close()` error is the uniform repo convention — same pattern at `processor.go:483,537,634,725,1525,1558,1625`. The `cerr != nil && err == nil` pattern appears nowhere in the codebase. `rows.Close()` still runs via defer, so there is no connection leak. Changing this one site would introduce an inconsistency.

### 🟢 #7 `L2ToLx` naming — intentional
One detector handles both L2→L1 and L2→L2 (destination is the only difference); `Lx` reflects that single code path. Docs describe the user-facing directions separately for clarity.

### 🟢 #8 e2e timeout — out of scope

### 🟢 #9 Comment typo — ✅ fixed in 421ba230
`l1infotreesync/processor_verifybatches.go:23` grammar corrected.

Net: 8 of 9 findings are non-defects (by-design, already-documented, convention-consistent, or based on an incorrect premise); the typo is fixed.

joanestebanr:
@codex review

.github/workflows/govulncheck.yml
```diff
@@ -10,10 +10,47 @@ jobs:
   govulncheck_job:
     runs-on: ubuntu-latest
     name: Run govulncheck
+    env:
+      # GO-2026-4479: github.com/pion/dtls/v2 AES-GCM nonce reuse. No fixed version exists
+      # upstream (v2.2.12 is latest). Only reachable via go-ethereum's simulated backend,
+      # used in test/helpers and the tools/backward_forward_let debug utility - not part of
+      # any shipped aggkit binary's network path. Remove once pion/dtls ships a fix.
+      GOVULNCHECK_IGNORE_IDS: GO-2026-4479
```
joanestebanr: If we ignore this error, we won't know when it's been fixed. Also, even though it's currently only used for testing, that could change without anyone noticing.

WDYT?

claimsync/storage/storage.go
```diff
@@ -147,10 +148,18 @@ func (s *claimStorage) getQuerier(tx dbtypes.Querier) dbtypes.Querier {
...
+// It is idempotent: if the block already exists (PRIMARY KEY / UNIQUE constraint
+// violation on block.num), the insert is treated as a successful no-op. This guards
+// against the startup race where the L2ClaimSyncer auto-start goroutine and the
+// aggsender's SetClaimSyncerNextRequiredBlock loop both bootstrap the same block.
 func (s *claimStorage) InsertBlock(
 	_ context.Context, tx dbtypes.Querier, blockNum uint64, blockHash common.Hash,
 ) error {
 	if err := meddler.Insert(s.getQuerier(tx), "block", &blockRecord{Num: blockNum, Hash: blockHash.Hex()}); err != nil {
+		if sqliteErr, ok := db.SQLiteErr(err); ok && sqliteErr.Code == sqlite.ErrConstraint {
+			// Block already inserted by a concurrent bootstrap; nothing more to do.
```
joanestebanr: Maybe you can check that the blockHash is the same?

claimsync/claimsync.go
```diff
@@ -99,7 +99,7 @@ func NewClaimSync(
 		appender,
 		[]common.Address{cfg.BridgeAddr},
 		rh,
-		rd.GetFinalizedBlockType(),
```
joanestebanr: This param is which block is considered as finalized, doesn't make too much sense to use the target synchronization block. The rest of syncers use rd.GetFinalizedBlockType()

l1infotreesync/downloader.go
```diff
@@ -21,6 +21,11 @@ var (
 	verifyBatchesTrustedAggregatorSignature = crypto.Keccak256Hash(
 		[]byte("VerifyBatchesTrustedAggregator(uint32,uint64,bytes32,bytes32,address)"),
 	)
+	// verifyPessimisticStateTransitionSignature is emitted by pessimistic/aggchain rollups instead of
+	// VerifyBatchesTrustedAggregator. Its newLocalExitRoot field carries the rollup's new LER.
+	verifyPessimisticStateTransitionSignature = crypto.Keccak256Hash(
+		[]byte("VerifyPessimisticStateTransition(uint32,bytes32,bytes32,bytes32,bytes32,bytes32,address)"),
```
joanestebanr: After looking at the contract code, it doesn't seem that the claim that VerifyPessimisticStateTransition is emitted instead of VerifyBatchesTrustedAggregator is correct. It appears that VerifyBatchesTrustedAggregator is always emitted, and in the PP flow, VerifyPessimisticStateTransition is emitted as well.

Does this event provide any additional information?

autoclaim/config/config.go
```diff
@@ -61,16 +70,43 @@ type L1ToL2BridgeDetector struct {
...
+// L2ToLxBridgeDetector configures L2-to-Lx (rollup-origin) bridge exit discovery, covering both
+// L2-to-L1 and L2-to-L2 bridges. Enabling it requires AutoClaim.BridgeServiceFinder to be configured
+// with a valid RollupManagerAddr, since the detector resolves each source rollup's bridge service
+// through it.
+type L2ToLxBridgeDetector struct {
 	Enabled bool `mapstructure:"Enabled"`
+	// StartL1Block is the L1 block used to derive a newly discovered source network's initial LER
+	// cursor (via the GER at that block). 0 means full history (from_ler omitted on first fetch).
+	StartL1Block               uint64            `mapstructure:"StartL1Block"`
+	PollInterval               cfgtypes.Duration `mapstructure:"PollInterval"`
```
joanestebanr: There are an object `RetryPolicyGenericConfig` to define retry policies more acurated (the same for L1ToL2BridgeDetector)

claimsync/claimsync.go
arnaubennassar: Fixed in ad933339 — back to `rd.GetFinalizedBlockType()`, like the rest of the syncers. For context, the change had been made because op-reth in the `op-pp-2chains` kurtosis env never advances the `finalized` tag, which permanently stalled the L2 claim syncer's safe-zone check at block 0 — but as you point out it misused the param, and it also meant every downloaded block was marked finalized and never tracked by the reorg detector. The env problem is now handled where it belongs: `[ReorgDetectorL2] FinalizedBlock = "LatestBlock"` in that env's configs (no L2 reorgs there, so it's safe).

l1infotreesync/downloader.go
arnaubennassar: You're right — I checked both the current `AgglayerManager.sol` and the previously deployed `PolygonRollupManagerPessimistic.sol`: the pessimistic flow emits `VerifyBatchesTrustedAggregator` too ("Same event as verifyBatches to support current bridge service to synchronize everything"), with the same `newLocalExitRoot`. The extra fields of `VerifyPessimisticStateTransition` (pessimistic roots, l1InfoRoot) aren't anything l1infotreesync stores, so indexing it only produced a duplicate VerifyBatches event per pessimistic verification (absorbed by the processor's same-ExitRoot dedup). Removed the handler entirely in 41a21ce7 and fixed the comments/docs that claimed pessimistic rollups don't emit `VerifyBatchesTrustedAggregator`.

claimsync/storage/storage.go
arnaubennassar: Done in 58e76d3a — the no-op now only applies when the existing row has the same hash; a constraint violation with a different hash is returned as an error.

autoclaim/config/config.go
arnaubennassar: Good catch — digging into it, these two fields (on both detectors) were validated but never consumed anywhere: the runtime doesn't wire them, and a failed poll is simply logged and retried on the next `PollInterval` tick. Rather than wrapping dead knobs in a richer config object, I removed them in 3dd6fd31 (structs, defaults, docs, env configs). If we ever add intra-poll retries to the detectors, `common.RetryPolicyGenericConfig` is the shape I'd use.

.github/workflows/govulncheck.yml
arnaubennassar: Agreed — reverted in aca4c8b7: `govulncheck.yml` is back to exactly the `develop` version (no ignore mechanism), so any finding fails the job visibly. Separately I upgraded `google.golang.org/grpc` to v1.82.1 (GO-2026-6061) and `golang.org/x/text` to v0.39.0 (GO-2026-5970), the two findings that were failing the job on this branch; GO-2026-4479 (pion/dtls) is no longer reported by govulncheck at all.

autoclaim/bridgedetector/l2_to_lx.go
```diff
@@ -0,0 +1,662 @@
...
+	// next LER update is observed the fetch still uses from_ler = <old cursor>, which covers every
+	// candidate missed in between. A hard error above returns before this point, leaving the cursor
+	// unchanged so the whole window is retried.
+	nextCursor := autoclaimtypes.BridgeCursor{
+		FromBlock: fromBlock,
+		ToBlock:   toBlock,
+		BlockNum:  toBlock,
+		BlockPos:  0,
+	}
+	if err := w.cursorStore.SaveBridgeCursor(ctx, w.cursorName, nextCursor, w.now()); err != nil {
```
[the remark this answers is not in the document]
arnaubennassar: Fixed in 5e586357. The block-window cursor is now held just before the verify row of any source skipped for a transient reason (finder miss or `ErrCandidatesNotSynced`), so the row is re-observed and the source retried on every later poll — even if it never publishes another LER. Covered by `TestL2ToLxRetriesSkippedSourceOnLaterPollWithoutNewLER` (skip → recover → process, no new LER in between) and a window-start edge-case test.

autoclaim/bridgedetector/l2_to_lx.go
```diff
@@ -0,0 +1,662 @@
...
+) ([]ClaimCandidate, error) {
+	all := make([]ClaimCandidate, 0)
+	// The bridge service /claim-candidates endpoint uses 1-based pagination (page_number must be
+	// > 0, mirroring /bridges); page 0 is rejected with HTTP 400. Start at page 1.
+	pageNumber := uint32(1)
+	for {
+		query := ClaimCandidatesQuery{
+			URL:                   url,
+			DestinationNetworkIDs: destinationIDs,
+			FromLER:               fromLER,
```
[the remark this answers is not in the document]
arnaubennassar: Fixed in 5e586357 — `fetchAllCandidates` now partitions the destination filter into batches of at most `bridgeservice.MaxNetworkIDs` and pages through each batch. Covered by `TestL2ToLxBatchesDestinationNetworkIDs` (7 destinations → two requests of 5 + 2).

autoclaim/bridgedetector/l2_to_lx.go
```diff
@@ -0,0 +1,662 @@
...
+// is used; a zero LER (the network had no LER yet at that block) also requests the full history.
+func (w *L2ToLx) initialFromLER(ctx context.Context, sourceID uint32) (*common.Hash, error) {
+	if w.startL1Block == 0 {
+		return nil, nil
+	}
+
+	leaf, err := w.source.GetLatestL1InfoLeafUntilBlock(ctx, w.startL1Block)
+	if err != nil {
+		return nil, fmt.Errorf("get latest l1 info leaf until block %d for source %d: %w",
+			w.startL1Block, sourceID, err)
```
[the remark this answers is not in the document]
arnaubennassar: Fixed in 5e586357 — `initialFromLER` now treats `l1infotreesync.ErrNotFound` (StartL1Block predates the first L1 info tree leaf) like the documented zero-LER case: `from_ler` omitted, full history fetched. `ErrBlockNotProcessed` still errors, since that genuinely means "retry when synced". Covered by `TestL2ToLxInitialCursorLeafNotFoundOmitsFromLER`.

joanestebanr (approved):
LGTM

PR #1729, opened 2026-07-19
fix(exit-certificate): re-seek debug_accountRange to avoid truncated state dump

krlosMata:
## Problem

The cdk-erigon RPC (behind a load balancer) intermittently returns a valid HTTP 200 `debug_accountRange` page with an **empty `next` cursor before the trie ends**. Step A trusted that as "done", so the state dump was silently truncated (observed on polygon zkEVM mainnet: **735,744 of 899,166 accounts**). The dropped accounts under-collected native-ETH holders, causing Step B/C/F to under-report balances (~152.7 ETH native shortfall vs the LBT).

## Fix

For the erigon dialect, the dump no longer treats an empty cursor as authoritative: it re-seeks strictly past the highest key seen and only stops once a fresh re-seek genuinely returns no new account (retried a few times to ride out transient empty pages). The geth path keeps the plain cursor semantics.

## Verification

Against mainnet the dump now returns the full **899,166 accounts** (Step A total 908,296), matching an independent full-trie walk that sums to exactly the **200,000,000 ETH** genesis reserve.

Made with [Cursor](https://cursor.com)


joanestebanr:
The  [feature/exit-certificate-tool](https://github.com/agglayer/aggkit/tree/feature/exit-certificate-tool)  branch has already been merged into `develop`. It was kept to preserve the history, so the PR should target `develop`

test/e2e-exit_certificate/20-prepare_network.sh
```diff
@@ -115,7 +115,13 @@ log_info "ERC-20 scenario state saved to: $ERC20_STATE_FILE"
...
-log_info "⏳ Wait for the certificate to be settled"
-run_quiet "${EXIT_CETIFICATE_SCRIPT_DIR}/agglayer_certificate_status.sh" --wait
+# The L2 bridge's current root is the exact local exit root the agglayer must settle before the
+# exit certificate can be generated (its AET-11 guard requires L2 bridge getRoot() == settled LER).
+# No more L2 bridge exits happen after this point, so the root is stable. Waiting for this specific
+# root closes the race where the aggsender has not yet submitted the new certificate — a plain
+# "no pending certificate == settled" check would return prematurely and leave the bridge LER ahead.
+EXPECTED_LER=$(cast call --rpc-url "$L2_RPC_URL" "$BRIDGE_ADDR" "getRoot()(bytes32)" | cut -f1 -d' ')
+log_info "⏳ Wait for the certificate settling L2 bridge root ${EXPECTED_LER}"
+run_quiet "${EXIT_CETIFICATE_SCRIPT_DIR}/agglayer_certificate_status.sh" --wait --expected-ler "$EXPECTED_LER"
```
joanestebanr, line 125: Why do you need this new wait method `--expected-ler`?

krlosMata: sanity check. This is the test that was failing. IA can use this flag as a sanity check localy

joanestebanr (dismissed):
LGTM

joanestebanr (approved):
LGTM

PR #1730, opened 2026-07-20
feat(tools): force_ger_update — force periodic L1 GER updates

arnaubennassar:
## 🔄 Changes Summary
- Add `force_ger_update`: a standalone, long-running CLI tool that guarantees the **L1 Global Exit
  Root (GER) is updated at least every configurable `X`**. It watches L1 for `UpdateL1InfoTree`
  events and, if none happens organically within the window, sends a `bridgeMessage` with
  `forceUpdateGlobalExitRoot = true` to force a new L1 info root.
- **Motivation (OP-FEP / DA provability):** for aggchains running OP-FEP, every GER update on L1
  appends a leaf to the L1 info tree that embeds an **L1 block hash**, and the aggchain proof uses
  that block hash to attest to what happened on L1 — **including data availability (DA)**. Anything
  posted on L1 *after* the last L1 info root update is covered by no block hash inside any L1 info
  root and therefore **cannot be proven** until a new update lands. If DA is posted after the last
  update, certificate progress stalls until an organic GER update happens — with no bound on the
  wait. This tool bounds that wait to `X`.
- Design: no syncer / reorg detector / event DB — boot scan via chunked `FilterLogs`, then WS
  `WatchUpdateL1InfoTree` (if `L1WSURL` set) or `FilterLogs` polling; sends via the standard
  `zkevm-ethtx-manager`; keys via `go_signer` (local keystore or GCP/AWS KMS).
- Wired into `Makefile` (`build-force_ger_update`, added to `build-tools`) and shipped in the
  Docker image at `/usr/local/bin/force_ger_update`.
- CI: the Tier-2 real-network e2e (`TestForceGERUpdateE2E`) now runs in CI on its own dedicated,
  isolated job/runner (`test-go-e2e-force-ger-update` in `.github/workflows/test-go-e2e.yml`)
  instead of being permanently skipped, so it is exercised automatically without risking the shared
  `test-go-e2e` env/job.

## ⚠️ Breaking Changes
- 🛠️ **Config**: none (new standalone tool config; no existing config changed).
- 🔌 **API/CLI**: none (additive — new tool binary only).
- 🗑️ **Deprecated Features**: none.

## 📋 Config Updates
- 🧾 New standalone TOML with a `[ForceGERUpdate]` root section (+ nested `[…EthTxManager]` /
  `[…EthTxManager.Etherman]`). See `tools/force_ger_update/example-config.toml` and the README's
  configuration reference. Mandatory & validated: `L1URL`, `GlobalExitRootManagerAddr`, `BridgeAddr`,
  `DestinationNetwork` (≠ 0), `MaxTimeWithoutGERUpdate` (> 0), `CheckInterval` (> 0),
  `FilterLogsChunkSize` (> 0), and `EventPollInterval` (> 0, poll mode only).

## ✅ Testing
- 🤖 **Automatic**:
  - **Tier-1** (unit + integration on the simulated backend, run by `make test-unit`): config
    load/validate, monitor (boot scan + watch/poll), sender (calldata/selector, statuses, dry-run),
    main loop (stale-on-boot, event-reset, in-flight guard, clean shutdown) — `go test -race
    ./tools/force_ger_update/...`; plus a simulated-backend integration test
    (`integration_test.go`) exercising both watch and poll modes and asserting real on-chain
    `UpdateL1InfoTree` events and the `0x240ff378` selector.
  - **Tier-2** (isolated real-network e2e CI job): `TestForceGERUpdateE2E` (`test/e2e/`) execs the
    built binary against the docker-compose `op-pp` env and proves causality — the new GER update's
    tx `From` == tool sender and input starts with `0x240ff378`. It is env-gated
    (`RUN_FORCE_GER_UPDATE_E2E=true`) so it stays skipped in the shared `make test-e2e` / main
    `test-go-e2e` CI job (avoiding interference with that job's shared post-test bridge health
    check), and instead runs automatically in CI on its own dedicated runner —
    `test-go-e2e-force-ger-update`, a new job in `.github/workflows/test-go-e2e.yml` that reuses the
    existing docker-image build/pull pipeline and calls `make test-e2e-force_ger_update`. That
    target sets both `RUN_FORCE_GER_UPDATE_E2E=true` and `E2E_SKIP_POSTTEST_BRIDGE_CHECK=true` (the
    latter opts the isolated job out of `TestMain`'s shared post-test L1↔L2 bridge health check,
    since forcing GER updates legitimately leaves that check unhealthy — same root cause as the
    existing `remove_ger` e2e skips). Locally: `make test-e2e-force_ger_update` (docker-compose env
    must be up).
- 🖱️ **Manual**:
  - `DryRun = true` smoke test against a live env logs the boot-derived last-GER-update age without
    sending a transaction.

## 🐞 Issues
- N/A

## 🔗 Related PRs
- [Optional]

## 📝 Notes
- Timer semantics: `lastGERUpdate` is an L1 block timestamp; elapsed is measured against the tool's
  wall clock. Clock skew is immaterial next to a minutes-to-hours threshold and is safe in both
  directions (a block timestamp slightly ahead of wall-clock yields a negative elapsed → the tool
  simply waits longer; it never over-fires and never panics).
- The timer is reset only by an *observed* `UpdateL1InfoTree` event, never by the send completing;
  in poll mode this permits at most one redundant (harmless) send per window if the observation lags
  the mine. Sends are serialized by an atomic in-flight guard, so they are never concurrent or
  unbounded.
- `gasOffset` is hardcoded to 0; gas is controlled via the ethtxmanager config
  (`ForcedGas`/`GasPriceMarginFactor`/`MaxGasPriceLimit`).
- The Tier-2 e2e's isolation contract (`RUN_FORCE_GER_UPDATE_E2E`, `E2E_SKIP_POSTTEST_BRIDGE_CHECK`)
  is documented in the README's "How to test" section and in code comments at both env-var
  check-sites.


joanestebanr (approved):
LGTM

PR #1732, opened 2026-07-21
fix(l2gersync): recover automatically when an invalid GER is removed on-chain

### LINKED ISSUE #1637: L2GERsync shoul not get stucked after invalid ger (or at least recover if there is remove ger a posteriory)

### LINKED ISSUE #1697: Restore remove GER e2e tests

arnaubennassar:
## 🔄 Changes Summary
Make l2gersync **recover automatically** when an invalid GER that was injected on the L2 sovereign GER manager is later removed on-chain.

- **Root cause**: while an invalid GER is present on L2 but absent from the L1 info tree, the sovereign insert appender returned an error and the generic sync layer retried it forever (or `log.Fatalf`'d, crash-looping, on a non-negative `MaxRetryAttemptsAfterError`). The existing removal check never converged: it was gated behind `IsUpToDate`, and a latent nil-pointer deref could panic the downloader.
- **Fix (`l2gersync` + surgical `sync/`)**: on each retry the appender re-checks removal **at latest**, decoupled from `IsUpToDate`. It skips the invalid insert **iff BOTH** a `UpdateRemovalHashChainValue` removal-event scan (insert block→latest) finds ≥1 event **AND** `GlobalExitRootMap[ger] == 0`. Until then it stays **blocked and noisy** (unchanged, intentional — an invalid GER is an incident operators must notice) but is now guaranteed **never fatal** for this class via a new `sync.ErrRetryForeverNonFatal` sentinel. Nil-guards added on the contract reads. The retry loop also now exits cleanly on context cancellation (reorg reset / shutdown).
- **Observability**: bridgeservice `GET /bridge/v1/sync-status` gains an `l2_ger_info` object (`is_active`, `last_processed_block`) so l2gersync progress is externally visible (`is_active=false` on L1 bridgeservice). Swagger regenerated.
- **E2E**: the 5 previously-skipped remove-GER tests are re-enabled with new `/sync-status`-based stall + unstuck + still-alive assertions, split across three sub-20-min CI matrix groups on `op-pp` (`removeger-fast`, `removeger-b1`, `removeger-b2`) with the default group given an explicit exclusion regex so they don't run twice.

## ⚠️ Breaking Changes
- 🛠️ **Config**: none — no new/changed TOML options.
- 🔌 **API/CLI**: no public breaking changes. Internal only: the `bridgeservice.L2GERSyncer` interface gains `GetLastProcessedBlock(ctx)`; `tools/remove_ger.ExecuteRecovery` now returns `(*RecoveryResult, error)`. The `/sync-status` response is **additive** (`l2_ger_info`).
- 🗑️ **Deprecated Features**: none.

## 📋 Config Updates
- No config changes.

## ✅ Testing
- 🤖 **Automatic**:
  - Unit: appender stays non-fatal (`ErrRetryForeverNonFatal`) while the invalid GER is present and skips the insert once removal is observed; processor delete of a never-inserted row is a no-op; reorg-out with no removal event never falsely unsticks; `/sync-status` handler (l2gersync present / nil / processor-error) + client decode. `make test-unit` green.
  - E2E (`op-pp`): all four `TestRemoveGER_*` scenarios pass **twice consecutively** locally — each confirms l2gersync stalls at the invalid GER (`last_processed_block` below the insert block while the L2 head advances), then unsticks and processes past the actual `removeGlobalExitRoots` block after recovery, then serves fresh state (`/injected-l1-info-leaf`). Post-test bridge health check green.
- 🖱️ **Manual**: follow `docs/remove_ger_runbook.md`; verify recovery via `/sync-status` (`l2_ger_info.last_processed_block` advances past the removal block).

## 🐞 Issues
- close #1637 
- close #1697 

## 🔗 Related PRs
- Was stacked on **#1705** (`feat/autoclaim-l2-lx`, itself on top of **#1691**); both have merged into `develop`, so this branch has been rebased onto `develop` and now targets it directly. The diff is the 8 l2gersync/`/sync-status`/e2e commits only.

## 📝 Notes
- **Halt semantics preserved**: the syncer stays blocked (noisy, never fatal) while the invalid GER is present and unsticks only on a real on-chain removal. Requiring the removal **event log** (not the zeroed map alone) is what makes a bare insert-block reorg-out never falsely unstick.
- **Adversarial review** (branch): one Medium finding fixed (retry loop now honors ctx cancellation); false-unstick, reorg-safety, read-API consistency, liveness, and `/sync-status` nil/race surfaces cleared.
- **`TestGenerateInvalidGER`** skips on dev machines whose host `cast`/foundry binary can't open outbound TCP to Docker-published ports (a machine-local defect — curl/Go reach the same endpoint); it runs fully in CI where foundry is installed fresh.
- **Two pre-existing flakes** (verified on `develop` at identical rates, unrelated to this change): `claimsync/TestClaimCalldata` (ephemeral-Geth RPC EOF) and `l2gersync/TestL2EVMGERReader_GetRemovedGERsForRange` (~20% `t.TempDir` cleanup race from a leaked `EVMDriver.Sync` goroutine). A clean `make test-unit` pass was obtained.

https://claude.ai/code/session_01U6PnumJ6vGNSVz4HSfTNg2


l2gersync/evm_downloader_sovereign.go
```diff
@@ -148,40 +148,40 @@ func (d *downloaderSovereign) buildAppender(
...
-				}
+
+			if d.isGERRemovedFromL2(ctx, b.Num, gerHash) {
+				log.Infof("GER %s got removed from L2, skipping stale insert event", gerHash.Hex())
+				return nil
 			}
+
 			return fmt.Errorf("failed to fetch l1 info tree for global exit root %s: %w",
-				common.Hash(insertGEREvent.NewGlobalExitRoot).Hex(), err)
+				gerHash.Hex(), sync.ErrRetryForeverNonFatal)
```
joanestebanr: Maybe it makes sense to keep the same behaviour as before, an error goes trought `RetryHandler` , and you can decide in the configuration whether to retry indefinitely or emit a `Fatal` error after _n_ retries

l2gersync/evm_downloader_sovereign.go
```diff
@@ -148,40 +148,40 @@ func (d *downloaderSovereign) buildAppender(
...
+			gerHash := common.Hash(insertGEREvent.NewGlobalExitRoot)
 			log.Errorf("failed to fetch l1 info tree for global exit root %s (block: %d): %v",
-				common.Hash(insertGEREvent.NewGlobalExitRoot).Hex(), b.Num, err)
+				gerHash.Hex(), b.Num, err)
 			ctx := context.Background()
+
+			// IsUpToDate is informational only from here on: it no longer gates the removal check
+			// below, since it is independent of whether l1infotreesync has caught up to L1's
+			// finalized head (see plan S2 §1).
 			isUpToDate, upToDateErr := d.l1InfoTreeSync.IsUpToDate(ctx, d.l1Client)
```
joanestebanr: Maybe it makes sense to remove  `l1InfoTreeSync`enterely, since is not used in the case of `sovereign chain`

joanestebanr (commented):
T

joanestebanr (approved):

PR #1780, opened 2026-08-10
fix(bridgesync): index UPPER(from_address) and destination_network on bridge table

### LINKED ISSUE #1779: [BUG] GET /bridge/v1/bridges with from_address filter full-scans the bridge table and times out
# Bug Report

## Description

`GET /bridge/v1/bridges` degrades to a full table scan of the `bridge` table whenever the `from_address` query parameter is supplied. On a bridge table of ~1.1M rows this reliably exceeds the request timeout and the endpoint returns a gateway error, regardless of `page_size` or how many rows actually match the filter.

This is the same class of problem as the one fixed for claims in #1774 (`fix(claimsync): prevent GetClaimsPaged timeout on large claim tables`): query cost scales with total table size rather than with page size or result-set size.

### Root cause

Two compounding issues.

**1. No usable index for the `from_address` predicate.**

The complete index set on `bridge` is:

| Index | Migration |
| --- | --- |
| `idx_bridge_deposit_count_desc (deposit_count DESC)` | `bridgesync0005.sql:7` |
| `idx_bridge_block_num_block_pos_desc (block_num DESC, block_pos DESC)` | `bridgesync0006.sql:7` |
| `idx_bridge_txn_sender (txn_sender)` | `bridgesync0007.sql:5` |
| `idx_bridge_block_num_block_pos_asc (block_num ASC, block_pos ASC)` | `bridgesync0009.sql:6` |

There is no index on `from_address`. There is also none on `destination_address` or on `destination_network`, the latter backing the `network_ids` filter.

Additionally, even adding a plain index would not help, because the predicate is non-sargable — `bridgesync/processor.go:585`:

```go
clauses = append(clauses, fmt.Sprintf("UPPER(from_address) = UPPER($%d)", paramIndex))
```

SQLite cannot use a plain index on `from_address` through the `UPPER()` call. This requires either an expression index on `UPPER(from_address)`, or normalising the stored address casing at write time so the `UPPER()` wrapper can be dropped from the query.

**2. An unconditional `COUNT(*)` on every paged request.**

`GetBridgesPaged` (`bridgesync/processor.go:506-551`) issues two queries. The first is `GetTotalNumberOfRecordsWithParams` (`bridgesync/processor.go:1476`):

```go
query := "SELECT COUNT(*) AS count FROM " + tableName + whereClause + ";"
```

With the `from_address` filter applied this is a full scan of `bridge` on every single request, before the page query has even run. The page query then walks `idx_bridge_deposit_count_desc` evaluating `UPPER(from_address)` per row until it fills `LIMIT + OFFSET`.

## To Reproduce

Against a deployment with a large `bridge` table (~1.1M rows):

1. `GET /bridge/v1/bridges?network_id=0&page_size=1` — returns promptly.
2. `GET /bridge/v1/bridges?network_id=0&deposit_count=<n>` — returns promptly (`deposit_count` is indexed).
3. `GET /bridge/v1/bridges?network_id=0&from_address=<any address>&page_size=1` — times out.
4. Repeat step 3 with an address that matches **zero** rows — also times out.

Step 4 is the diagnostic: the cost is independent of the result-set size, and reducing `page_size` to 1 does not help. Both are consistent with a full table scan rather than a large response.

## Expected behavior

Filtering by `from_address` should be index-backed and return in time comparable to the `deposit_count` filter. Query cost should scale with page size, not with total table size.

## Additional context

Related, and probably worth addressing in the same pass:

- **`sync-status` pays the same `COUNT(*)`.** `populateNetworkSyncInfo` (`bridgeservice/bridge.go:1304`) calls `GetBridgesPaged(ctx, 1, 1, nil, nil, "")` purely to obtain the total row count for `synchronized_deposit_count`, doing a full-table count per network on every call. A cheaper count path would speed up this endpoint noticeably.

- **`OFFSET`-based pagination is linear in page number.** The page query is `ORDER BY deposit_count DESC LIMIT ? OFFSET ?` (`bridgesync/processor.go:785-820`). Response time grows steadily with `page_number` even when no filter is applied — deep pages are roughly an order of magnitude slower than the first page. Keyset pagination on `deposit_count` would remove this.

- **Timeout layering.** `DBQueryTimeout` and the bridge service `ReadTimeout` both default to `5m` (`config/default.go:45,157`), which is far longer than typical ingress/gateway timeouts. The client receives a gateway error while the scan continues server-side.

- **`GetBridgesInDepositRange` (`bridgesync/processor.go:603`) and `GetBridgesByContent` (`bridgesync/processor.go:1580`)** use the same `COUNT(*)` + `OFFSET` pattern and are likely affected similarly.

- **`bridge_archive` (`bridgesync0014.sql:40`) has no indexes at all**, if it is ever queried on these paths.

### Suggested fix

A new `bridgesync` migration in the shape of `claimsync0002.sql` from #1774:

```sql
CREATE INDEX IF NOT EXISTS idx_bridge_from_address_upper ON bridge (UPPER(from_address));
CREATE INDEX IF NOT EXISTS idx_bridge_destination_network ON bridge (destination_network);
```

Preferably combined with normalising `from_address` casing on write so the expression index can be replaced by a plain one, plus making the `COUNT(*)` avoidable and moving paging to a keyset scheme.

arnaubennassar:
## 🔄 Changes Summary
- Fixes `GET /bridge/v1/bridges` full-scanning the `bridge` table (~1.1M rows) whenever `from_address` is supplied, timing out regardless of `page_size` or how many rows actually match.
- Root cause: `buildBridgesFilterClause` (`bridgesync/processor.go:586`) filters via `UPPER(from_address) = UPPER($n)`, and there is no index — plain or expression — usable through the `UPPER()` wrapper. `GetTotalNumberOfRecordsWithParams` (`bridgesync/processor.go:1476`) runs the same predicate as an unconditional `COUNT(*)` on every paged request, so both the count and the page query degrade to a full table scan.
- Fix: new `bridgesync0016` migration adding `idx_bridge_from_address_upper` (expression index on `UPPER(from_address)`, matching the query predicate exactly) and `idx_bridge_destination_network` (backing the `network_ids` filter, which also had no index).

## ⚠️ Breaking Changes
- 🛠️ **Config**: None
- 🔌 **API/CLI**: None (endpoint behavior/response semantics unchanged)
- 🗑️ **Deprecated Features**: None

## 📋 Config Updates
- 🧾 **Diff/Config snippet**: None — new DB migration `bridgesync0016` applies automatically on startup

## ✅ Testing
- 🤖 **Automatic**: New `TestMigration0016_IndexesExist` asserts both indexes exist after running migrations. Existing `bridgesync`/`bridgesync/migrations` suites pass unchanged.
- 🖱️ **Manual**: `go build ./...`, `go test ./bridgesync/...` (all green), `make lint` on `bridgesync/migrations/...` (0 issues). Verified with `EXPLAIN QUERY PLAN` against an in-memory SQLite DB that both `SELECT ... WHERE UPPER(from_address) = UPPER(?)` and `SELECT COUNT(*) ... WHERE UPPER(from_address) = UPPER(?)` switch from a table scan to `SEARCH bridge USING INDEX idx_bridge_from_address_upper`.

## 🐞 Issues
- Closes #1779

## 🔗 Related PRs
- Same pattern as #1774 (`fix(claimsync): prevent GetClaimsPaged timeout on large claim tables`)

## 📝 Notes
- Scoped to the index fix the issue's "Suggested fix" section calls out as the core problem. The issue also lists follow-on ideas (avoiding `COUNT(*)` on `sync-status`, keyset pagination, timeout layering, `GetBridgesInDepositRange`/`GetBridgesByContent`, `bridge_archive` indexes) as separate, lower-priority work worth tracking independently rather than bundling into this fix.

### DIFF
--- a/bridgesync/migrations/migrations_test.go
+++ b/bridgesync/migrations/migrations_test.go
@@ -752,6 +752,25 @@ func TestMigration0015(t *testing.T) {
 	require.False(t, tableExists("unset_claim"), "unset_claim table should be dropped by migration 0015")
 }
 
+func TestMigration0016_IndexesExist(t *testing.T) {
+	dbPath := path.Join(t.TempDir(), "bridgesyncTest0016.sqlite")
+	err := RunMigrations(dbPath)
+	require.NoError(t, err)
+
+	database, err := db.NewSQLiteDB(dbPath)
+	require.NoError(t, err)
+	defer database.Close()
+
+	for _, indexName := range []string{"idx_bridge_from_address_upper", "idx_bridge_destination_network"} {
+		var name string
+		err := database.QueryRow(
+			`SELECT name FROM sqlite_master WHERE type='index' AND name=?`, indexName,
+		).Scan(&name)
+		require.NoError(t, err, "index %s should exist", indexName)
+		require.Equal(t, indexName, name)
+	}
+}
+
 func TestMigrationsDown(t *testing.T) {
 	dbPath := path.Join(t.TempDir(), "bridgesyncTestDown.sqlite")
 	err := RunMigrations(dbPath)
--- a/bridgesync/migrations/bridgesync0016.sql
+++ b/bridgesync/migrations/bridgesync0016.sql
@@ -1,0 +1,10 @@
+-- +migrate Down
+DROP INDEX IF EXISTS idx_bridge_from_address_upper;
+DROP INDEX IF EXISTS idx_bridge_destination_network;
+
+-- +migrate Up
+-- Backs the "UPPER(from_address) = UPPER($n)" predicate in buildBridgesFilterClause
+-- (bridgesync/processor.go). A plain index on from_address cannot be used through the
+-- UPPER() wrapper, so this is an expression index matching it exactly.
+CREATE INDEX IF NOT EXISTS idx_bridge_from_address_upper ON bridge (UPPER(from_address));
+CREATE INDEX IF NOT EXISTS idx_bridge_destination_network ON bridge (destination_network);


taylanpince (approved):

bridgesync/migrations/migrations_test.go
```diff
@@ -752,6 +752,25 @@ func TestMigration0015(t *testing.T) {
...
+func TestMigration0016_IndexesExist(t *testing.T) {
+	dbPath := path.Join(t.TempDir(), "bridgesyncTest0016.sqlite")
+	err := RunMigrations(dbPath)
+	require.NoError(t, err)
+
+	database, err := db.NewSQLiteDB(dbPath)
+	require.NoError(t, err)
+	defer database.Close()
+
+	for _, indexName := range []string{"idx_bridge_from_address_upper", "idx_bridge_destination_network"} {
```
hadjiszs, line 764: @arnaubennassar FWIW I wanted to run a quick claude review before approving fully blindly, I dont have context so cannot tell if it's a legit P0, but sharing just in case

> **P0: `idx_bridge_destination_network` hijacks the planner and cancels this PR's own fix**
>
> Was the second index measured? With no `ANALYZE` anywhere in aggkit, SQLite assumes an indexed equality matches ~10 rows, so it drops `idx_bridge_deposit_count_desc` (which serves the `ORDER BY` and lets `LIMIT` exit early) in favour of `destination_network` plus `USE TEMP B-TREE FOR ORDER BY`. On a 1.1M-row fixture that regresses `?network_ids=` paging from 0.000s to 0.185s and `/claim-candidates` (where `destination_network_ids` is mandatory) from 0.000s to 0.072s; worse, when `network_ids` accompanies `from_address` it steals the plan from the new expression index, leaving the target endpoint at 0.134s instead of 0.000s. The `UPPER(from_address)` index alone fixes the reported bug completely (0.192s to 0.000s), so this is a pure deletion.
>
> Also drop `"idx_bridge_destination_network"` from the loop at `migrations_test.go:764`. If the `network_ids` filter should be indexed, the candidate is a composite `(destination_network, deposit_count DESC)`, but it still beats the expression index on the combined-filter count (0.226s vs 0.000s), so it needs its own measurement in a follow-up.

arnaubennassar: will fix on this PR https://github.com/agglayer/aggkit/pull/1784

PR #1784, opened 2026-08-10
fix: bridgetracker API.md drift, fail-fast on unset L1GlobalExitRootAddress, unlimited default rate limit (#1781, #1782, #1783)

### LINKED ISSUE #1781: bridgetracker: API.md wire-format drift — enums serialize as bare strings, undocumented fields/nesting, expected_duration never emitted
## Summary

`docs/bridgetracker/API.md` (at `v0.11.0-rc4`, rev `27831b4a`) states:

> All the enums are serialized as numeric values, and each enum field has a companion `<field>_string` field with its string representation.

The live tracker responses only follow this convention for **two** fields. A client implemented from the doc will fail to parse real responses.

## Observed wire format vs. doc

Verified against live captures from a kurtosis devnet running `ghcr.io/agglayer/aggkit:0.11.0-rc4` (`--components=proxy,tracker`), covering all three bridge typologies through their full lifecycle (26 captured `TrackingData` snapshots: registered → resolved → each step transition → finished, plus error shapes):

| Field | API.md says | Actually on the wire |
|---|---|---|
| `TrackingData.tracking_status` | int (0–3) + `tracking_status_string` | bare string only (`"registered"`/`"running"`/`"error"`/`"finished"`), no numeric field, no `_string` companion |
| `BridgeStatus.bridge_type` | int (0–2) + `bridge_type_string` | bare string only (`"L1->L2"`, `"L2->L1"`, `"L2->L2"`) |
| step entry enum | `step` int (0–7) + `step_string` | field is named **`step_name`** and is a bare string (`"WaitingGERUpdate"`, …); no `step` int, no `step_string` |
| step entry `status` | int (0–3) + `status_string` | bare string only (`"pending"`/`"inProgress"`/`"done"`/`"error"`) |
| `ErrorStep.error_type` | int + `error_type_string` | ✅ matches the doc exactly |
| `CertificateData.status` | int + `status_string` | ✅ matches the doc exactly |

Additional undocumented structure observed consistently:

- `BridgeStatus` is not flat: `bridge_leaf_type`, origin/destination, amount, and deposit count nest under an **`event`** object; `bridge_status` also carries `block_timestamp`.
- Each `all_steps[]` entry carries a `step_index` int (not in the doc).
- Optional per-step fields (`start_date`, `end_date`, `result`, `error`, and `CertificateData.settlement_tx_hash` pre-Candidate) are **omitted** when not applicable, not serialized as `null` (the doc's examples suggest nullable fields).
- `expected_duration` is documented (and appears in the doc's examples) but was **never present in any captured response** across all typologies and lifecycle stages.

## Why it matters

The string values themselves are stable and self-describing, so a client written against the *actual* wire works fine — but a client written against API.md (numeric enums, `step` field, non-null optionals, `expected_duration`) will misparse or crash on real responses. We had to reverse the format from live captures while building an SDK client.

## Suggested fix

Either update API.md to match the serializer (bare-string enums, `step_name`, `event` nesting, omitted-optionals, drop or implement `expected_duration`), or change the serializer to match the doc — but consistently, since `error_type` and certificate `status` already follow the documented convention while their siblings don't.

## Environment

- aggkit `v0.11.0-rc4` (git rev `27831b4a`), image `ghcr.io/agglayer/aggkit:0.11.0-rc4`
- aggkit-proxy with `--components=proxy,tracker`, kurtosis-cdk 2-L2 devnet, tracker queried via REST `GET /tracker/v1/network/{network_id}/tx/{tx_hash}`

### LINKED ISSUE #1782: bridgetracker: unset [Tracker].L1GlobalExitRootAddress silently stalls every L1→L2 route at WaitingGERUpdate (zero-address log filter, no fallback, reference recipe affected)
## Summary

If `[Tracker].L1GlobalExitRootAddress` is not set in the aggkit-proxy config, it defaults to the zero address and the bridge tracker's `WaitingGERUpdate` step — step 0 of **every** L1→L2 bridge route — can never resolve. There is no error, no warning, and no on-chain discovery fallback: tracking just sits at step 0 forever. The reference recipe `proxy/scripts/configuration_based_on_kurtosis.sh` doesn't set this field either, so a tracker stood up from aggkit's own example is silently broken for all L1→L2 bridges.

## Root cause (source, at `v0.11.0-rc4` / rev `27831b4a`)

- `bridgetracker/domain/resolve_step_waiting_ger_update.go` → `WaitingGERUpdateResolver` calls `GERSource.FindFirstL1InfoTreeAfterBlock`.
- `bridgetracker/sources/ger.go` does `client.FilterLogs(ctx, ethereum.FilterQuery{..., Addresses: []common.Address{s.ContractGlobalExitRootAddress}, ...})` against L1.
- That address comes straight from `bridgetracker.Config.L1GlobalExitRootAddress`, wired in `proxy/cmd/run.go:182` (`sources.NewGERSource(finder, rpcClients, trackerCfg.L1GlobalExitRootAddress, ...)`).
- Unlike `BridgeAddrs` — whose doc comment says an absent entry still matches logs on the event signature alone — `L1GlobalExitRootAddress` has **no fallback**. Unset → Go zero value → `FilterLogs` filters by address `0x0000…0000` → never matches a real `UpdateL1InfoTree`/`UpdateL1InfoTreeV2` log.

## Reproduction (live devnet)

kurtosis 2-L2 devnet, `ghcr.io/agglayer/aggkit:0.11.0-rc4`, proxy `--components=proxy,tracker`, `[Tracker]` configured with the agglayer gRPC URL but without `L1GlobalExitRootAddress`:

1. Send an L1→L2 bridge; it autoclaims normally on-chain (confirmed via `/bridge/v1/claims`).
2. `GET /tracker/v1/network/0/tx/{hash}` still reports `WaitingGERUpdate` / `inProgress` 5+ minutes after the claim landed, and indefinitely thereafter. Not a timing issue — waiting never helps.
3. Setting `L1GlobalExitRootAddress` to the deployment's GlobalExitRoot contract and restarting the proxy immediately unblocks resolution (a re-registered tx goes straight to `finished`).

L2→L1 and L2→L2 routes are unaffected (their routes don't start with `WaitingGERUpdate`... note L2→L2 *does* include `WaitingGERInjection`, which uses a different source and works).

## Suggested fixes (any or all)

1. Set `L1GlobalExitRootAddress` in `proxy/scripts/configuration_based_on_kurtosis.sh` (the value is available from the kurtosis contract-setup outputs, e.g. the same source other services use for the GER address).
2. Fail loudly (or at least log a warning) at tracker startup when `L1GlobalExitRootAddress` is the zero address — today the misconfiguration is completely silent.
3. Longer term: give `GERSource` the same signature-based fallback `BridgeAddrs` has, so the field becomes optional in reality as well as in appearance.

## Environment

- aggkit `v0.11.0-rc4` (git rev `27831b4a`), image `ghcr.io/agglayer/aggkit:0.11.0-rc4`
- kurtosis-cdk 2-L2 enclave; tracker via REST through haproxy

For reference, kurtosis-cdk's recipe was fixed on our side by templating the address from the deployment outputs into `[Tracker]` — but the gap originates in the reference recipe both configs are modeled on.

### LINKED ISSUE #1783: REST: default MaxRequestsPerIPAndSecond to unlimited — rate limiting belongs at the infra layer (field is currently declared but unenforced)
## Summary

Request: make **unlimited** the default for `[REST].MaxRequestsPerIPAndSecond` (and its proxy counterpart), and document that request rate limiting is the infrastructure layer's job (reverse proxy / API gateway / ingress), not the service's.

Two supporting observations from `v0.11.0-rc4` (rev `27831b4a`):

1. **The field is currently dead config.** `MaxRequestsPerIPAndSecond` is declared in `common/config.go:25-27` and defaulted to `10` in `config/default.go` (×3 REST sections) and `proxy/config/default.go:30`, but nothing consumes it: `common/httpserver.go` builds the gin engine with only `gin.Recovery()` + the request logger, no rate-limit middleware is registered anywhere in the tree, and `tollbooth` appears in `go.mod` only as an `// indirect` leftover. So today the config surface *advertises* a 10 req/s per-IP limit that is not actually enforced — operators reading the config will size infrastructure around a limit that doesn't exist (we did).

2. **Per-IP limiting inside the service is the wrong layer anyway.** Any deployment that fronts aggkit with a reverse proxy (haproxy, nginx, an ingress controller — i.e., essentially all of them, including aggkit's own kurtosis recipes) collapses every client onto the proxy's source IP. A per-IP budget then becomes a *global* budget shared by all users: in our kurtosis 2-L2 devnet, one browser tab of the bridge dev-ui (activity polling + one tracker poll per pending transaction, both at 5s) can innocently approach a small per-IP budget, and a second user would halve everyone's headroom. Fixing that inside the service means parsing `X-Forwarded-For`/`X-Real-IP`, which introduces its own spoofing/trust configuration — complexity the fronting proxy already handles natively, with better tooling (sliding windows, per-route policies, ban lists, observability).

## Suggested change

- Default `MaxRequestsPerIPAndSecond = 0` with `0` documented as "unlimited" — or remove the field entirely while it's still unenforced (removing later, once people depend on it, will be a breaking change; note the config loader hard-fails on deprecated fields, so removal is cleaner now than after enforcement ships).
- Add a line to the REST config docs: "aggkit does not rate-limit requests; apply rate limiting at your reverse proxy / gateway if needed", optionally with a haproxy/nginx snippet.
- If in-service enforcement is ever implemented deliberately, keep unlimited as the default and make the limiter proxy-aware (trusted-proxies + forwarded-header parsing) so it doesn't punish everyone behind one LB address.

## Context

Found while integrating the rc4 bridge tracker into the kurtosis-cdk 2-L2 devnet + bridge dev-ui: we initially raised the kurtosis template's `MaxRequestsPerIPAndSecond` from 10 to 50 specifically because haproxy funnels all browser traffic from a single IP — before discovering the limit isn't enforced at all in the current code. Both the phantom limit and the layer mismatch cost investigation time.

arnaubennassar:
## 🔄 Changes Summary
- **#1781 — bridgetracker API.md wire-format drift**: rewrote `docs/bridgetracker/API.md` to match the actual serializer. Most enums are bare strings (`tracking_status`, `bridge_type`, `event.leaf_type`, `step_name`, step `status`); only `ErrorStep.error_type` and `CertificateData.status` use the numeric+`_string` convention. Documented the `event` nesting inside `BridgeStatus`, `block_timestamp`, `step_index`, and omitempty optionals (omitted from the JSON, not `null`). Flagged `GERData` as domain-internal (never on the wire) and `expected_duration` as wired-but-never-populated. Refreshed every example payload to the real shapes.
- **#1782 — silent stall on unset `[Tracker].L1GlobalExitRootAddress`**: added `bridgetracker.Config.Validate()` (zero address → error) called from `runTracker` in `proxy/cmd/run.go`, so the proxy now fails fast at startup with an actionable message instead of silently stalling every L1→L2 route at `WaitingGERUpdate`. `proxy/scripts/configuration_based_on_kurtosis.sh` now emits `[Tracker].L1GlobalExitRootAddress`, read from `polygonZkEVMGlobalExitRootAddress` in the same aggkit config artifact it already uses for `RollupManagerAddr` (the e2e script already set it).
- **#1783 — dead `MaxRequestsPerIPAndSecond` config**: defaulted it to `0` in the sections where it is genuinely dead (`[PublicREST]`, `[AdminREST]` in `config/default.go`, `[REST]` in `proxy/config/default.go` — all `common.RESTConfig`, which has no rate-limit middleware), documented `0` as unlimited there, and documented (field comment + new `RESTConfig` section in `docs/common_config.md`) that aggkit's REST servers do not rate-limit in-process and rate limiting belongs at the reverse proxy / gateway / ingress layer. **`[RPC]` keeps its default of `10`**: it maps to `jRPC.Config` from `github.com/0xPolygon/cdk-rpc`, whose server enforces the field via tollbooth — and `tollbooth.NewLimiter(0, nil)` is *not* unlimited, it is burst-of-1-then-429-forever (an earlier iteration of this PR set it to `0` and broke every e2e job that queries the RPC port more than once; caught by CI and reverted here).
- **Follow-up to #1780 review ([discussion](https://github.com/agglayer/aggkit/pull/1780#discussion_r3748368773)) — drop `idx_bridge_destination_network`**: the index added in migration 0016 hijacks SQLite's query planner (aggkit never runs `ANALYZE`, so an indexed equality is assumed highly selective) and steals the plan from `idx_bridge_deposit_count_desc` / `idx_bridge_from_address_upper`. Reproduced on a 1M-row fixture: network-filtered bridge paging regressed ~0.5ms → ~600ms, network+`from_address` combo ~0.1ms → ~230ms, claim-candidate range queries ~0.5ms → ~180ms. The `UPPER(from_address)` expression index alone fixes the bug 0016 targeted, so the index is deleted in place (migration 0016 has not shipped in any release) and the migration test now asserts it does **not** exist.

## ⚠️ Breaking Changes
- 🛠️ **Config**: running the proxy with the `tracker` component now **fails at startup** if `[Tracker].L1GlobalExitRootAddress` is unset/zero (previously it started and silently never resolved any L1→L2 bridge). Deployments that run the tracker must set the address — the only in-repo config that runs it (`test/e2e/envs/op-pp-2chains`) already does.
- 🔌 **API/CLI**: none (doc-only for the tracker wire format; the serializer is unchanged).
- 🗑️ **Deprecated Features**: none. `MaxRequestsPerIPAndSecond` is kept (removing it would hard-fail existing configs); only its default and documentation changed.

## 📋 Config Updates
- 🧾 **Diff/Config snippet**:
```toml
# was 10 in [PublicREST], [AdminREST] (aggkit) and [REST] (proxy); 0 = unlimited (the limit was never enforced there anyway)
# [RPC].MaxRequestsPerIPAndSecond stays 10 — it IS enforced (tollbooth via cdk-rpc) and 0 would mean "1 request per IP, ever"
MaxRequestsPerIPAndSecond = 0

# now required when running the proxy with the tracker component (generated automatically by
# proxy/scripts/configuration_based_on_kurtosis.sh and configuration_based_on_e2e.sh)
[Tracker]
L1GlobalExitRootAddress = "0x..."
```

## ✅ Testing
- 🤖 **Automatic**: new unit test `TestConfig_Validate` (bridgetracker) covering zero-address → error and valid address → nil. `make build`, `make lint` (0 issues), and `go test ./bridgetracker/... ./proxy/... ./config/... ./common/...` all green. The op-pp-2chains e2e env already sets the GER address, so the new validation does not affect CI.
- 🤖 **Planner regression check**: `TestMigration0016_IndexesExist` red/greened — with the index present the new `must not exist` assertion fails; with the fix it passes. Benchmarks above measured via EXPLAIN QUERY PLAN + timed queries on a migrated 1M-row fixture using the production connection pragmas.
- 🖱️ **Manual**: doc corrections in API.md were cross-checked field-by-field against the JSON struct tags in `bridgetracker/api/` and `bridgetracker/types/` (they also match the live rc4 captures reported in #1781).

## 🐞 Issues
- Closes #1781
- Closes #1782
- Closes #1783

## 🔗 Related PRs
- N/A

## 📝 Notes
- Two of #1781's claims were corrected against the code rather than copied: the leaf-type key under `event` is `leaf_type` (not `bridge_leaf_type`), and `expected_duration` *is* wired in the serializer (`omitempty`) but no resolver populates it — it's documented as reserved/never-emitted rather than deleted.
- #1782's longer-term suggestion (signature-based fallback in `GERSource` so the address becomes truly optional, like `BridgeAddrs`) is intentionally **not** included — fail-fast + fixed reference recipe covers the operational gap; the fallback can be a follow-up if wanted.
- #1783: if in-process rate limiting is ever implemented deliberately, keep `0`/unlimited as the default and make the limiter proxy-aware (trusted proxies + forwarded headers), per the issue.

### DIFF
--- a/bridgesync/migrations/bridgesync0016.sql
+++ b/bridgesync/migrations/bridgesync0016.sql
@@ -1,10 +1,12 @@
 -- +migrate Down
 DROP INDEX IF EXISTS idx_bridge_from_address_upper;
-DROP INDEX IF EXISTS idx_bridge_destination_network;
 
 -- +migrate Up
 -- Backs the "UPPER(from_address) = UPPER($n)" predicate in buildBridgesFilterClause
 -- (bridgesync/processor.go). A plain index on from_address cannot be used through the
 -- UPPER() wrapper, so this is an expression index matching it exactly.
+-- Deliberately NO index on destination_network: aggkit never runs ANALYZE, so SQLite's
+-- default heuristics prefer such an index over idx_bridge_deposit_count_desc (or the
+-- expression index above), replacing early-exit ORDER BY deposit_count plans with full
+-- scans + temp B-tree sorts.
 CREATE INDEX IF NOT EXISTS idx_bridge_from_address_upper ON bridge (UPPER(from_address));
-CREATE INDEX IF NOT EXISTS idx_bridge_destination_network ON bridge (destination_network);
--- a/bridgesync/migrations/migrations_test.go
+++ b/bridgesync/migrations/migrations_test.go
@@ -761,14 +761,19 @@ func TestMigration0016_IndexesExist(t *testing.T) {
 	require.NoError(t, err)
 	defer database.Close()
 
-	for _, indexName := range []string{"idx_bridge_from_address_upper", "idx_bridge_destination_network"} {
-		var name string
-		err := database.QueryRow(
-			`SELECT name FROM sqlite_master WHERE type='index' AND name=?`, indexName,
-		).Scan(&name)
-		require.NoError(t, err, "index %s should exist", indexName)
-		require.Equal(t, indexName, name)
-	}
+	var name string
+	err = database.QueryRow(
+		`SELECT name FROM sqlite_master WHERE type='index' AND name=?`, "idx_bridge_from_address_upper",
+	).Scan(&name)
+	require.NoError(t, err, "index idx_bridge_from_address_upper should exist")
+	require.Equal(t, "idx_bridge_from_address_upper", name)
+
+	// Without ANALYZE stats, an index on destination_network hijacks the query planner away
+	// from idx_bridge_deposit_count_desc and idx_bridge_from_address_upper, so it must not exist.
+	err = database.QueryRow(
+		`SELECT name FROM sqlite_master WHERE type='index' AND name=?`, "idx_bridge_destination_network",
+	).Scan(&name)
+	require.ErrorIs(t, err, sql.ErrNoRows, "index idx_bridge_destination_network must not exist")
 }
 
 func TestMigrationsDown(t *testing.T) {
--- a/bridgetracker/config.go
+++ b/bridgetracker/config.go
@@ -1,6 +1,7 @@
 package bridgetracker
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/agglayer/aggkit/agglayer"
@@ -109,4 +110,13 @@ type Config struct {
 	// adapter (single instance); inject a shared-store implementation so several tracker
 	// instances behind a proxy answer for any registered tx
 	Registry SupervisedRegistry `mapstructure:"-"`
+}
+
+// Validate checks if the configuration is valid
+func (c *Config) Validate() error {
+	if c.L1GlobalExitRootAddress == (common.Address{}) {
+		return fmt.Errorf("[Tracker].L1GlobalExitRootAddress is not set (zero address): " +
+			"the L1 GlobalExitRoot contract address is required for L1->L2 bridge tracking to work")
+	}
+	return nil
 }
--- a/bridgetracker/config_test.go
+++ b/bridgetracker/config_test.go
@@ -1,0 +1,42 @@
+package bridgetracker
+
+import (
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/require"
+)
+
+func TestConfig_Validate(t *testing.T) {
+	tests := []struct {
+		name          string
+		config        Config
+		expectedError string
+	}{
+		{
+			name: "valid config",
+			config: Config{
+				L1GlobalExitRootAddress: common.HexToAddress("0x1f7ad7caA53e35b4f0D138dC5CBF91aC108a2674"),
+			},
+			expectedError: "",
+		},
+		{
+			name:          "zero L1GlobalExitRootAddress",
+			config:        Config{},
+			expectedError: "[Tracker].L1GlobalExitRootAddress",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := tt.config.Validate()
+
+			if tt.expectedError == "" {
+				require.NoError(t, err)
+			} else {
+				require.Error(t, err)
+				require.Contains(t, err.Error(), tt.expectedError)
+			}
+		})
+	}
+}
--- a/common/config.go
+++ b/common/config.go
@@ -23,7 +23,9 @@ type RESTConfig struct {
 	WriteTimeout types.Duration `mapstructure:"WriteTimeout"`
 
 	// MaxRequestsPerIPAndSecond defines how many requests a single IP can
-	// send within a single second
+	// send within a single second. 0 (the default) means unlimited: aggkit
+	// does not enforce request rate limiting in-process. Apply rate limiting
+	// at the fronting reverse proxy / API gateway / ingress if needed.
 	MaxRequestsPerIPAndSecond float64 `mapstructure:"MaxRequestsPerIPAndSecond"`
 }
 
--- a/config/default.go
+++ b/config/default.go
@@ -149,21 +149,21 @@ Host = "0.0.0.0"
 Port = 5576
 ReadTimeout = "2s"
 WriteTimeout = "2s"
-MaxRequestsPerIPAndSecond = 10
+MaxRequestsPerIPAndSecond = 0
 
 [PublicREST]
 Host = "0.0.0.0"
 Port = 5577
 ReadTimeout = "5m"
 WriteTimeout = "5m"
-MaxRequestsPerIPAndSecond = 10
+MaxRequestsPerIPAndSecond = 0
 
 [AdminREST]
 Host = "0.0.0.0"
 Port = 5579
 ReadTimeout = "5m"
 WriteTimeout = "5m"
-MaxRequestsPerIPAndSecond = 10
+MaxRequestsPerIPAndSecond = 0
 
 [BridgeL1Sync]
 DBPath = "{{PathRWData}}/bridgel1sync.sqlite"
--- a/config/default.go
+++ b/config/default.go
@@ -149,7 +149,8 @@ Host = "0.0.0.0"
 Port = 5576
 ReadTimeout = "2s"
 WriteTimeout = "2s"
-MaxRequestsPerIPAndSecond = 0
+# Enforced (tollbooth in cdk-rpc); 0 does NOT mean unlimited here, it means burst-of-1 then 429 forever.
+MaxRequestsPerIPAndSecond = 10
 
 [PublicREST]
 Host = "0.0.0.0"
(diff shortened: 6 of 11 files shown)


arnaubennassar:
@claude

hadjiszs (approved):

PR #1794, opened 2026-08-12
fix(bridgeservice): correct l1-info-tree-index fallback and cross-syncer error codes

### LINKED ISSUE #1795: fix(autoclaim): proof preparer reuses the INC-129 index-space bug (deposit count used as L1 info tree index)
`autoclaim/proof/preparer.go` contains a verbatim copy of the INC-129 bug that was fixed for the
bridge service in #1794.

## The bug

In `getFirstL1InfoTreeIndexForL1Bridge`'s equivalent fallback path (`autoclaim/proof/preparer.go`,
around line 231):

```go
root, err = p.bridgeL1.GetLastRoot(ctx)
...
lastInfo, err = p.l1InfoTree.GetInfoByIndex(ctx, root.Index)
```

`root.Index` is a position in the **L1 bridge exit tree** — i.e. a deposit count (~1,146,035 on
mainnet). `GetInfoByIndex` runs `SELECT * FROM l1info_leaf WHERE position = $1`, which expects an
**L1 info tree index**. These are two different trees with two different counters, both typed as a
bare `uint32`, so nothing catches the mix-up at compile time. The row cannot exist, and the call
fails with `sql: no rows in result set` wrapped as `failed to get last info for L1`.

The failure triggers whenever the L1 bridge syncer trails the L1 info tree syncer — the same skew
that produced 89 alerts on kinexys in INC-129.

## The fix applied in #1794

Clamp on the block number instead, which is a value both trees agree on:

```go
lastInfo, err = b.l1InfoTree.GetLatestL1InfoLeafUntilBlock(ctx, root.BlockNum)
```

`tree.Root` carries `BlockNum`, so this returns the most recent info-tree leaf at or before the
last block the bridge syncer indexed — the clamp the original code intended, in the correct index
space. A `root.BlockNum == 0` guard is also needed (it maps to `ErrNoBlock0`).

## Why it was not fixed in #1794

Deliberately scoped out to keep that PR's diff confined to the bridge service. The TODO comment
added at the call site references this issue.

## Suggested work

- Apply the `GetLatestL1InfoLeafUntilBlock(ctx, root.BlockNum)` clamp plus the `BlockNum == 0` guard.
- Add a regression test with the deposit count and block number kept **orders of magnitude apart**.
  Every pre-existing bridge-service fixture set `Index == BlockNum == depositCount`, and that
  coincidence is exactly why the bug hid for so long — #1794 added
  `require.Greater(t, uint64(lastRootIndex), lastRootBlockNum*100)` to prevent it recurring.
- Audit for any other `GetInfoByIndex(ctx, root.Index)` call sites while there.

Related: #1794, INC-129.

arnaubennassar:
## 🔄 Changes Summary
- **Fixes the INC-129 root cause.** The fallback in `getFirstL1InfoTreeIndexForL1Bridge` took `root.Index` from `bridgeL1.GetLastRoot` — a position in the **L1 bridge exit tree**, i.e. a deposit count (~1,146,035) — and passed it to `l1InfoTree.GetInfoByIndex`, which queries `SELECT * FROM l1info_leaf WHERE position = $1`, an **L1 info tree index**. Two different trees, two different counters, both bare `uint32`. The row could not exist, so `/bridge/v1/l1-info-tree-index` returned HTTP 500 `sql: no rows in result set` whenever the L1 bridge syncer trailed the L1 info tree syncer.
- The fallback now calls `GetLatestL1InfoLeafUntilBlock(ctx, root.BlockNum)`, clamping to the most recent info-tree leaf at or before the last block the bridge syncer indexed — the clamp the original code intended, in the correct index space. `root.BlockNum == 0` is guarded. The tip-anchored binary search, the `GetRootByLER` join, and `ErrNotOnL1Info` are all retained; the search's `upperLimit`/`bestResult` inherit the clamp from the reassigned `lastInfo` without a second assignment.
- **Cross-syncer error taxonomy.** Endpoints that join syncers which can sit at different heights now distinguish "not indexed yet" from a genuine fault, via a single shared `httpStatusForSyncerError` helper: `ErrBlockNotProcessed`, `l1infotreesync.ErrNotFound`, `ErrNoBlock0`, `db.ErrNotFound`, `ErrNotOnL1Info` → **404**; `sync.ErrInconsistentState` → **503**; anything else → **500**.
- **`ClaimProofHandler` metrics fix.** `statusCode` was initialised to `200` and never reassigned, so all 13 exit paths reported `200` to `reportMetrics` — every failure was invisible in metrics. All exits now set it correctly.
- **`GetInfoByIndex` not-found was unwrapped.** `getInfoByIndexWithTx` skipped `db.ReturnErrNotFound`, returning a raw `sql: no rows in result set`. That was the literal 500 body in INC-129, and it meant `errors.Is(err, db.ErrNotFound)` was dead code at three call sites. One-line wrap added.
- **Log levels.** Not-ready conditions now log at `Debug` (404) / `Warn` (503) instead of `Error`, so the INC-129 alert volume does not simply migrate from HTTP 500s into ERROR logs.
- **Same bug fixed in `autoclaim`** (closes #1795). `autoclaim/proof/preparer.go` carried a verbatim copy of the index-space confusion in its own `GetRootByLER` fallback. It now clamps on `root.BlockNum` too. After this, `grep -rn "GetInfoByIndex(ctx, .*\.Index)"` over non-test Go returns **zero** hits — that was the last one.

## ⚠️ Breaking Changes
- 🔌 **API/CLI**: **HTTP 500 → 404 / 503 on three endpoints.** `/bridge/v1/l1-info-tree-index`, `/bridge/v1/claim-proof`, and the **L1** path of `/bridge/v1/injected-l1-info-leaf` now return `404` when a syncer has not indexed the requested data yet (callers should treat it as *not ready yet, retry later*) and `503` when a syncer is halted / in an inconsistent state. Previously all of these were `500`.
  - `/injected-l1-info-leaf`'s L2 path already returned `404`; only its L1 path changes.
  - `bridgeservice/client`: `GetL1InfoTreeIndex` and `GetClaimProof` switch from `doRequest` to `doRequestAllowNotFound`, so they now return the `ErrNotFound` sentinel on `404` instead of a generic error. **Signatures and return types are unchanged**, but callers that treated any error as fatal may now want to retry on `ErrNotFound`. A `503` deliberately stays a plain non-`ErrNotFound` error so a halted syncer is not mistaken for "not indexed yet".
- No success-path response body or schema changed anywhere.

## 📋 Config Updates
- None.

## ✅ Testing
- 🤖 **Automatic**:
  - `make lint` → `0 issues.`
  - `make test-unit` → exit 0 (whole repo).
  - New regression test `TestGetFirstL1InfoTreeIndexForL1Bridge_KinexysIndexSpaceSkew` reproduces the exact skew: `GetRootByLER` cannot resolve the tip leaf's MER, and `GetLastRoot` returns `Index = 1_146_035` (a deposit count) with `BlockNum = 8_412`. The fixtures keep the two **~136× apart on purpose** — and assert that skew — because every pre-existing fixture set `Index == BlockNum == depositCount`, which is precisely the coincidence that let an index-space confusion hide for so long.
  - **Red/green proof**: reverting only the one fallback line back to `GetInfoByIndex(ctx, root.Index)` makes both subtests fail on the unexpected mock call for `GetLatestL1InfoLeafUntilBlock`; restoring it byte-for-byte returns them to green.
  - The four existing fallback subtests were **rewritten, not deleted**. One of them previously *pinned the bug* by asserting the error contained `"failed to get last info for L1"`; it now asserts `ErrorIs(l1infotreesync.ErrNotFound)`, a 404 mapping, and `NotContains` that old message. A new subtest covers the `root.BlockNum == 0` guard.
  - `TestHTTPStatusForSyncerError` (14 rows — every sentinel bare *and* wrapped) and `TestRespondSyncerError` cover the mapping directly, including `db.ErrNotFound` and `l1infotreesync.ErrNotFound` as separate 404 cases, since the new fallback returns the latter while its siblings return the former.
  - Per-endpoint status tables (10/12/24 rows). `ClaimProofHandler`'s exits are asserted **through the Prometheus counter**, and every failure row also asserts the `200` counter did *not* move — pinning the metrics bug shut.
  - Client tests strengthened: `TestGetL1InfoTreeIndex`'s weak `require.Error` became `require.ErrorIs(err, ErrNotFound)`; `TestGetClaimProof` gained its first `404` subtest; both gained a `503` subtest.
  - **autoclaim (#1795)**: `TestFirstL1InfoTreeIndexForL1BridgeIndexSpaceSkew` reproduces the same skew against the proof preparer (`GetLastRoot().Index = 1_146_035` vs `BlockNum = 8_412`, skew asserted), plus tests for the `BlockNum == 0` guard and the clamp-lookup error wrap. Its probed block (`6_200`, the midpoint of `[4_000, 8_400]`) doubles as proof the clamp actually tightened `upperLimit` — unclamped, the search would probe `6_250` and miss the fixture. Proven red by reverting the one line: the test fails with `l1infotreesync has no L1 info tree leaf at or before L1 block 8412`.
- 🖱️ **Manual**: not reproduced against a live environment. The kinexys condition is reproduced deterministically in the regression test above rather than by hand.

## 🐞 Issues
- Fixes **INC-129** (kinexys `/l1-info-tree-index` HTTP 500, 89 alerts).
- Closes #1795 (the same index-space bug duplicated in `autoclaim/proof/preparer.go`).

## 🔗 Related PRs
- None.

## 📝 Notes

**Scope widening — `l1infotreesync/processor.go`.** The diff intentionally reaches one line outside `bridgeservice/`: `getInfoByIndexWithTx` now wraps its return in `db.ReturnErrNotFound(...)`. Without it, the handlers' `errors.Is(err, db.ErrNotFound)` checks are dead code and the endpoint keeps returning a raw `sql: no rows in result set`. All 9 non-test callers were audited; the decisive evidence that this is the right layer is that three `autoclaim` call sites already work around the asymmetry by checking `errors.Is(db.ErrNotFound) || errors.Is(sql.ErrNoRows)`. `translateError` was deliberately **not** added at the syncer wrapper, so the sentinel stays `db.ErrNotFound`, matching every sibling method.

**Incidental swagger drift (pre-existing, not new).** Regenerating the swagger artefacts also picks up `/l1-info-tree-leaf-by-ger` and `/root-by-ler` entries plus the `types.RootByLERResponse` schema. These were **already annotated in `bridge.go` on `develop`** but had never been regenerated into the committed docs. They are not new functionality introduced here — verified with `git show origin/develop:bridgeservice/bridge.go`.

**The `autoclaim` copy of the bug is now fixed here too (#1795).** `autoclaim/proof/preparer.go` had a verbatim copy — a bridge-exit-tree `root.Index` from `GetLastRoot` fed into `GetInfoByIndex`, with the same `"failed to get last info for L1"` wrapper. It was first flagged with a `TODO`, then fixed in this PR. Two details worth a reviewer's eye:

- The `root.BlockNum == 0` guard reuses the existing `bridgeservice.ErrNotOnL1Info` rather than introducing a sentinel. Its only caller, `selectL1InfoTreeIndex`, already treats that sentinel as *retry next cycle* (`Result{Ready: false}`), which is the correct reading of "bridgesync L1 has indexed nothing yet" — a transient startup state, not a fault. Letting `l1infotreesync.ErrNoBlock0` escape instead would leak an l1infotreesync-internal sentinel into autoclaim's error contract.
- **The `GetLastRoot` fallback in `autoclaim/proof` had zero test coverage** — no pre-existing test ever set `lastRoot` or `rootErrs` on the fake. The copied bug came with a copied coverage gap, which is why it survived. `preparer.go:111` retains a *legitimate* `GetInfoByIndex` call whose argument is already in info-tree space; that one is correct and should not be "fixed".

**Deliberate non-changes.** The deposit-anchored rewrite (`GetBridgeByDepositCount` + a `(block_num, block_pos) >` query) is the better long-term design and is *not* attempted here — this fix is deliberately minimal and in place. `getFirstL1InfoTreeIndexForL2Bridge` is untouched beyond error mapping, and the syncers are not merged. `L1InfoTreeLeafByGERHandler` already matched the taxonomy and was left alone. One pre-existing inconsistency remains: `L1InfoTreeIndexForBridgeHandler`'s nil-syncer path still answers `500` while the same condition answers `503` in `ClaimProofHandler`; fixing it cleanly needs a new sentinel, so it is left as-is.

https://claude.ai/code/session_01WvzbGpjp6VeiW6ccThguGC

---

## 🔍 Review follow-up

Applied from the automated review:

- **Pessimistic `statusCode` default** (suggestion 2). `L1InfoTreeIndexForBridgeHandler`, `InjectedL1InfoLeafHandler`, and `ClaimProofHandler` now initialise `statusCode` to `500` and set `200` explicitly on the success path. Worth noting the suggestion as literally written would have introduced a bug: each of these handlers' success responses used `c.JSON(statusCode, ...)` relying on the `http.StatusOK` initialiser, so flipping the default alone would have returned HTTP 500 on success. Setting 200 explicitly is what makes the pessimistic default safe.
- **Autoclaim duplicate bug** (suggestion 3). Filed as #1795, and then **fixed in this PR** rather than deferred — see the Notes section. The `TODO` comment is gone; the audit the issue asked for is clean.

- **DB-level coverage of the clamp** (integration-coverage suggestion). Added `TestGetLatestL1InfoLeafUntilBlockClamp` in `l1infotreesync/processor_test.go`. The pre-existing table test for that method only asserted the three error sentinels (`ErrNoBlock0`, `ErrBlockNotProcessed`, `ErrNotFound`), so the actual **"newest leaf at or before `blockNum`"** semantics the fallback now depends on were never exercised against real rows. This was the one genuinely untested link, and it is the link a mocked syncer *cannot* cover: a mock returns whatever it is told, so it can never catch an index-space error — it only proves the handler passes `root.BlockNum` to the clamp, not that the query answers correctly.

  The new test inserts leaves at L1 blocks `1_000 / 5_000 / 9_000` (indexes `0 / 1 / 2` — kept orders of magnitude apart on purpose, same discipline as the handler fixtures) and asserts index, block, and `leaf.BlockNumber <= requested` across eight cases: exact-boundary, above-all, between-leaves, ±1 around a leaf, and below-all → `ErrNotFound`. Verified non-vacuous by mutation: flipping `ORDER BY … DESC` to `ASC` and `block_num <= $1` to `>= $1` each make it fail (source restored, md5 verified).

Declined, with reasons:

- **Nil-syncer taxonomy unification** (suggestion 1) — the review itself notes this may be intentional, and it is: a nil syncer is a *configuration* fault, not a syncer *state*, and routing it through `respondSyncerError` would require a new sentinel. Left as documented under Deliberate non-changes.
- **`GetLatestL1InfoLeafUntilBlock` pointer parameter** (suggestion 4) — the premise is incorrect. The `*uint64` is the pre-existing **processor** signature, and its `nil` path *is* exercised: `l1infotreesync/l1infotreesync.go:301` passes `nil` for the unbounded "latest leaf" variant, so making it non-pointer would break that caller. The public wrapper this PR consumes already takes a plain `uint64`, which is exactly what the `L1InfoTreeSyncer` interface addition mirrors.
- **Test constant renaming** (suggestion 5) — marked optional by the review, and the proposed names (e.g. `fallbackL1Block_DepositCountBelowBlock`) use underscores in mixedCaps identifiers, which Go style and this repo's linters discourage.
- **A full HTTP → handler → syncer → **database** wired test** (the rest of the integration-coverage suggestion). Two of its three bullets were already satisfied before the suggestion was made: the handler status tests go through a **real gin router** via `performRequest` and assert `w.Code` *and* `w.Body`, and `assertHandlerStatus` reads the **actual Prometheus counter** before and after each request, asserting both that the expected status was reported to `reportMetrics` and that the `200` counter did not move. So "verify the full HTTP response (status + body + metrics)" and "test the `reportMetrics` calls with the correct status codes" are covered. The mocks are also already at the syncer-interface seam — there is no higher seam short of a real database.

  What remained was standing up real `l1infotreesync` + `bridgesync` instances with populated sqlite behind a live `BridgeService`. That is a substantial bespoke harness, and it duplicates what the repo's kurtosis/docker e2e suites already do end-to-end (they are excluded from `make test-unit` by `-short`). The targeted DB-level test above closes the actual coverage hole at a fraction of the cost and maintenance surface. Happy to add the wired harness if reviewers want it, but it did not look like the right trade here.

### DIFF
--- a/autoclaim/proof/preparer.go
+++ b/autoclaim/proof/preparer.go
@@ -28,6 +28,7 @@ type L1BridgeSyncer interface {
 // L1InfoTreeSyncer exposes the L1 info tree methods needed to prepare L1-origin claim proofs.
 type L1InfoTreeSyncer interface {
 	GetInfoByIndex(ctx context.Context, index uint32) (*l1infotreesync.L1InfoTreeLeaf, error)
+	GetLatestL1InfoLeafUntilBlock(ctx context.Context, blockNum uint64) (*l1infotreesync.L1InfoTreeLeaf, error)
 	GetRollupExitTreeMerkleProof(ctx context.Context, networkID uint32, root common.Hash) (treetypes.Proof, error)
 	GetLastInfo() (*l1infotreesync.L1InfoTreeLeaf, error)
 	GetFirstInfo() (*l1infotreesync.L1InfoTreeLeaf, error)
@@ -228,14 +229,17 @@ func (p *Preparer) firstL1InfoTreeIndexForL1Bridge(
 		if root == nil {
 			return 0, fmt.Errorf("failed to get last root for L1: empty result")
 		}
-		// TODO(#1795): root.Index is a position in the L1 bridge exit tree (a deposit count),
-		// but GetInfoByIndex expects an L1 info tree index. The two counters are unrelated, so
-		// this lookup fails whenever the L1 bridge syncer trails the L1 info tree syncer. The
-		// same bug was fixed in bridgeservice/bridge.go by clamping on root.BlockNum via
-		// GetLatestL1InfoLeafUntilBlock; this call site still needs the equivalent fix.
-		lastInfo, err = p.l1InfoTree.GetInfoByIndex(ctx, root.Index)
+		if root.BlockNum == 0 {
+			return 0, fmt.Errorf("bridgesync L1 has not indexed any block yet: %w", bridgeservice.ErrNotOnL1Info)
+		}
+		// root.Index is a position in the L1 bridge exit tree (i.e. a deposit count), not an L1 info
+		// tree index, so it must never be fed into an l1infotreesync leaf lookup. Clamp by the last
+		// L1 block indexed by bridgesync L1 instead, which is the same index space for both syncers.
+		lastInfo, err = p.l1InfoTree.GetLatestL1InfoLeafUntilBlock(ctx, root.BlockNum)
 		if err != nil {
-			return 0, fmt.Errorf("failed to get last info for L1: %w", err)
+			return 0, fmt.Errorf(
+				"l1infotreesync has no L1 info tree leaf at or before L1 block %d "+
+					"(last block indexed by bridgesync L1): %w", root.BlockNum, err)
 		}
 		if lastInfo == nil {
 			return 0, fmt.Errorf("failed to get last info for L1: empty result")
--- a/autoclaim/proof/preparer.go
+++ b/autoclaim/proof/preparer.go
@@ -228,6 +228,11 @@ func (p *Preparer) firstL1InfoTreeIndexForL1Bridge(
 		if root == nil {
 			return 0, fmt.Errorf("failed to get last root for L1: empty result")
 		}
+		// TODO(#1795): root.Index is a position in the L1 bridge exit tree (a deposit count),
+		// but GetInfoByIndex expects an L1 info tree index. The two counters are unrelated, so
+		// this lookup fails whenever the L1 bridge syncer trails the L1 info tree syncer. The
+		// same bug was fixed in bridgeservice/bridge.go by clamping on root.BlockNum via
+		// GetLatestL1InfoLeafUntilBlock; this call site still needs the equivalent fix.
 		lastInfo, err = p.l1InfoTree.GetInfoByIndex(ctx, root.Index)
 		if err != nil {
 			return 0, fmt.Errorf("failed to get last info for L1: %w", err)
--- a/autoclaim/proof/preparer_test.go
+++ b/autoclaim/proof/preparer_test.go
@@ -564,14 +564,17 @@ func (f *fakeL1BridgeSyncer) GetLastRoot(_ context.Context) (*treetypes.Root, er
 }
 
 type fakeL1InfoTreeSyncer struct {
-	infoByIndex      map[uint32]*l1infotreesync.L1InfoTreeLeaf
-	infoByIndexErr   error
-	rollupProof      treetypes.Proof
-	rollupProofErr   error
-	lastInfo         *l1infotreesync.L1InfoTreeLeaf
-	firstInfo        *l1infotreesync.L1InfoTreeLeaf
-	infoAfterBlock   map[uint64]*l1infotreesync.L1InfoTreeLeaf
-	rollupProofCalls []rollupProofCall
+	infoByIndex          map[uint32]*l1infotreesync.L1InfoTreeLeaf
+	infoByIndexErr       error
+	latestInfoUntilBlock *l1infotreesync.L1InfoTreeLeaf
+	latestInfoUntilErr   error
+	latestInfoUntilCalls []uint64
+	rollupProof          treetypes.Proof
+	rollupProofErr       error
+	lastInfo             *l1infotreesync.L1InfoTreeLeaf
+	firstInfo            *l1infotreesync.L1InfoTreeLeaf
+	infoAfterBlock       map[uint64]*l1infotreesync.L1InfoTreeLeaf
+	rollupProofCalls     []rollupProofCall
 }
 
 func (f *fakeL1InfoTreeSyncer) GetInfoByIndex(
@@ -586,6 +589,17 @@ func (f *fakeL1InfoTreeSyncer) GetInfoByIndex(
 		return nil, errors.New("info not found")
 	}
 	return info, nil
+}
+
+func (f *fakeL1InfoTreeSyncer) GetLatestL1InfoLeafUntilBlock(
+	_ context.Context,
+	blockNum uint64,
+) (*l1infotreesync.L1InfoTreeLeaf, error) {
+	f.latestInfoUntilCalls = append(f.latestInfoUntilCalls, blockNum)
+	if f.latestInfoUntilErr != nil {
+		return nil, f.latestInfoUntilErr
+	}
+	return f.latestInfoUntilBlock, nil
 }
 
 func (f *fakeL1InfoTreeSyncer) GetRollupExitTreeMerkleProof(
@@ -615,6 +629,12 @@ func (f *fakeL1InfoTreeSyncer) GetFirstInfoAfterBlock(blockNum uint64) (*l1infot
 	}
 	return info, nil
 }
+
+// Compile-time assertion: fakeL1InfoTreeSyncer implements L1InfoTreeSyncer.
+var _ L1InfoTreeSyncer = (*fakeL1InfoTreeSyncer)(nil)
+
+// Compile-time assertion: fakeL1BridgeSyncer implements L1BridgeSyncer.
+var _ L1BridgeSyncer = (*fakeL1BridgeSyncer)(nil)
 
 type getProofCall struct {
 	depositCount  uint32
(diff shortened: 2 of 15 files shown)


arnaubennassar:
@claude

arnaubennassar:
@claude

taylanpince (approved):

PR #1835, opened 2026-09-08
fix(bridgeservice): encode BridgeResponse.GlobalIndex as BigIntString, not bare JSON number

### LINKED ISSUE #1820: BridgeResponse.GlobalIndex is emitted as a JSON number, bypassing BigIntString; swagger declares string
`BridgeResponse.GlobalIndex` (`bridgeservice/types/types.go:110`, tag `v0.11.0-rc7`) is declared as `*big.Int`, not the repo's own `BigIntString` wrapper, so it is emitted on the wire as a bare JSON number. The struct tag and the generated swagger both say `string`; the actual runtime encoding does not.

```go
// bridgeservice/types/types.go:95
type BridgeResponse struct {
    ...
    // bridgeservice/types/types.go:110
    GlobalIndex *big.Int `json:"global_index" example:"4294967296" swaggertype:"string"`
    ...
    // bridgeservice/types/types.go:131 -- sibling field, correctly typed
    Amount BigIntString `json:"amount" example:"1000000000000000000"`
}
```

`BigIntString` is defined right above, at `types.go:20-22`:

```go
// BigIntString is a wrapper type for big.Int for Swagger compatibility
type BigIntString string
```

Every other `GlobalIndex` field in this same file uses it correctly: `ClaimResponse.GlobalIndex` (`types.go:172`), `UnsetClaimResponse.GlobalIndex` (`types.go:440`), `SetClaimResponse.GlobalIndex` (`types.go:462`) are all `BigIntString`. `BridgeResponse.GlobalIndex` is the one holdout still typed as the raw `*big.Int`.

The `swaggertype:"string"` tag only steers the swag-generated doc (`bridgeservice/docs/swagger.yaml:89-93`, under `types.BridgeResponse`, `type: string`) -- it has no effect on `encoding/json` at runtime. `*big.Int` implements its own `MarshalJSON` (`math/big`'s `intmarsh.go`) that writes the unquoted decimal digits directly, so the doc and the wire disagree: swagger says string, the actual response is a number.

This affects every response from `GET /bridges` (`bridgeservice/bridge.go:158`), since the handler returns `types.BridgesResult` -> `[]*BridgeResponse`.

**Consumer impact**: any JSON parser that maps JSON numbers to IEEE-754 doubles (i.e. every standard `JSON.parse` in JS, and equivalents in most other languages) loses precision above `2^53`. `global_index` for an L1-origin deposit is commonly >= `2^64`-scale (mainnet-flag bit set, e.g. `18446744073709551621`), so a plain `JSON.parse` silently rounds it to a different value -- which can resolve to the wrong bridge/claim entirely. `BigIntString`, being a Go `string`, is JSON-quoted and survives any parser untouched, which is exactly why it exists ("wrapper type for big.Int for Swagger compatibility").

Related: `#1816` is a proof-of-concept ("contract-first pipeline for the bridge service") built around this exact defect class -- same underlying report, not merged.

**Suggested fix**: change `BridgeResponse.GlobalIndex` from `*big.Int` to `BigIntString`, matching every other `GlobalIndex` field in the package.

arnaubennassar:
## 🔄 Changes Summary
`BridgeResponse.GlobalIndex` (`bridgeservice/types/types.go`) was declared as `*big.Int`. `math/big`'s own `MarshalJSON` writes unquoted decimal digits, so `global_index` went out on the wire as a **bare JSON number**, even though the struct tag said `swaggertype:"string"` and the generated swagger declared `type: string`. Every sibling `GlobalIndex` field in the same file (`ClaimResponse.GlobalIndex`, `UnsetClaimResponse.GlobalIndex`, `SetClaimResponse.GlobalIndex`) already used the repo's own `BigIntString` wrapper — `BridgeResponse` was the lone holdout.

For an L1-origin deposit, `global_index = 2^64 + deposit_count` (`bridgesync/processor.go`), which exceeds JavaScript's `Number.MAX_SAFE_INTEGER`. A plain `JSON.parse` on the bare-number form silently rounds it to a **different** value, which can resolve to a different bridge/claim entirely.

This PR switches `BridgeResponse.GlobalIndex` to `BigIntString`, matching the other three fields, and updates the two call sites:
- `NewBridgeResponse` (`bridgeservice/utils.go`) now writes `globalIndex.String()` into the wrapper.
- `bridgetracker/sources/activity.go`'s `ActivitySource.ClaimInfo` recovers a `*big.Int` via `.ToBigInt()` where the downstream client API still expects one.
- `bridgetracker/activity.go`'s cache uses the string form directly as its map key (previously `.String()` on the `*big.Int`, now a plain string cast) — this is an equality-lookup key, never sorted, so no ordering behavior changes.
- `test/e2e/bridge_utils.go` and `test/e2e/proxy_tracker_test.go` call `.ToBigInt()` where `GlobalIndex` is passed into contract-binding calls that expect `*big.Int`.

## ⚠️ Breaking Changes
- 🛠️ **Config**: N/A
- 🔌 **API/CLI**: **Yes.** `GET /bridge/v1/bridges` and `GET /tracker/v1/activity/from/{from_address}` now emit `global_index` as a JSON **string** instead of a bare JSON **number**. Any consumer that parses `global_index` as a numeric type (e.g. `JSON.parse` mapping it straight to a JS `number`) must switch to reading/parsing it as a string. Note that the generated swagger schema for these fields is **unchanged** — it already declared `global_index` as `type: string` (via the `swaggertype:"string"` override on the old `*big.Int` field); only the runtime wire encoding changes to finally match the documented contract.

## 📋 Config Updates
- N/A

## ✅ Testing
- 🤖 **Automatic**:
  - New: `TestNewBridgeResponse_GlobalIndexMarshalsAsQuotedString` (asserts the quoted form is present and the unquoted form absent), `TestBigIntString_ToBigIntRoundTrip`, `TestActivityItemMarshalJSON_EmbeddedBridgeGlobalIndexIsQuotedString` (covers the tracker's embedded copy of the bridge in activity items).
  - `make lint` → `0 issues.`, exit 0.
  - Full non-e2e unit suite: 94 packages `ok`, 0 FAIL, including `bridgeservice`, `bridgeservice/types`, `bridgetracker`, `bridgetracker/api`, `bridgetracker/sources`.
  - `make generate-swagger-docs` produced a byte-identical result — zero generated-file diff, since the swagger already said `type: string`.
- 🖱️ **Manual**: Live before/after capture from a 2-L2 anvil devnet (aggkit-proxy serving `/bridge/v1` and `/tracker/v1`), same deposit position (L1-origin, `deposit_count 2`, `global_index 2^64+2 = 18446744073709551618`), `claim_status: claimed`. The two captures come from two runs of a deterministic environment at the same deposit position, not one continuous chain.

  Before (unmodified `v0.11.0-rc9`), `GET /tracker/v1/activity/from/0x8943...`:
  ```
  "global_index":18446744073709551618
  ```
  Unquoted — a plain `JSON.parse` on this yields `18446744073709552000` (`typeof number`), a different number than the one on the wire.

  Interestingly, that same pre-fix payload also contained the sibling `ClaimResponse` field, already on `BigIntString`, right next to it:
  ```
  "global_index":"18446744073709551618"
  ```
  quoted — one response demonstrating both the bug and the sibling field that proves the fix is right.

  After (this branch), same endpoint:
  ```
  "global_index":"18446744073709551618"
  ```
  Quoted, zero unquoted occurrences anywhere in the payload. `JSON.parse` now yields the exact string `"18446744073709551618"` (`typeof string`).

  Same result on `GET /bridge/v1/bridges?network_id=0`: all three L1-origin rows go from bare numbers to quoted strings.

## 🐞 Issues
- Closes #1820

## 🔗 Related PRs
- agglayer/sdk#28 — the consumer that surfaced this defect.
- agglayer/aggkit#1816 — a related contract-first POC covering the same defect class (reference only; this PR does not supersede it).

## 📝 Notes
- Chose `BigIntString` over a custom `MarshalJSON` on `*big.Int` because it matches all three sibling `GlobalIndex` fields already in this file, and keeps the Go *type* itself honest about the wire encoding, rather than fixing the bytes on the wire while leaving the type inconsistent with its siblings and hiding the encoding behind a method.
- The generated swagger schema does not change: `global_index` was already documented as `type: string` (via `swaggertype:"string"` on the old `*big.Int`), and `BigIntString` is natively a Go string, so the docs and the runtime now finally agree — no schema diff, but a genuine wire-format change for consumers parsing the bare number.
- `agglayer/sdk`'s client-side `quoteGlobalIndex` workaround (added to cope with the bare-number encoding) should stay in place until its minimum supported aggkit version contains this fix — it was verified idempotent over the new quoted form, so it keeps working unchanged after this lands.

https://claude.ai/code/session_01Gzd3Uy5KMYZGm4q7y4hpvc


joanestebanr (commented):
bridge service swagger must be updated?
This is a breaking change, maybe the bridgetracker must support both versions

joanestebanr:
@claude review

arnaubennassar:
Thanks for the review, @joanestebanr — answering both points with evidence.

## Q1 — "bridge service swagger must be updated?"

No, and this is actually the sharpest way to describe the underlying bug: the schema **already** declared `global_index` as `type: string`, via a `swaggertype:"string"` override that was sitting on the old `*big.Int` field:

```go
// before this PR
GlobalIndex *big.Int `json:"global_index" example:"4294967296" swaggertype:"string"`
```

Regenerating docs against this branch (`make generate-swagger-docs`) produces a **byte-identical** result — `git diff origin/develop...HEAD -- bridgeservice/docs bridgetracker/api/docs` is empty. The published spec still reads:

```json
"global_index": {
    "description": "Global index of the bridge event (consisted of mainnet flag, rollup id and deposit count)",
    "type": "string",
    "example": "4294967296"
}
```

So there is no schema change at all — the contract has always promised `string`; only the Go implementation lied about it (encoding a `*big.Int` as a bare JSON number despite the override). This PR just makes the wire format honor the contract the docs already published.

## Q2 — "maybe the bridgetracker must support both versions"

Two parts:

**(a) Dual-format isn't possible server-side.** `encoding/json` emits exactly one form per field per response — a server can't return both a quoted string and a bare number for the same `global_index` in the same payload. The only way to "support both" would be API versioning (freezing a `v1` that keeps the precision-loss bug forever), which is far heavier than the client-side handling that already exists.

**(b) It isn't needed — the consumer set is known, and both known consumers are fine:**

- **`agglayer/sdk`** already has a client-side `quoteGlobalIndex` pre-parse (added specifically to cope with the bare-number bug), and it's verified idempotent over the new quoted form (`quoteGlobalIndex(x) === x` against a live post-fix capture). So the SDK works unchanged before and after this lands — it degrades to a no-op rather than needing a lockstep release. (SDK-side: `agglayer/sdk#37`, merged into `feat/aggkit-bridge-client`, the head of `agglayer/sdk#28`.)
- **`agglayer/agglayer-bridge-hub-api`** is the only other known consumer, and it's verified unaffected — checked against `main` @ `e8adf52` (2026-08-27; a newer deploy or unmerged branch could differ):
  - Its `IBridgeTx` interface (`packages/consumer/src/interfaces/bridge_tx.ts`) doesn't declare `global_index` at all. `mapBridgeTransactions` reads `deposit_count`/`block_num`/`block_pos`/`tx_hash` and **recomputes** the index locally via `computeGlobalIndex(indexLocal, sourceNetworkId)` (`packages/auto-claim/src/services/transaction.ts`) — it never trusts the wire field on `/bridges`.
  - Its only `global_index` *reads* come from `/claims` — `IClaimTx.global_index: string` → `BigInt(...)` in `packages/consumer/src/mappers/transaction.ts` (via `mapClaimTransactions`) — and `ClaimResponse.GlobalIndex` was already `BigIntString`, untouched by this PR.
  - Its `/bridges` consumption takes exactly one field off the bridge object: `bridge_tx_metadata: txData.bridges[0].metadata` (`packages/api/src/services/proof.ts`) — no raw pass-through.
  - `zod`/`@hono/zod-openapi` there validates bridge-hub-api's *own* routes, not aggkit responses (those go through plain `response.json()`), so a strict-schema rejection on ingest isn't in play either.
  - Forward-looking note: bridge-hub-api is safe today precisely because it recomputes rather than trusting `/bridges`. If it ever starts reading `global_index` off `/bridges` directly, that read would be unsafe against any aggkit *older* than this PR (bare number → `JSON.parse` corrupts every L1-origin value) and safe after it — its existing `/claims` handling (string → `BigInt`) is the pattern to copy.
  - `/tracker/v1/activity`, the other endpoint this PR touches, isn't consumed by bridge-hub-api at all.

**So: shipping the breaking change is the right call.** It's breaking only for a hypothetical consumer parsing `global_index` as a number on L1-origin values (≥ 2^64) — and such a consumer was already receiving a **silently wrong value**, not a working one:

```
pre-fix:  "global_index":18446744073709551618   (bare)     -> JSON.parse -> 18446744073709552000 (typeof number, wrong value)
post-fix: "global_index":"18446744073709551618" (quoted)   -> JSON.parse -> "18446744073709551618" (exact string)
```
(Same deposit position — L1-origin, `deposit_count 2`, `global_index = 2^64 + 2`, `claim_status: claimed`. Honest caveat: these are two runs of a deterministic devnet at the same deposit position, not one continuous chain observed across a live upgrade.)

Corroboration worth a line: `agglayer-dev-ui` independently hit the same bug and built its own `quotePrecisionUnsafeIntegers` pre-parse workaround, with comments recording the identical bridge-bare / claim-quoted asymmetry. Three codebases, same conclusion, same fix direction.

## On the automated `@claude review`

Addressing its two non-blocking observations directly:

1. **Captures not attached** — quoting the raw before/after lines inline above (see the `pre-fix`/`post-fix` block) rather than just describing them.
2. **`ToBigInt()` returns a zero `*big.Int` on `SetString` failure** — this is pre-existing, not introduced here: it already applied to the three sibling `BigIntString` fields (`ClaimResponse`, `UnsetClaimResponse`, `SetClaimResponse`) before this PR. It grants no new capability — a hostile bridge service could already return a different decimal — so it's a follow-up, not a regression. One concrete shape worth flagging for that follow-up: `SetString(s, 0)`'s base-0 auto-detection newly admits `"0755"` → 493 / `"0x10"` → 16 for this field, which only matters if `BigIntString` ever becomes a request-input type.

On its release-note suggestion: this repo has no `CHANGELOG.md`, no `.changeset/`, and no `.github/release.yml` auto-notes config — GitHub Releases here are hand-written per-tag (see `v0.11.0-rc9`) with a "⚠️ Breaking Changes" section grouped by component. The next one covering this fix should say, concretely:
- **Affected endpoints**: `GET /bridge/v1/bridges`, `GET /tracker/v1/activity/from/{from_address}`
- **Required action**: parse `global_index` as a string, not a number (matches the swagger contract, which never changed)
- **Observable impact**: only L1-origin values ≥ 2^64 differ between before/after — everything else was already round-tripping correctly

Happy to discuss further if there's a concrete consumer I'm missing.

joanestebanr:
> Q2 — "maybe the bridgetracker must support both versions"

I mean that the aggkit-proxy client should be compatible with both versions to simplify deployment and avoid forcing all bridge services to be upgraded simultaneously.

arnaubennassar:
@joanestebanr — correcting myself on Q2. **"Isn't needed" was wrong.** I conflated the two JSON directions: the *marshal* side genuinely can't emit both forms (one field, one encoding per response), but the *unmarshal* side both can and must accept both — and aggkit's own bridgetracker/proxy is exactly the consumer that needs it, since it fans out to remote bridge services deployed and upgraded independently. Your second clause was right.

**Concrete failure this would have caused:** a post-fix proxy/tracker talking to a remote bridge service still on an older aggkit fails to decode:
```
json: cannot unmarshal number into Go struct field ... of type BigIntString
```
Path: `bridgetracker/sources/activity.go:74` gets one client per network via `aggkitBridgeClientFor(networkID)`, fans out over `:73 s.finder.NetworkIDs()`, `:114 svc.GetBridges(...)` → `bridgeservice/client/client.go:144` (`var resp types.BridgesResult`) → `json.Unmarshal` at `:545`. The entire activity scan for that network breaks, not just one field.

**Fix landed in `0d02a0eb`** (PR now 4 commits). `BigIntString.UnmarshalJSON` accepts a quoted string, a bare number, or `null`. Marshaling is untouched — still emits the quoted string. Strict on write, tolerant on read.

The numeric form decodes via `json.Number`, which is the non-obvious part: the obvious implementations silently reintroduce the bug. On `18446744073709551618` (2^64 + 2): `float64` → `18446744073709551616` (corrupted), `int64` → overflow error, `json.Number` → exact (preserves the literal text).

Tests: the 4-way `{quoted, bare} × {small, ≥2^64}` matrix pinning the exact 20-digit value (so a float64/int64 reimplementation fails the build), plus `null` and invalid-JSON, plus `TestBridgesResult_UnmarshalJSON_PreFixBareNumberGlobalIndex` decoding a realistic pre-fix `/bridges` payload straight into `types.BridgesResult` — the direct cross-version regression. The three pre-existing marshal tests still pass untouched, so write-side behavior is provably unchanged.

Since the tolerance lives on the `BigIntString` type itself, it covers every field of that type, not just `global_index`: all seven `BigIntString`-typed fields on this branch — the six that already used the type (`Amount` on `BridgeResponse`/`ClaimResponse`/`LegacyTokenMigrationResponse`, and `GlobalIndex` on `ClaimResponse`/`UnsetClaimResponse`/`SetClaimResponse`) plus `BridgeResponse.GlobalIndex`, which this PR retypes to `BigIntString`.

**What stands, unchanged:** Q1 — the swagger needs no update, the schema always declared `global_index` as `type: string`, regeneration is still byte-identical. The marshal-side reasoning that a server can't emit both forms at once. The bridge-hub-api analysis — it never reads `global_index` off `/bridges`, so it's unaffected either way. The emitted form is still the quoted string.

Thanks for catching this.

joanestebanr (approved):
After running a smoke test against the current Bridge Service instances (with the field represented as an integer), everything looks good on my end. Approving the PR.

PR #1844, opened 2026-09-16
feat(bridgetracker): parallelize activity/tracker scans and make bridge-service timeout configurable. Fix skip step

### LINKED ISSUE #1843: feat: proxy: allow configure timeout for bridge calls
# Feature Request
- **Allow configure timeout for bridge calls**: if a service is failing now it takes 30 seconds
- **`activitiy` end-point must call in parallel**: To ensure that the total time of a request is not too much

joanestebanr:
## 🔄 Changes Summary
- `GET /activity/from/{from_address}` scanned every configured network's aggkit-bridge service sequentially (`bridgetracker/sources/activity.go`'s `BridgesFrom`); now scans up to `Tracker.ActivitySourceBridgeService.MaxConcurrentNetworkScans` (default 10) networks concurrently at once, so one slow/misbehaving network no longer serializes the whole request behind it, without letting the in-flight call count grow unbounded with however many networks are configured.
- The tracking engine's poll `tick` (`bridgetracker/engine.go`) resolved every active supervised bridge one at a time; now resolves up to `Tracker.MaxConcurrentResolutions` (default 50) of them concurrently at once. Safe because `SupervisedStore`'s in-memory implementation already guards all its state with its own mutex; bounded because the supervised list can hold up to `MaxTrackedBridges` (100,000 by default) entries, and resolving all of them at once would fire that many simultaneous outbound source calls.
- Added `Tracker.ActivitySourceBridgeService.Timeout` (`bridgetracker/config.go`): the per-HTTP-call timeout the activity source's aggkit-bridge REST client uses (`GetBridges`, `GetClaims`, `GetSyncStatus`, `GetL1InfoTreeIndex`, `GetInjectedL1InfoLeaf`). Previously this was hardcoded to `bridgeservice/client.DefaultTimeout` (30s) with no way to change it. Defaults to 30s when unset, so behavior is unchanged unless configured. Other bridge-service consumers (certificate/claim/GER sources, used by the tracker's own resolution path) are unaffected — they keep falling back to the client's own hardcoded default.
- Fixed the claimed-bridge fallback (`bridgetracker/domain/resolve_steps.go`'s `skipToClaimed`) forcing `ErrorType` to `StepErrorSkipped` on every step it rescued, including the one whose real `Transient`/`Permanent` error actually triggered the fallback — e.g. a step like `WaitL1SettledGER` failing with a genuine unresolved error would show `status: skipped` **and** `error.error_type_string: "skipped"`, indistinguishable from a step that was simply never attempted. That step now keeps its own real `ErrorType`.
- Went a step further: a step skipped alongside the real failure, but whose resolver was never even called, now omits `error` entirely instead of a placeholder `StepErrorSkipped` entry — there is nothing to report for it. `StepErrorType.StepErrorSkipped` had no other producer, so it is removed from the type, and `types.CurrentAPIRevision` is bumped 3 → 4 to flag the wire contract change; the bridgetracker swagger docs are regenerated to match.
- And further still: that same never-attempted step also had its `end_date` stamped to "now", despite never having a `start_date` — an end time for something that never started. It now keeps both dates nil, same as any other step that hasn't run yet; only the step that actually ran and triggered the fallback gets an `end_date`.

- Bounded both of the concurrency changes above (flagged by Codex review on this PR): unbounded goroutine fan-out in `Engine.tick` (one per active bridge, capped by nothing) and in `BridgesFrom` (one per configured network, each fanning out two more internally) could otherwise scale outbound calls directly with registry size / network count instead of a fixed cap.

- Bumped `google.golang.org/grpc` v1.82.1 → v1.83.2, fixing the two CVEs the CI `govulncheck` job flagged (GO-2026-6443, fixed in v1.82.2; GO-2026-6348, reachable from `aggsender/validator`'s generated grpc client, fixed in v1.83.1). `go mod tidy` pulled forward the usual transitive google-cloud-go/otel/x-net versions grpc/genproto depend on; no source changes needed.

## ⚠️ Breaking Changes
- None. All new behavior is additive/opt-in; the timeout defaults to the previous hardcoded value.
- Wire contract change (bumped `CurrentAPIRevision` 3 → 4): for a step rescued by the claimed-bridge fallback, `error.error_type`/`error_type_string` now reports the real cause (`"transient"`/`"permanent"`) instead of always `"skipped"`; a step skipped alongside it that was never attempted now omits `error` entirely instead of a placeholder `"skipped"` entry. `status: "skipped"` itself is unchanged, and `StepErrorSkipped` no longer exists as a possible `error_type` value. A never-attempted skipped step also now omits `start_date`/`end_date` (both stay absent, rather than a spurious `end_date`).

## 📋 Config Updates
- 🧾 New optional keys:
  ```toml
  [Tracker]
  MaxConcurrentResolutions = 50

  [Tracker.ActivitySourceBridgeService]
  Timeout = "10s"
  MaxConcurrentNetworkScans = 10
  ```

## ✅ Testing
- `go build ./...`
- `go test -race ./bridgetracker/...`
- `golangci-lint run ./bridgetracker/...`
- Updated `TestResolveStepsOnTerminalStepErrorIsRescuedWhenClaimed`, `TestResolveStepsSkipsOnAlreadyClaimed`, `TestResolveStepsRecoversAfterResolverTimeout` and `TestEngine...` (bridgetracker) to assert the rescued step's real `ErrorType` is preserved and that a never-attempted skipped step carries no `Error` or dates
- Added `TestEngineTickBoundsConcurrentResolutions` and `TestActivitySource_BridgesFrom_BoundsConcurrentNetworkScans` to pin the new concurrency caps
- Confirmed with `govulncheck` locally that both grpc CVEs are gone after the bump

- Regenerated bridgetracker swagger docs (`bridgetracker/api/docs/`) to drop `StepErrorSkipped`

## 🐞 Issues
- Closes #1843

## 📝 Notes
- The RPC-based activity fallback scanner still has no timeout of its own (bounded only by whatever context it's handed); out of scope for this change, called out separately if worth a follow-up.


joanestebanr:
@claude review

arnaubennassar (approved):

Issue #1848, opened 2026-09-16
[FEATURE] Rename the BlockFinality config fields so that different semantics get different names


arnaubennassar:
# Feature Request

## Is your feature request related to a problem? Please describe

Almost every component exposes a config field called `BlockFinality` (or a close variant), all typed `aggkittypes.BlockNumberFinality`, all documented with the same enum list — but they mean at least three different things:

|                   Field                   |                                   What it actually controls                                    |
| :---------------------------------------: | :--------------------------------------------------------------------------------------------: |
|    `L1Multidownloader.BlockFinality`      | **Reorg-safety boundary**: below it the multidownloader stops re-verifying headers and marks blocks final; above it is the "unsafe zone" that is re-checked every step. Shared by every syncer registered on it. |
|      `ReorgDetector.FinalizedBlock`       | Same role as above for the legacy sync path.                                                   |
| `L1InfoTreeSync.BlockFinality`, `BridgeSync.BlockFinality`, `L2GERSync.BlockFinality`, `ClaimSync.BlockFinality`, `BridgeServiceFinder.BlockFinality`, `BridgeTracker.L1BlockFinality` / `L2BlockFinality` | **Sync target** (`SyncerConfig.ToBlock`): how far the syncer wants blocks downloaded and processed. Does **not** affect reorg checking. |
| `AggSender.BlockFinalityForL1InfoTree`, `AggSender.Validator.BlockFinalityForL1InfoTree` | **Read-side selection**: which L1 info tree root aggsender picks to prove against (must be acceptable to agglayer-node). Does not affect what is synced. |

Because they share a name, operators (and code) reasonably assume they are the same thing and should be set to the same value. They are not, and setting them equal is actively harmful: lowering `L1Multidownloader.BlockFinality` to match a relaxed syncer target disables reorg detection above that tag for every syncer, while raising a syncer target to match the multidownloader's boundary just makes ingestion slower for no safety gain. The intended configuration is asymmetric (conservative reorg boundary, relaxed sync targets, conservative proof-root selection), and the naming hides that.

The confusion has also leaked into code: `L1InfoTreeSync.Finality()` returns the multidownloader's boundary in one mode and the syncer's target in the other (#1846), and aggsender's startup guard compares its own field against whichever one it gets.

## Describe the solution you'd like

Give each semantic its own name and keep the shared type. Suggestion (names are a starting point):

[...]

Issue #1706, opened 2026-07-08
bug: v0.10.0-rc6: failed to build ceritificate GET exists on L1 but cannot be proved against selected root


joanestebanr:
# Bug Report
The code must discard the bridges that can't be proved against choosed *L1InfoTreeRoot*. Maybe this case there was no more bridges or maybe the code have an issue

```
error getting certificate build params: ppFlow - error adjusting block range: GER 0xd7477e3bcffa9538258628f7475e3beef1ce7a394f0d708bc1dc48306ef9e557 exists on L1 but cannot be proved against selected root 0x27adf3e42c1a166ae5d9c7686d67b98582db3cb7dd6d741f92128bec272e9f03: ger is not provable against the selected root: invalid proof: calculated root 0xade812fce47e23a622240c84cf6b14e973485db87fa5f01bafed30499db57305 does not match expected root 0x27adf3e42c1a166ae5d9c7686d67b98582db3cb7dd6d741f92128bec272e9f03
````
Running v0.10.0-rc6 with next relevant configuracion: 

```
[L1InfoTreeSync]
BlockFinality = "FinalizedBlock"
``` 

Adding this configuration fix the problem [here](https://0xpolygon.slack.com/archives/C07J2T5QBM4/p1783455012572099?thread_ts=1783425385.163769&cid=C07J2T5QBM4): 
```
[L1InfoTreeSync]
    BlockFinality = "LatestBlock/-12"

    [AggSender]
    TriggerCertMode = "ASAP"
    BlockFinalityForL1InfoTree = "LatestBlock/-12"

    [L1Multidownloader]
    Enabled = true
    BlockFinality = "LatestBlock/-12"
```

Thiago suggest: 
| would it make sense to review default values? I'd expect all l1 block finality related fields reflect the value from  L1InfoTreeSync.BlockFinality

joanestebanr:
Binding the configurations is being ruled out because the L1InfoTreeSync.BlockFinality field is used by aggoracle to determine the finality of the GERs to inject.

PR #1573, opened 2026-04-08
refactor(aggsender): unify block range adjustment

arnaubennassar:
## 🔄 Changes Summary
- Unify aggsender certificate block-range adjustment under a single `AdjustBlockRange` flow path.
- Move max-L2 limiting, finalized-root claim provability checks, certificate-size trimming, and invalid-claim posterior-unclaim validation into the same ordered pipeline.
- Remove the old standalone max-L2 limiter and the `CertificateBuildParams` range mutator helpers, then update PP and aggchain prover flows and mocks to use the unified adjuster.
- Rewrite and extend aggsender tests to cover the new adjustment flow and root-based behavior.

## ⚠️ Breaking Changes
- 🛠️ **Config**: None.
- 🔌 **API/CLI**: Internal aggsender flow interfaces now use `AdjustBlockRange`; `LimitCertSize`, `Range`, `AdjustToBlock`, and the standalone max-L2 limiter path were removed.
- 🗑️ **Deprecated Features**: Removed `MaxL2BlockNumberLimiter` and its dedicated tests.

## 📋 Config Updates
- 🧾 **Diff/Config snippet**: None.

## ✅ Testing
- 🤖 **Automatic**: `go test ./aggsender/...`
- 🖱️ **Manual**: Not run.

## 🐞 Issues
- None.

## 🔗 Related PRs
- None.

## 📝 Notes
- Claim filtering for finalized data is now driven by `L1InfoTreeRootFromWhichToProve`: proposer paths select the finalized root up front, while validator paths assert that the provided root is finalized before accepting the range.
- Invalid-claim/unclaim validation remains the last adjustment step so earlier truncations cannot silently invalidate the final certificate range.

PR #1849, opened 2026-09-16
fix: make L1 finality knobs independent — Finality() passthrough (#1846) and claim trimming (#1847)

### LINKED ISSUE #1846: [BUG] L1InfoTreeSync.Finality() returns the multidownloader's BlockFinality instead of the syncer's own target in multidownloader mode
# Bug Report

## Description

`L1InfoTreeSync.Finality()` is supposed to report the finality the syncer syncs up to (its `SyncerConfig.ToBlock`, i.e. `L1InfoTreeSync.BlockFinality`). In multidownloader mode it reports `L1Multidownloader.BlockFinality` instead, which is a different knob with a different meaning (the multidownloader's reorg-safety boundary, not the download target).

Call chain:

- `L1InfoTreeSync.Finality()` → `d.downloader.Finality()` ([`l1infotreesync/l1infotreesync.go`](https://github.com/agglayer/aggkit/blob/develop/l1infotreesync/l1infotreesync.go#L231-L234))
- multidownloader path: `mdrsync.EVMDownloader.Finality()` → `d.multidownloader.Finality()` → `dh.cfg.BlockFinality` ([`multidownloader/sync/evmdownloader.go`](https://github.com/agglayer/aggkit/blob/develop/multidownloader/sync/evmdownloader.go#L53-L55), [`multidownloader/evm_multidownloader_syncers.go`](https://github.com/agglayer/aggkit/blob/develop/multidownloader/evm_multidownloader_syncers.go#L160-L162))
- legacy path: `sync.EVMDownloader.Finality()` → `d.finality`, which *is* `L1InfoTreeSync.BlockFinality` ([`sync/evmdownloader.go`](https://github.com/agglayer/aggkit/blob/develop/sync/evmdownloader.go#L122-L124))

So the two code paths disagree, and the multidownloader one is wrong: the multidownloader downloads every registered syncer up to that syncer's own `ToBlock` regardless of `dh.cfg.BlockFinality` (see `mapBlockTagToBlockNumber`), so `dh.cfg.BlockFinality` says nothing about how far l1infotreesync actually syncs.

## Impact

The only consumer is the aggsender startup guard in [`aggsender/query/l1info_tree_data_query.go`](https://github.com/agglayer/aggkit/blob/develop/aggsender/query/l1info_tree_data_query.go#L41-L51), which refuses to start if `AggSender.BlockFinalityForL1InfoTree` is less final than `l1InfoTreeSyncer.Finality()` ("block finality misconfiguration ... will never be fulfilled"). With the bug, a perfectly valid configuration is rejected:

```toml
[L1Multidownloader]
BlockFinality = "FinalizedBlock"       # reorg-safety boundary, should stay conservative

[L1InfoTreeSync]
BlockFinality = "LatestBlock/-12"      # sync target

[AggSender]
BlockFinalityForL1InfoTree = "LatestBlock/-12"   # <= L1InfoTreeSync target, should be accepted
```

Aggsender aborts because it compares `LatestBlock/-12` against `FinalizedBlock` (the multidownloader's tag). The only way past the guard is to lower `L1Multidownloader.BlockFinality` too, which silently disables reorg checking above that tag for **every** syncer sharing the multidownloader. Operators end up degrading reorg safety to satisfy a check that is looking at the wrong field.

## To Reproduce

1. Enable `L1Multidownloader` with the default `BlockFinality = "FinalizedBlock"`.
2. Set `L1InfoTreeSync.BlockFinality = "LatestBlock/-12"` and `AggSender.BlockFinalityForL1InfoTree = "LatestBlock/-12"`.
3. Start aggkit with aggsender enabled.
4. Aggsender fails at startup with `block finality misconfiguration (LatestBlock/-12): l1infotreeSyncer finality (FinalizedBlock) is lower; will never be fulfilled`.
5. Same config with `L1Multidownloader.Enabled = false` (legacy path) starts fine.

## Expected behavior

`L1InfoTreeSync.Finality()` returns `L1InfoTreeSync.BlockFinality` (the syncer's `ToBlock`) in both modes. Either `mdrsync.EVMDownloader.Finality()` should return `syncerConfig.ToBlock` rather than the multidownloader's finality, or `L1InfoTreeSync.Finality()` should return `d.cfg.BlockFinality` directly.

## Additional context

`L1Multidownloader.BlockFinality`, `L1InfoTreeSync.BlockFinality` and `AggSender.BlockFinalityForL1InfoTree` are three independent knobs (reorg boundary / download target / proof-root selection) and are meant to be configurable independently. This bug is what makes it *look* like they must all be set to the same value.

### LINKED ISSUE #1847: [BUG] aggsender: adjustClaimsNotProvableAgainstRoot errors instead of trimming claims whose GER is not yet under the selected L1 info root
# Bug Report

## Description

`baseFlow.adjustClaimsNotProvableAgainstRoot` ([`aggsender/flows/adjust_block_range.go`](https://github.com/agglayer/aggkit/blob/develop/aggsender/flows/adjust_block_range.go#L213-L247)) is part of the `AdjustBlockRange` pipeline and, by name and position, is expected to *adjust* the certificate range when a claim references a GER that cannot be proved against `L1InfoTreeRootFromWhichToProve`. It never does: when `GetProofForGER` returns `ErrGERNotProvableAgainstRoot` and the GER does exist on L1, it returns a hard error (`GER %s exists on L1 but cannot be proved against selected root %s`) and the whole certificate build fails.

`L1InfoTreeDataQuerier.IsGERFinalized` ([`aggsender/query/l1info_tree_data_query.go`](https://github.com/agglayer/aggkit/blob/develop/aggsender/query/l1info_tree_data_query.go#L218-L231)) exists on the `L1InfoTreeDataQuerier` interface precisely for this decision, but has no callers in `aggsender/`.

## Why this happens in a correctly configured setup

The supported configuration is:

- `L1InfoTreeSync.BlockFinality` relaxed (e.g. `LatestBlock/-N`) so aggoracle can inject new GERs on L2 quickly.
- `AggSender.BlockFinalityForL1InfoTree = FinalizedBlock` so the root the certificate proves against is one agglayer-node will accept.

With that, there is always a window (the L1 finality delay) where a GER has already been injected on L2 and used by claims, but its L1 info tree leaf is not yet under the finalized root aggsender selects. Any claim in that window makes the entire certificate fail instead of being deferred to the next certificate. With frequent certificates (e.g. `TriggerCertMode = "ASAP"`) the window is hit almost continuously and aggsender stalls until operators lower `AggSender.BlockFinalityForL1InfoTree`, which defeats the point of the separate knob.

## To Reproduce

1. Configure `L1InfoTreeSync.BlockFinality = "LatestBlock/-12"` and `AggSender.BlockFinalityForL1InfoTree = "FinalizedBlock"` (aggoracle reading from the same l1infotreesync).
2. Bridge L1 → L2 so aggoracle injects a fresh GER, and claim it on L2 before the L1 block containing that GER is finalized.
3. Let aggsender build the next PP certificate covering that L2 claim.
4. Certificate build fails with `ppFlow - error adjusting block range: GER <ger> exists on L1 but cannot be proved against selected root <root>: ger is not provable against the selected root: invalid proof: calculated root <x> does not match expected root <root>`, and keeps failing until the GER's L1 block is finalized.

## Expected behavior

Claims whose GER is not yet provable against the selected root are trimmed from the certificate: reduce `ToBlock` to the last L2 block whose claims are all provable against `L1InfoTreeRootFromWhichToProve` (using `IsGERFinalized` / the leaf index vs. the selected root's leaf count), and only return an error when nothing provable remains in the range (or when the offending claim is at `FromBlock` and the certificate cannot be shrunk). The trimmed claims are naturally picked up by a later certificate once the root has advanced.

A GER that exists on L1 but is not under the selected root is an expected transient state in this configuration, not an inconsistency.

## Additional context

This is the aggsender-side half of making the finality knobs independent; the l1infotreesync-side half is the `L1InfoTreeSync.Finality()` passthrough bug filed separately. Related: #1706.

arnaubennassar:
## 🔄 Changes Summary
This PR makes the three L1 finality knobs of aggkit (`L1Multidownloader.BlockFinality`,
`L1InfoTreeSync.BlockFinality`, `AggSender.BlockFinalityForL1InfoTree`) actually independent, fixing
two bugs filed from the spec ASAP-mode incident (`history/2026-09-16-aggkit-block-finality-semantics-asap-spec.md`
in the agglayer-workspace repo):

- **#1846** — `L1InfoTreeSync.Finality()` returned `L1Multidownloader.BlockFinality` in multidownloader
  mode instead of the syncer's own `BlockFinality`. The only consumer, the aggsender startup guard
  (`aggsender/query/l1info_tree_data_query.go:41-51`), therefore rejected valid configs such as
  `{Multidownloader=Finalized, L1InfoTreeSync=Latest/-12, AggSender=Latest/-12}` with
  `block finality misconfiguration ... will never be fulfilled`. Fix: `Finality()` now returns
  `d.cfg.BlockFinality` in both constructors (legacy and multidownloader).
- **#1847** — `baseFlow.adjustClaimsNotProvableAgainstRoot` hard-errored (`GER %s exists on L1 but
  cannot be proved against selected root %s`) instead of trimming the certificate when a claim's GER
  exists on L1 but is not yet under the selected L1 info root. With `L1InfoTreeSync` relaxed and
  `AggSender` conservative, this window is hit continuously in ASAP mode and aggsender stalls. Fix:
  the offending claims are now trimmed (mirroring the pre-#1573 `adjustCertificateIfNonFinalizedClaims`
  behaviour, which used `IsGERFinalized` and had zero callers since PR #1573 replaced it with the hard
  error), erroring only when the offending claim is at `FromBlock` (nothing provable remains), the GER
  is under the root but the proof still fails (real inconsistency, kept as a hard error), or
  `L1InfoTreeLeafCount == 0` (defensive, avoids a `uint32` underflow in `IsGERFinalized`).

**No validator code was needed.** `CertificateValidator.ValidateCertificate` builds its pre-params from
the proposer-*declared* range and rebuilds + full-diffs the certificate against the received one, so a
proposer that trimmed is validated on exactly the range it declared, and a proposer that did not trim is
already rejected (`"certificate not equal to expected: certificates mismatch..."`). This was verified in
review (see Notes) rather than assumed.

## ⚠️ Breaking Changes
None. Behaviour change: certificates are now **trimmed** (to the last L2 block before the first claim
whose GER is not yet under the selected L1 info root) instead of the certificate build failing outright.

## 📋 Config Updates
No config key renames (aggkit#1848, which renames the `BlockFinality` fields, is explicitly out of
scope for this PR). Recommended asymmetric configuration to make ASAP-mode aggsender resilient to L1
reorg-safety boundaries without stalling:

```toml
[L1Multidownloader]
BlockFinality = "FinalizedBlock"   # reorg-safety boundary shared by every syncer

[L1InfoTreeSync]
BlockFinality = "LatestBlock/-N"   # N must stay below the chain's latest-to-finalized lag

[AggSender]
BlockFinalityForL1InfoTree = "FinalizedBlock"   # which L1 info tree root aggsender proves against
```

## ✅ Testing
**Unit tests** (all in `./aggsender/...` and `./l1infotreesync/...`, all green):
- `TestFinality_MultidownloaderModeReturnsOwnBlockFinality` (`l1infotreesync/l1infotreesync_test.go`) —
  direct proof of the #1846 fix: constructs the syncer in multidownloader mode with
  `cfg.BlockFinality = LatestBlock/-12` and a mocked multidownloader whose own `Finality()` returns
  `FinalizedBlock`, and asserts `Finality()` returns the syncer's own cfg value, not the
  multidownloader's.
- `Test_baseFlow_adjustClaimsNotProvableAgainstRoot_*` (`aggsender/flows/adjust_block_range_test.go`) —
  10 cases covering: single/multiple not-yet-provable claims trimmed to the minimum offending block;
  offending claim at `FromBlock` → hard error; GER under root but proof still fails → hard error kept
  (unchanged message); `L1InfoTreeLeafCount == 0` → hard error; GER absent on L1 → untouched (left to
  `adjustInvalidClaimsAreNotUnclaimed`); cache reuse.
- Four existing tests that previously asserted the hard error updated to the new trimming behaviour:
  `flow_base_test.go`, `builder_flow_pp_test.go` (trims to `[6,9]`),
  `builder_flow_aggchain_prover_test.go` (retry path, trims to `[1,9]` with the proof regenerated,
  confirming the retry-cert path is trim-safe).

**E2E**: `TestAggsenderIndependentL1FinalityKnobs` (`test/e2e/finality_knobs_test.go`), env
`anvil-2chains`, CI matrix group `finality-knobs` (anchored regex
`^TestAggsenderIndependentL1FinalityKnobs$`). It restarts `aggkit-001` with
`L1InfoTreeSync.BlockFinality = "LatestBlock"` and `AggSender.BlockFinalityForL1InfoTree =
"LatestBlock/-N"` (multidownloader left at the default `FinalizedBlock`), performs L1→L2 bridge+claims
inside the finality window, and asserts both the trim warning appears and the hard-error string never
does, then asserts eventual certificate settlement. All sizing (`N`, timeouts) is derived from a runtime
measurement of the anvil L1 block time and latest→finalized lag, not hardcoded.

**Local red→green evidence** (both runs with verified, distinct image IDs — GREEN `sha256:e47f1b20…`,
RED `sha256:3705eee9…` — and the pre-fix tree confirmed genuinely pre-fix: `git diff` against
`origin/develop` on the temporary pre-fix worktree contained only the three test/CI/docs files, and
`l1infotreesync/l1infotreesync.go:233` still read `return d.downloader.Finality()`):

- **GREEN — PASS, 164.71s** (image `e47f1b20…`). Step-1 measured L1:
  `latest=254 finalized=252 lag=2 blockTime=1s windowBlocks=121 N=127 trimWait=5m9s settleWait=9m18s`.
  Neither `block finality misconfiguration` nor the transient
  `error getting latest l1 info tree info until block num` appeared; the trim warning appeared; a
  certificate covering the claimed block settled.
- **RED — FAIL, 26.99s** (image `3705eee9…`), crash-looping **7×** with `RestartCount` climbing, verbatim
  from the `aggkit-001` container log:
  `cmd/run.go:322 FATAL failed to create AggSender: … error creating L1 Info tree data querier: block
  finality misconfiguration (LatestBlock/-127): l1infotreeSyncer finality (FinalizedBlock) is lower;
  will never be fulfilled`.

The e2e genuinely exercises **both** issues on the multidownloader path: `aggkit-001` runs the
multidownloader (confirmed via `Enabled = true` default with no `[L1Multidownloader]` override, and
live in the container log: `cmd/run.go:695 L1 Info Tree Syncer using MultiDownloader based
implementation`), so #1846's guard genuinely crash-loops pre-fix and #1847's trim genuinely fires and
settles post-fix.

## 🐞 Issues
- Closes #1846
- Closes #1847
- Related #1706
- Related #1848

## 🔗 Related PRs
- None

## 📝 Notes
- **Why the #1846 fix lives in `L1InfoTreeSync.Finality()` and not the multidownloader**:
  `mdrsync.EVMDownloader` does not store the per-syncer `SyncerConfig` (it is passed per call), so
  fixing it there would require a constructor change shared by several syncers. Fixing at
  `L1InfoTreeSync.Finality()` — returning `d.cfg.BlockFinality`, which both constructors already
  store — is the surgical, contained option.
- **Why a GER unknown to the syncer keeps the existing hard error**: `GetProofForGER` returns
  `ErrGERNotProvableAgainstRoot` only when `tree.VerifyProof` fails on a GER the syncer does know
  about; a GER unknown to the syncer surfaces as a different error ("error getting info by global exit
  root") and is out of scope for #1847 — it is a genuine inconsistency, not a not-yet-finalized window.
- **Pre-#1573 history**: `adjustCertificateIfNonFinalizedClaims` trimmed via `IsGERFinalized` before PR
  #1573 replaced it with the hard error, leaving `IsGERFinalized` on the `L1InfoTreeDataQuerier`
  interface with zero callers. This PR restores that trimming behaviour (with new criteria matching
  today's code shape) rather than reintroducing the old function verbatim.
- **Why no validator change was needed**: `CertificateValidator.ValidateCertificate` builds its
  pre-build-params from the proposer-*declared* `FromBlock`/`ToBlock`, reruns `AdjustBlockRange`,
  rebuilds the certificate, and full-diffs it against the one received from the proposer. A proposer
  that trimmed is therefore validated on exactly the (already-trimmed) range it declared, and a
  proposer that failed to trim (or trimmed differently) is already rejected today as
  `"certificate not equal to expected: certificates mismatch..."`. No new validator code or test was
  needed or added.
- **Operator note**: under the recommended asymmetric config, a trim that empties a PP certificate
  yields `errNoBridgesAndClaims`, logged at **Error** and fed into the `RetriesToBuildAndSendCertificate`
  ladder — so Error-level log entries and a populated `aggsender_status` last-error become routine
  rather than rare. This is functionally correct (the retry ladder *is* the wait-for-root mechanism) and
  strictly better than the pre-fix hard failure. It was deliberately **not** silenced with an
  `IsEmpty()` early-return, since that would remove useful polling behaviour in the `NewBridge` and
  `EpochBased` trigger modes.
- The `L1InfoTreeLeafCount == 0` guard is a defensive hard error at the call site, added specifically to
  prevent reaching a `uint32` underflow (`leafCount - 1`) inside `IsGERFinalized`.
- **Pre-existing drift deliberately left alone** (out of scope, not touched or worsened): the stale
  `default` CI matrix group referenced in `docs/e2e_tests.md`, and a stale `BlockFinality` row (should
  read `BlockFinalityForL1InfoTree`) in the older Configuration table in `docs/aggsender.md`.

https://claude.ai/code/session_011pFZCNjin5V8DcrhEpuQmA


arnaubennassar:
@claude

arnaubennassar:
Thanks for the review. Three of the four suggestions check out as no-ops; one needed verifying. No code changes resulted.

**1. `IsGERFinalized == true` + proof failure — is that branch actually reachable?**

Yes, and it is deliberately kept as a hard error. `L1InfoTreeLeafCount` is `root.Index + 1`, so `IsGERFinalized(ger, leafCount)` is precisely "`leaf.Index <= root.Index`", i.e. the leaf lies within the selected root's range. `GetProofForGER` returns `ErrGERNotProvableAgainstRoot` only when `tree.VerifyProof` fails. So the branch means: the leaf is inside the root's range, yet a proof against that root does not verify — the local tree disagrees with its own root. That is reachable after an L1 reorg that rewrites leaves at or below that index once the root has already been selected. Trimming there would mask a genuine inconsistency and risk submitting a certificate that lands InError, so it stays loud.

Related and also deliberate: the loop keeps scanning after finding a trim candidate, so a *later* under-root-but-unprovable claim still errors even though the trim would have dropped it. Pre-fix every one of these inputs hard-errored, so the post-fix erroring set is a strict subset of the pre-fix one — this introduces no new stall vector.

**2. Cache across adjustment steps** — agreed, correct as-is. One addition worth recording: trim candidates are cached `existsOnL1 = true` before the trim, and a claim *below* the trim point whose GER is missing on L1 can still trigger a second, independent trim in `adjustInvalidClaimsAreNotUnclaimed`. That is intended, not a double-trim of the same cause.

**3. Leaf-count-zero error message length** — leaving as-is. The root hash and leaf count are what make it diagnosable in a log, and the lines stay inside the repo's 120-char limit.

**4. Mock generation consistency** — verified, nothing to do, though the premise is slightly off: `mock_downloader_interface.go` was deleted and *not* replaced, and the test does not use `NewDriverInterfaceMock`. `.mockery.yaml:149-156` configures this package with `all: true`, so removing the interface simply stops that file being generated. I re-ran `make generate-mocks` to confirm: no downloader mock reappears under `l1infotreesync/`. The only diffs it produces are a mockery version banner (`v2.53.3` → unversioned) across six files plus one uncommitted `bridgetracker` mock — all pre-existing on `develop` and unrelated to this PR, so they are not included here.

Unrelated to the review: `govulncheck` was failing on two pre-existing grpc CVEs (`GO-2026-6443`, `GO-2026-6348` in `google.golang.org/grpc@v1.82.1`, both reachable from the validator gRPC client and server). Bumped to v1.83.2 in 08a90eff; `govulncheck` is now green.

joanestebanr (approved):

Issue #1850, opened 2026-09-16
multidownloader: eth_getLogs block-range-cap errors are never recognised, so an oversized range is retried forever


arnaubennassar:
## Summary

`multidownloader` does not recognise explicit `eth_getLogs` **block-range cap** errors, so such an error propagates out of its shrink logic and the surrounding infinite retry loop re-issues the *same oversized range* forever. It already defaults to `BlockChunkSize = 10000` (`config/default.go:429`), so a range-capped RPC provider can wedge an L1/L2 multidownloader today.

## Detail

`isEthClientErrorTooManyResults` (`multidownloader/evm_multidownloader.go:879`) only matches *"too many results"*-style responses. On a match, `requestLogs` shrinks via `currentSyncBlockChunkSize /= chunkSizeReductionFactor` (10), floored at `minChunkSize` (`:1012-1013`). Any error that is instead an explicit range cap — e.g. `block range too large, max range: N`, `exceeded maximum block range: N`, `eth_getLogs is limited to a N range`, `query exceeds max block range N` — fails the check at `:997`, is returned as a hard error, and the chunk size is never reduced. With `MaxRetryAttemptsAfterError = -1` the same query is retried indefinitely.

Those four shapes are already parsed by `aggkitcommon.ParseMaxRangeFromError` (`common/errors.go`), which `multidownloader` does not use.

## Context

PR for #1842 consolidated every other `eth_getLogs` size/range handler into a single shared decision point, `aggkitcommon.NextEthGetLogsWindow` (`common/errors.go`), and wired it into:

- `sync/evmdownloader.go` (shared by BridgeL1Sync, BridgeL2Sync, ClaimL1Sync, ClaimL2Sync, L1InfoTreeSync legacy, L2GERSync main loops)
- `claimsync/claimsync.go` (settled-IBE RPC scan)
- `l2gersync/l2_evm_ger_reader.go` and `l2gersync/evm_downloader_sovereign.go`

`multidownloader` was deliberately left out of that change because fixing it means editing a retry loop coupled to its own batching, partial-availability and reorg-check logic, which warrants a dedicated test pass. It is **not** a regression from that PR — its `BlockChunkSize = 10000` default predates it and was not modified.

## Suggested fix

[...]

PR #1851, opened 2026-09-16
fix(aggsender): bounded fallback for claim syncer start block when settled IBE can't be resolved (#1842)

### LINKED ISSUE #1842: aggsender recovery hard-fails forever when an imported bridge exit's block can't be resolved for an already-settled certificate
## Summary

On startup/recovery, aggsender computes the claim syncer's next-required-block by resolving
block numbers for three settlement sources from the latest **settled** AggLayer certificate,
one of which is the block of the certificate's imported bridge exit (IBE), resolved via the
local `l2ClaimSyncer` with an RPC fallback if it isn't found locally.

If **both** the local lookup and the RPC fallback fail to resolve that block number, aggsender
treats this as a hard, unrecoverable error and retries forever without making progress — it
never starts its certificate-sending loop. Since the certificate in question is already
`Settled` on the AggLayer, this unresolved historical detail shouldn't be blocking: the block
number is only used to choose a safe starting point for the claim syncer, not to re-validate
an already-settled certificate.

## Where this happens

- `aggsender/query/certificate_query.go` — `getBlockNumFromGlobalIndex` (around line 210-229)
  looks up the claim in the local `l2ClaimSyncer` and returns an error
  (`"no claim found for bridge exit hash %s"`) if no match exists.
- `aggsender/query/initial_block_to_claimsync_setter.go` —
  `claimSyncerStartingBlockBasedOnLatestSettledCert` (lines 99-126) already has one fallback:
  if the local lookup errors, it retries via `l2ClaimSyncer.GetLatestBlockNumByGlobalIndexFromRPC`
  (lines 106-120). But if *that* also fails or returns not-found, the function returns an error.
- `aggsender/types/types.go` — `SettledBlocks.EarliestBlock()` (lines 423-451) unconditionally
  propagates `LastImportedBridgeExitBlockErr` as fatal (lines 432-434), with no path to proceed
  without it.
- That error bubbles up through `claimSyncerStartingBlock` into
  `SetInitialBlockToClaimSyncer.SetClaimSyncerNextRequiredBlock` (lines 38-80 in the same file),
  which retries with `MaxAttemptsInfinite` — so the process spins on this error indefinitely
  instead of degrading gracefully or surfacing a way to recover.

## Why this can happen

The IBE's block number can be genuinely unresolvable from either source when the claim syncer's
local database doesn't cover the relevant historical range (e.g. it was reset/re-seeded and
hasn't backfilled that far back, or was never running long enough to observe the original claim
event) and the RPC fallback can't locate it either (e.g. due to node pruning, indexing gaps, or
the configured RPC not exposing that historical range). This is a data-availability gap on the
querying side, not evidence that anything is wrong with the already-settled certificate itself.

## Suggested fix

aggsender's recovery path should tolerate an unresolved imported-bridge-exit block number for
an **already-settled** certificate instead of hard-failing forever. Some options, not mutually
exclusive:

1. Don't gate claim-syncer startup on resolving `LastImportedBridgeExitBlock` when the
   certificate status is already `Settled` — it's only needed to pick a safe starting point,
   not to re-validate the certificate.
2. When unresolved, fall back to a conservative starting point derived from the other resolved
   sources (e.g. `LastBridgeExitBlock` / `LastSettledL2BlockNum`) rather than erroring out, with
   a clear warning log so the gap is visible instead of silently masked.
3. Emit a metric/log signal when this fallback path is taken, so a real indexing gap doesn't go
   unnoticed even though the process recovers.

## Impact

We've hit this class of failure more than once on integration devnets: aggsender gets stuck in
an infinite retry loop at startup and never resumes sending certificates, requiring manual
intervention. In at least one case, deleting the aggsender's local database was used as a
workaround — which is risky and not something we want to rely on, since it can itself change
recovery behavior in ways that are hard to predict.

arnaubennassar:
## 🔄 Changes Summary
When the L2 claim syncer DB is empty at startup, aggsender derives the claim syncer's start block from the
latest **settled** AggLayer certificate. If the settled imported bridge exit (IBE) block could not be
resolved from the local claim DB nor via an RPC log-scan fallback, `SetClaimSyncerNextRequiredBlock`
retried **forever** and aggsender never reached its certificate loop.

**Confirmed in production** (Loki, namespace `bali-82-op`, `v0.11.0-rc8`):

```
query/initial_block_to_claimsync_setter.go / common/retry_handler_delays.go:145
fails execution of Setting next required block for claim syncer based on agglayer's latest settled
certificate try N/INFINITE. delay 1s.  due to error: error getting next block number for claim
syncer: error getting last settled certificate to block:
no claim found for global index 18446744073710703653 via RPC fallback
```

This looped `try N/INFINITE` with **no termination**, ~12-14s per attempt (a single full-range
`eth_getLogs` probe, slow as well as fruitless). The RPC fallback had returned `found=false, err=nil` —
the L2 RPC simply could not return that claim log for this global index.

**The fix** adds a bounded resolution order for the claim syncer's start block, so a genuinely
unresolvable IBE degrades to a safe lower bound instead of spinning:

| # | Source | Result | Backfill cost |
|---|--------|--------|---------------|
| 1 | Local claim DB (`GetClaimsByGlobalIndex`) | exact | none |
| 2 | RPC backward **always-chunked** log scan up to `LatestBlock` (chunk = `SyncBlockChunkSize`, adaptive shrink on max-range errors) | exact | none |
| 3 | Provable lower bound from aggsender's own certificate storage: `FromBlock` of the locally-built certificate whose `ImportedBridgeExits` contains the settled IBE (**proposer only** — the validator has no certificate storage) | lower bound | small (that cert's range) |
| 4 | `0`, capped to `InitialBlockNum` by the claim syncer | lower bound | full history |

The fallback (tiers 3/4) only engages once the other two `SettledBlocks` sources
(`LastBridgeExitBlockErr`, `LastSettledL2BlockNumErr`) are healthy. If either of those is still
erroring, `EarliestBlock()` can't succeed regardless of the IBE, so the existing retry loop keeps
retrying those sources on their own — **transient bridgesync lag still retries exactly as it did
before this change**, it does not get short-circuited into the fallback.

Also raises `SyncBlockChunkSize`'s default and unifies `eth_getLogs` range-cap error handling so that
default change is safe on upgrade (see **Config Updates** below).

## ⚠️ Breaking Changes
Breaking changes: none.

## 📋 Config Updates
**Config updates: YES.** `SyncBlockChunkSize` default raised **100 → 10000** in `config/default.go` for
`[L1InfoTreeSync]`, `[BridgeL1Sync]`, `[BridgeL2Sync]`, `[L2GERSync]`, with `[ClaimL1Sync]`/`[ClaimL2Sync]`
inheriting the new value via templating.

⚠️ **This widens every syncer's `eth_getLogs` window, not just the claim-syncer startup scan.**
Operators on range-limited RPC providers may need to lower `SyncBlockChunkSize` back down; a loud WARN
now fires whenever the adaptive shrink kicks in, naming the configured chunk size.

This default change is safe because of the accompanying commit in this PR
(`fix(sync): unify eth_getLogs range-cap/too-many-results shrink handling across every downloader`),
which adds one shared decision point (`aggkitcommon.NextEthGetLogsWindow` in `common/errors.go`) and
wires it into every `eth_getLogs` path that previously lacked adaptive shrink or had a bespoke one
(`sync/evmdownloader.go` — inherited by `BridgeL1Sync`/`BridgeL2Sync`/`ClaimL1Sync`/`ClaimL2Sync`/
`L1InfoTreeSync`'s legacy downloader — plus both `l2gersync` downloader paths). Without that commit, a
range-capped provider would hard-stall a syncer on the new, larger default with no self-recovery
(`MaxRetryAttemptsAfterError = -1` retries the same oversized range forever).

**Known residual gap:** `multidownloader/evm_multidownloader.go` keeps its own shrink logic that does
not recognize the range-cap error family. This is pre-existing — its `BlockChunkSize = 10000` default
predates this PR and is unaffected by the `SyncBlockChunkSize` change here — and is tracked as a
follow-up: https://github.com/agglayer/aggkit/issues/1850.

## ✅ Testing
**Unit tests** — red/green verified per fix:
- `aggsender/query`: setter fallback tiers (RPC not-found, RPC exhaustion, storage lower bound, small
  validator-shaped retry budgets), `aggsender/types` `EarliestBlock()` cases.
- `claimsync/storage`: idempotent `InsertClaim`/`InsertUnsetClaim`/`InsertSetClaim` on `UNIQUE
  constraint failed: claim.block_num, claim.block_pos` (mirrors the guard `InsertBlock` already had).
- `common/errors.go`'s `NextEthGetLogsWindow`, and per-call-site tests in `sync/evmdownloader_test.go`,
  `l2gersync/evm_downloader_sovereign_test.go`, `l2gersync/l2_evm_ger_reader_test.go` — including a
  red/green pair on `sync/evmdownloader.go` proving it previously retried the same oversized range
  forever on a range-cap error and now shrinks and completes.

**E2E**: `TestAggsenderClaimSyncerRecoveryAfterDBWipe` (`test/e2e/claimsync_recovery_test.go`), env
`anvil-2chains`, own `claimsync-recovery` CI matrix group in `.github/workflows/test-go-e2e.yml`. It
wipes the claim syncer DB after a claim has settled and asserts aggsender recovers with a real,
IBE-bounded start block instead of spinning. Decisive evidence from the passing run:

```
Set next required block for claim syncer to 175. settled blocks: SettledBlocks{LastBridgeExitBlock: 0,
LastImportedBridgeExitBlock: 175, LastSettledL2BlockNum: 0, SettledImportedBridgeExit:
{GlobalIndex: 18446744073709551618, BridgeExitHash: 0x32d77376a6660f779cbfbbdee96cefa31997365ae6e0231690fbc5de7e3a06b3}}
```

`0 < N <= C` (`0 < 175 <= 175`), retry-failure count (`"fails execution of Setting next required
block..."`) = 0, `UNIQUE constraint failed: claim.block_num` count = 0 (down from ~186 occurrences in
the pre-storage-fix run, where the live-lock meant the post-restart "Set next required block" line
never appeared at all and the test failed on a 3-minute timeout).

## 🐞 Issues
- Closes #1842

## 🔗 Related PRs
- Follow-up: https://github.com/agglayer/aggkit/issues/1850 — `multidownloader` still has its own
  shrink logic that does not recognize range-cap errors (pre-existing, not introduced by this PR).

## 📝 Notes

**New metric**: `aggsender_claim_syncer_start_block_fallback_total{reason}`, reasons
`ibe_not_found_on_rpc` (RPC scan completed, no matching claim found) and `ibe_rpc_lookup_exhausted`
(RPC scan failed consecutively as many times as the effective retry budget allows). **Operator hint**:
if this metric fires, check whether the L2 RPC node used for the claim syncer can still serve the
claim event log for the affected global index — this typically means the node is pruned, is the wrong
node, or is missing history the claim syncer needs.

**Two honest limitations** — do not overclaim what this PR fixes:

1. If the L2 RPC can never return the claim log (pruned node, wrong node), no starting-block strategy
   fixes that. `AggSender` will start, backfill from the lower bound, and then fail downstream with the
   existing `"no claim found for bridge exit hash ..."` error. What this PR fixes is *visibility*: that
   failure now surfaces as a WARN log plus the new metric, instead of an opaque startup spin with no
   forward progress.
2. With `AutoStart` enabled, the claim syncer's own `EVMDriver` backfills from `InitialBlockNum`
   regardless of the computed start block (`claimsync/claimsync.go`), so the e2e test does **not** prove
   "no unnecessary backfill" — it only proves the setter itself resolves a real settled reference
   instead of falling back to 0. That "avoid backfilling when possible" property is covered by the unit
   tests, not the e2e test.

**Deferred findings from the adversarial review** (`tmp/1842_review.md`, not part of this PR's diff):
- **F5**: `BigIntMeddler.PreWrite` persists a nil `*big.Int` as the literal text `"<nil>"`, which
  nothing can read back (`PostRead` fails to parse it). Currently unreachable (no producer creates a
  nil `Amount`), pre-existing, and broader than this PR — worth its own fix.
- F3 (RPC tier-2 fallback matches global index only and returns the most recent match — pre-existing,
  limited blast radius), F4 (broad `sqlite.ErrConstraint` check in `claimsync/storage/storage.go`,
  diagnostic-quality only, pre-existing pattern), and F8 (no overall deadline on the chunked RPC scan,
  requires a hostile RPC shape to trigger) were also reviewed and deferred as pre-existing or low-risk,
  per the review's own disposition.

https://claude.ai/code/session_01XHXsq1huvCimxDxJxNbpsM


arnaubennassar:
@claude

arnaubennassar:
Thanks for the review. Addressed in `2b01c69c`. Summary of what was taken, and what wasn't, with reasoning.

## Implemented

**S5 — overall scan deadline (the most valuable point here).** You're right that this was a real gap. Each `FilterLogs` call had `sync.DefaultFilterLogsTimeout`, but the *total* always-chunked scan was unbounded — so on a large range one attempt could run for a very long time even though no single call hung. That partially undermined the PR's own "never spin forever" goal, since `maxIBERPCLookupFailures` bounds the attempt *count*, not the duration of an attempt.

Added `maxRPCScanDuration = 5 * time.Minute` covering the whole scan, nested inside the caller's context so an existing deadline is never extended. On expiry it returns an **error**, not "not found" — that distinction is load-bearing: `(0, false, nil)` means "scan completed, no match" and makes the setter fall back immediately, whereas `(0, false, err)` is counted toward the retry budget and retried first. A timeout is the latter. The message names the overall deadline and the range scanned so far, so it's distinguishable from a single-call timeout. Worst case before fallback is now bounded at `5 × 5m = 25m`, versus unbounded previously.

**S1 — justified the retry budget.** Expanded the comment on `maxIBERPCLookupFailures = 5`. One correction: the "~12-14s per attempt" figure is from the production logs of the *old* single full-range `eth_getLogs` probe, which this PR replaces with an always-chunked scan. The comment now states the bound in terms of the current implementation (at most `maxRPCScanDuration` per attempt). Left as a constant deliberately — no new config surface.

**S2 — actionable WARN.** Appended an operator hint about pruned/missing L2 RPC history. Prefix unchanged, since `docs/aggsender.md` and the e2e test reference it.

**O1 — named the hardcoded retry delay** as `defaultRetryPeriod` (only occurrence in the repo).

## Not implemented

**R1 — the premise doesn't hold, so there's no fix to make.** `bigIntStr`/`bigIntCmp` are defined locally in `claimsync/storage/storage.go` (~L299 and ~L313), not in `aggsender/types`, and that file doesn't import `aggsender/types` — so there's no cross-package dependency. `bigIntCmp` also already handles nil correctly for its use as an equality check: two nils compare equal (`0`), nil-vs-non-nil compares unequal (`±1`). I did apply the cosmetic part — `bigIntStr(nil)` now renders `"nil (unexpected)"`, since a real claim's `GlobalIndex` is always ABI-decoded and never nil.

**S3 — deliberately-failing multidownloader test.** An expected-to-fail test would leave CI permanently red. The gap is tracked in #1850, which includes the red/green acceptance criteria for whoever picks it up.

**O3 — additional RPC lookup metrics.** Reasonable idea, but scope creep for a bugfix PR; better as a follow-up alongside #1850.

**S4** — agreed, no change needed: `head` is already threaded through to `chunkedRemovalScan` rather than re-resolved. (That double call did exist briefly during development and was caught by a failing mock expectation.)

**O2** — no action.

## One correction to the summary

The PR does **not** prove "no unnecessary backfill" via the e2e test. With `AutoStart` enabled the claim syncer backfills from `InitialBlockNum` regardless of the computed start block, so that property is covered by the unit tests only. The PR body states this; flagging it here so the claim isn't carried forward.

arnaubennassar:
it turns out that this was not a bug, but a genuine L2 reorg https://0xpolygon.slack.com/archives/C07J2T5QBM4/p1786624014339709

PR #1852, opened 2026-09-17
feat(bridgetracker): add SQLite-backed cache for tracker/activity state (#1822)

### LINKED ISSUE #1822: feat: proxy: add DB cache
## Goal

Add a DB-backed cache for the bridge tracker's **activity** and **tracking** state (today both live only in memory — `ActivityCache` in `bridgetracker/activity.go` and `memoryRegistry` in `bridgetracker/registry.go`), so state survives restarts and is queryable directly.

This is meant to:
- **Reduce calls to external services**: both caches already avoid re-asking for something already known within a run (a bridge already scanned, a claim already confirmed, a tracking snapshot already resolved), but every restart starts from zero and re-issues the full set of bridge-service/agglayer calls again. A DB-backed cache keeps that already-resolved state across restarts.
- **Increase response speed**: serving `GET /activity/...` and `GET /tracker/...` from a local DB read instead of re-scanning bridge services / re-resolving tracking steps on every cold start.
- **Keep timestamps**: `lastAccess` (idle eviction, both caches) and `terminalSince` (terminal-entry retention, `memoryRegistry`) are currently reset to zero on every restart, since they only exist in memory — losing them means every entry looks freshly seen/non-terminal again right after a restart, which throws off idle/retention pruning. Persisting them lets pruning behave correctly across restarts instead of just within a single process lifetime.

joanestebanr:
## 🔄 Changes Summary
- New `bridgetracker/db` package (mirrors `aggsender/db`): `sqliteRegistry` implements `domain.SupervisedRegistry` over a new `tracked_bridge` table; `sqliteActivityStore` implements `domain.ActivityQuerier` over new `activity_address`/`activity_bridge` tables. Both share one migration set and can point at the same SQLite file.
- Tracking data survives via a `schema_version` column: a row written under a different version is treated as a cache miss and discarded rather than risking a misinterpreted decode.
- Pruning/retention is deliberately left in-memory-only for now: the SQLite adapters' PruneTerminal/PruneIdle (and the activity idle sweep) are no-ops, so rows are never deleted yet — a conscious, temporary trade-off while the DB-side retention policy is still undecided.
- `bridgetracker/domain`: `BridgeInfo` gains `BlockHash` (populated in `sources.BridgeEventSource`); `TrackingData` gains `IsTerminal()`/`TerminallyFailed()`, hoisted out of `registry.go` so both the in-memory and SQLite adapters share the same derivation.
- `bridgetracker.Config` gains `DBPath` (empty keeps the in-memory adapters) and an `Activity` override field; `proxy/cmd/run.go` wires the SQLite-backed stores when `DBPath` is set, and `proxy/config/default.go` enables it by default (`/tmp/bridgetracker.sqlite`).

## ⚠️ Breaking Changes
- 🛠️ **Config**: none breaking — `DBPath` is additive and defaults to empty in `Config` itself; the proxy's shipped `default.go` does set it to `/tmp/bridgetracker.sqlite`, switching the proxy binary's default registry/activity store from in-memory to SQLite-backed.

## 📋 Config Updates
- 🧾 New `[Tracker].DBPath` key (proxy/config/default.go):
  ```toml
  [Tracker]
  DBPath = "/tmp/bridgetracker.sqlite"
  ```

## ✅ Testing
- 🤖 **Automatic**: new `bridgetracker/db/sqlite_registry_test.go` and `sqlite_activity_store_test.go`; full suite verified with `go build ./...`, `go vet ./...`, `golangci-lint run ./bridgetracker/... ./proxy/...`, and `go test ./bridgetracker/... ./proxy/...` (all green).

## 🐞 Issues
- Closes #1822

## 🔗 Related PRs
- None

## 📝 Notes
- Pruning/retention (PruneTerminal/PruneIdle, activity idle sweep) stays in-memory-only for now — a deliberate, temporary trade-off until the DB-side retention policy is decided.


joanestebanr:
@claude review

bridgetracker/db/sqlite_activity_store.go
```diff
@@ -0,0 +1,610 @@
...
+	return filter == types.ActivityFilterPending ||
+		filter == types.ActivityFilterReadyToClaim ||
+		filter == types.ActivityFilterError
+}
+
+// saveBridgeRow upserts entry as addr's row for item's global index
+func (s *sqliteActivityStore) saveBridgeRow(
+	addr string, item *domain.ScannedBridge, entry *domain.ActivityEntry,
+) error {
+	data, err := json.Marshal(activityBridgeData{Bridge: entry.Bridge, Claim: entry.Claim, Errors: entry.Errors})
```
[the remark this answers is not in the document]
joanestebanr: Fixed in 97f2e3ee: `saveBridgeRow` now includes `Source: entry.Source` when marshaling `activityBridgeData`, so it round-trips through a cache reload instead of decoding back as `""`. Added `TestSQLiteActivityStorePersistsSource`, which fails without the fix and passes with it.

bridgetracker/db/sqlite_registry.go
```diff
@@ -0,0 +1,646 @@
...
+	var row trackedBridgeRow
+	err := meddler.QueryRow(r.db, &row,
+		"SELECT * FROM tracked_bridge WHERE network_id = ? AND tx_hash = ?", id.NetworkID, id.TxHash.Hex())
+	if err != nil {
+		return nil, aggkitdb.ReturnErrNotFound(err)
+	}
+	if row.staleSchema() {
+		return nil, aggkitdb.ErrNotFound
+	}
+	return &row, nil
```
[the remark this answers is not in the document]
joanestebanr: Fixed in 97f2e3ee: added an optional `BlockHashVerifier` port to `sqliteRegistry`. `selectRow`'s callers now go through `selectFreshRow`, which discards (ErrNotFound, same as a stale-schema row) any row whose `origin_block_hash` no longer matches the chain's current hash at `origin_block_number` — so `getOrCreate` re-registers and re-resolves it from scratch instead of serving a reorged-out deposit's status forever. Also fixed the same gap in `create()`'s 'lost the race' fallback, which previously would have handed back the stale row unchanged since it only checked `staleSchema()`. A verifier error (RPC briefly unreachable) is treated as inconclusive, not as proof of a reorg, so it doesn't discard the whole cache on a transient hiccup. Wired up in `proxy/cmd/run.go` using the tracker's existing per-network JSON-RPC clients (`RetrieveBlockHeaders`). Added `TestSQLiteRegistryDiscardsReorgedRow` and `TestSQLiteRegistryTrustsRowOnVerifierError`.

joanestebanr:
@claude review

arnaubennassar:
@claude

arnaubennassar:
@claude

arnaubennassar:
@claude

arnaubennassar:
@claude

arnaubennassar (commented):
## Code review: SQLite-backed tracker/activity cache

Reviewed the full diff (16 files, +2594/-48). 11 findings below, posted inline: **1-3 blocking**, **4-7 correctness**, **8-11 lower**.

The headline concern is that findings 1-3 all come from the same decision — `proxy/config/default.go` turns the SQLite path on by default while retention/pruning is still a no-op and the reorg check runs on every read. Shipping `DBPath` as opt-in would defuse all three without changing any of the storage code.

### Things I checked and cleared

- **JSON round-trip of the persisted shapes**: `types.Duration`, `ErrorStep`, `CertificateData`, `agglayertypes.CertificateStatus` all decode back correctly; the `json:"-"` fields live only on `GERData`, which is never persisted in `BridgeStepPath`.
- **`ON DELETE CASCADE` in `FlushActivity` works** — `db.NewSQLiteDB` sets `_foreign_keys=on` (plus WAL and `_busy_timeout=30000`, so the two `*sql.DB` handles on one file don't deadlock).
- **`BridgeInfo.BlockHash`** is populated on the only production construction site (`sources/bridge_event.go:146`) and `BridgeInfo` is not serialized into any API response, so no wire-format change.
- **Concurrent read-modify-write** on `UpdateTrackingStep`/`UpdateTrackingBridgeTx` is safe in practice: `Engine.tick` and `resolveTriggered` run on the same goroutine and parallelise only across distinct ids.
- **Index planner behaviour**: `idx_tracked_bridge_terminal_since` does get picked for `GetTrackerActives` (verified with `EXPLAIN QUERY PLAN`), but `terminal_since = 0` is the selective side here, so it is *not* a repeat of the PR-1784 planner hijack.
- **`go build ./...`** on the branch is clean.

proxy/config/default.go
```diff
@@ -48,6 +48,11 @@ AllowCredentials = false
 MaxAge = "12h"
 
 [Tracker]
+# DBPath: SQLite file the supervised-bridges registry persists to. Set by default so the
+# registry survives a restart (an already-resolved bridge doesn't need re-resolving); empty
+# it out to fall back to the in-memory adapter instead.
+DBPath = "/tmp/bridgetracker.sqlite"
```
arnaubennassar: **[blocking 1/11] Default `DBPath` lands in a world-writable shared directory.**

`DBPath = "/tmp/bridgetracker.sqlite"` ships enabled by default. `/tmp` is a shared, world-writable directory and the filename is fixed, so any unprivileged local user can pre-create `/tmp/bridgetracker.sqlite` (or a symlink) before the proxy starts; the proxy then opens it and serves attacker-controlled tracking/claim state, or writes through the symlink.

Secondary: `/tmp` is tmpfs on many hosts and is swept by `systemd-tmpfiles`, so the "survives a restart" promise silently doesn't hold, and two proxy instances on one host share one file.

A path under the proxy's own data dir, and/or opt-in rather than default-on, is the safe choice.

bridgetracker/db/sqlite_registry.go
```diff
@@ -0,0 +1,718 @@
...
+	defer r.countMu.Unlock()
+	return r.numEntries
+}
+
+// PruneTerminal implements domain.SupervisedStore. Deliberately a no-op for now: unlike the
+// in-memory adapter, this store never deletes a persisted row on its own — pruning/retention
+// stays an in-memory-only concern until a DB-side retention policy is decided (see issue #1822).
+// tracked_bridge is left to grow unbounded for the time being; this is a conscious, temporary
+// trade-off, not an oversight
+func (r *sqliteRegistry) PruneTerminal(time.Time) (int, error) {
```
arnaubennassar, line 731: **[blocking 2/11] No-op pruning + default-on `DBPath` = unbounded, ever-growing RPC load and a permanently full registry.**

`PruneTerminal`/`PruneIdle` (and `sqlite_activity_store.go:360`'s `sweepIdle`) are no-ops while `proxy/config/default.go` now enables `DBPath` by default.

Concrete failure 1: a bridge the user never claims stays non-terminal forever, so `GetTrackerActives` keeps returning it and `Engine.tick` re-resolves it every `PollInterval` (30s default) — forever. With the in-memory adapter `PruneIdle(30m)` dropped it. Outbound bridge-service/agglayer/RPC load therefore grows monotonically with the lifetime of the process's database, never with current demand.

Concrete failure 2: `numEntries` only ever grows, so once 100 000 rows accumulate (`MaxTrackedBridges` default) `create` returns `ErrRegistryFull` for every new tx permanently — the only recovery is deleting the DB file.

This is documented as a "temporary trade-off", but it is not safe to make it the shipped default.

bridgetracker/db/sqlite_registry.go
```diff
@@ -0,0 +1,718 @@
...
+
+// rowIsStale reports whether row can no longer be trusted as-is and must be discarded in favor
+// of a fresh registration: either its schema is stale (see staleSchema, already filtered out of
+// row by selectRow but not by selectRowIgnoringSchema) or, once resolved and with a
+// BlockHashVerifier configured, its persisted origin block hash no longer matches the chain's
+// current canonical hash at origin_block_number — a bridge accepted on a block since reorged
+// out. A verifier error (e.g. the origin RPC briefly unreachable) is not itself evidence of a
+// reorg, so it is logged and row is trusted as fresh rather than discarding the whole cache
+// over a transient hiccup
+func (r *sqliteRegistry) rowIsStale(id domain.TrackingID, row *trackedBridgeRow) bool {
```
arnaubennassar, line 245: **[blocking 3/11] `rowIsStale` does a blocking RPC on every read path.**

`rowIsStale` issues a synchronous `CanonicalBlockHash` RPC on *every* read path (`Get`, `GetAndAwait`, `Subscribe` -> `getOrCreate` -> `selectFreshRow`) for any row with `origin_block_number != 0`.

`sqliteActivityStore.refresh` (`sqlite_activity_store.go:528`) calls `supervised.Get(id, true)` once per unclaimed bridge, sequentially — so `GET /activity/from/{addr}` for an address with 50 cached bridges now fires 50 extra sequential `eth_getBlockByNumber` calls, each allowed up to `blockHashVerifyTimeout` (3s), i.e. up to ~150s of added latency inside one HTTP handler. There is no memoisation of the canonical hash and no per-row cooldown.

Also, the ctx is `context.Background()` (line 251), so the work is not cancelled when the client disconnects.

A reorg check that runs once at startup/reload (as the doc comment describes) rather than on every read would avoid this.

proxy/cmd/run.go
```diff
@@ -143,6 +144,49 @@ func runProxy(
...
+		return common.Hash{}, err
+	}
+	result, err := client.RetrieveBlockHeaders(ctx, []uint64{blockNumber}, 1)
+	if err != nil {
+		return common.Hash{}, err
+	}
+	if header, ok := result.Headers[blockNumber]; ok {
+		return header.Hash, nil
+	}
+	return common.Hash{}, result.Errors[blockNumber]
```
arnaubennassar: **[correctness 4/11] `CanonicalBlockHash` can return `(zero hash, nil error)` and silently nuke a good cache row.**

`return common.Hash{}, result.Errors[blockNumber]` returns `(zero hash, nil error)` whenever `blockNumber` appears in neither `result.Headers` nor `result.Errors`. `rowIsStale` treats a nil error as authoritative, compares `common.Hash{}` against the persisted hash, sees a mismatch, and discards a perfectly good cache row (logging a bogus `changed hash (0x... -> 0x000...)`).

Add an explicit guard, e.g. `if err == nil { err = fmt.Errorf("no header for block %d", blockNumber) }`.

bridgetracker/db/sqlite_registry.go
```diff
@@ -0,0 +1,718 @@
...
+// rowIsStale: stale schema or reorged-out origin block) is overwritten in place instead of
+// inserted as a new one, so it never counts twice against maxEntries
+func (r *sqliteRegistry) create(id domain.TrackingID) (*trackedBridgeRow, error) {
+	now := r.now()
+	row, err := r.freshRow(id, now)
+	if err != nil {
+		return nil, err
+	}
+
+	r.countMu.Lock()
```
arnaubennassar, line 344: **[correctness 5/11] Capacity gate precedes the stale-row overwrite, contradicting this function's own doc.**

The `numEntries >= maxEntries` gate runs before the stale-row overwrite path, contradicting `create`'s doc ("A stale row ... is overwritten in place instead of inserted as a new one, so it never counts twice against maxEntries").

At capacity, a row that is stale (schema bump, or origin block reorged out) can never be replaced: `getOrCreate` gets `ErrNotFound` from `selectFreshRow`, `create` returns `ErrRegistryFull`, and that tx is permanently unqueryable. Combined with finding 2 (the cap is reached eventually and never released) this is reachable in normal operation.

bridgetracker/db/sqlite_registry.go
```diff
@@ -0,0 +1,718 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("listing active tracked bridges: %w", err)
+	}
+
+	active := make([]*domain.TrackingData, 0, len(rows))
+	for _, row := range rows {
+		if row.staleSchema() {
+			continue
+		}
+		tracking, err := row.trackingData()
```
arnaubennassar, line 673: **[correctness 6/11] One undecodable row wedges the whole engine, permanently.**

`GetTrackerActives` returns an error if *any single* row's `data` fails to JSON-decode. `Engine.tick` (`engine.go:221`) logs and `return`s on that error, so one undecodable row stops resolution for **every** bridge, on every tick, permanently — and because nothing prunes, the bad row never goes away. The in-memory adapter has no such failure mode.

Skipping + logging the bad row (as is already done for `staleSchema()` two lines above) and/or deleting it would self-heal. `GetNetworks:654` has the same shape. Likewise `Get` (line 407) returns the decode error forever for that id, with no fallback to re-create.

bridgetracker/db/sqlite_registry.go
```diff
@@ -0,0 +1,718 @@
...
+	originBlockHash := common.Hash{}
+	if tx.Info != nil {
+		originBlockNumber = tx.Info.BlockNumber
+		originBlockHash = tx.Info.BlockHash
+	}
+
+	if _, err := r.db.Exec(
+		`UPDATE tracked_bridge SET
+			claim_status = ?, origin_block_number = ?, origin_block_hash = ?,
+			updated_at = ?, last_access = ?, terminal_since = ?, data = ?
```
arnaubennassar: **[correctness 7/11] Engine writes bump `last_access`, defeating the idle-eviction anchor.**

`UpdateTrackingBridgeTx` writes `last_access = ?` on every engine-driven update. `last_access` is documented as the *idle eviction anchor* ("unix seconds this row was last read"), and the in-memory adapter only stamps `lastAccess` on `Get`/`GetAndAwait`/`Subscribe`.

Since the engine touches every active bridge every poll tick, `last_access` can never age out — so when `PruneIdle` is eventually implemented on top of this column it will evict nothing, silently reproducing finding 2.

bridgetracker/db/sqlite_registry.go
```diff
@@ -0,0 +1,718 @@
...
+			}
+		}
+	}
+}
+
+// UpdateTrackingBridgeTx implements domain.SupervisedStore, notifying every subscriber of id.
+// Recomputes claim_status and, once terminal_since is first stamped, leaves it untouched
+// forever after — see domain.TrackingData.IsTerminal
+func (r *sqliteRegistry) UpdateTrackingBridgeTx(id domain.TrackingID, tx domain.TrackingBridgeTx) error {
+	row, err := r.selectRow(id)
```
arnaubennassar, line 557: **[lower 8/11] Real DB errors are laundered into `ErrTrackingNotFound`.**

This line and `UpdateTrackingStep:576` both map *any* `selectRow` error to `domain.ErrTrackingNotFound`, including genuine DB errors (`database is locked`, I/O error, corrupt page). `Engine.persist` then logs "failed to persist" and aborts the batch, and API callers see "bridge not tracked" for a bridge that *is* tracked.

Only `errors.Is(err, aggkitdb.ErrNotFound)` should map to `ErrTrackingNotFound`; the rest should be wrapped and propagated.

bridgetracker/db/sqlite_activity_store.go
```diff
@@ -0,0 +1,612 @@
...
+		 ON CONFLICT (from_address) DO UPDATE SET
+			schema_version = excluded.schema_version, updated_at = excluded.updated_at,
+			last_access = excluded.last_access, scan_state = excluded.scan_state`,
+		addr, activitySchemaVersion, now.Unix(), now.Unix(), empty)
+	return err
+}
+
+// recordScanWarnings merges warnings into addr's scan_state (see networkScanState); a no-op
+// when warnings is empty, so a fully successful scan never rewrites scan_state
+func (s *sqliteActivityStore) recordScanWarnings(addr string, warnings []domain.ActivityWarning, now time.Time) error {
```
arnaubennassar, line 724: **[lower 9/11] `scan_state` is written but never read back for anything.**

`scan_state` is written by `recordScanWarnings` and read back only by `recordScanWarnings` itself to merge. Nothing ever surfaces it: `GetActivity` returns only the warnings from the current scan.

The migration comment (`bridgetracker0002.sql`) claims the per-network error "survives across calls", but a persisted error is invisible to every subsequent request, so the column is pure write amplification today.

proxy/cmd/run.go
```diff
@@ -158,9 +202,17 @@ func runTracker(
 	if err := trackerCfg.Validate(); err != nil {
```
arnaubennassar, line 222: **[lower 10/11] Neither store is ever `Close()`d.**

Neither `sqliteRegistry.Close()` nor `sqliteActivityStore.Close()` is ever called (registry constructed here, activity store at line 258) — two `*sql.DB` handles on the same file are leaked for the process lifetime, and WAL is never checkpointed on shutdown.

Both types define `Close()` specifically for this; `runTracker` should register them on the shutdown path.

proxy/config/default.go
```diff
@@ -48,6 +48,11 @@ AllowCredentials = false
...
+
 # RetentionPeriod: how long a terminal bridge (finished, or failed to ever resolve) stays
```
arnaubennassar, line 65: **[lower 11/11] Retention/idle/cap comments now describe behaviour that no longer happens.**

The `RetentionPeriod`, `IdleTimeout` and `MaxTrackedBridges` comments in this file still describe eviction behaviour ("stays queryable before the tracker forgets it", "RetentionPeriod and IdleTimeout are what keep the registry under it during normal operation") that no longer happens now that `DBPath` is set by default.

`bridgetracker/config.go:150-156` has the same drift ("Only applies to the default in-memory adapter" is now the non-default path). Operators tuning these knobs will get no effect.

joanestebanr:
Addressed all 11 findings from the review above in 6192f3d2.

**Blocking 1-3** (unsafe `/tmp` default, no-op pruning + default-on, per-read reorg RPC): per your own summary, all three come from `DBPath` being default-on — fixed at the root by reverting `proxy/config/default.go` to `DBPath = ""` (opt-in, in-memory adapter by default, exactly as before this PR). Comments on `RetentionPeriod`/`IdleTimeout`/`MaxTrackedBridges` (both there and in `bridgetracker/config.go`) now say which of them only apply to the in-memory adapter.

**4** (`CanonicalBlockHash` zero-hash/nil-error): fixed — `proxy/cmd/run.go` now returns an explicit error when a block is absent from both `Headers` and `Errors`, instead of `(zero hash, nil)`.

**5** (capacity gate precedes stale-row overwrite): fixed — `create()` now checks for an existing row for `id` *before* the `maxEntries` gate; if that row is stale it's overwritten in place regardless of capacity, since replacing it can never grow the table. Added `TestSQLiteRegistryOverwritesStaleRowEvenAtCapacity`.

**6** (one undecodable row wedges the engine): fixed — `GetTrackerActives`/`GetNetworks` now skip and log an undecodable row instead of failing the whole call, and `rowIsStale` now also treats a decode failure as stale, so `Get`/`GetAndAwait`/`Subscribe` self-heal a corrupted row by discarding and re-registering it. Added `TestSQLiteRegistryCorruptRowIsSkippedAndSelfHeals`.

**7** (engine writes bump `last_access`): fixed — `UpdateTrackingBridgeTx` no longer stamps `last_access`; only an actual read (`Get`/`GetAndAwait`/`Subscribe`) does. Added `TestSQLiteRegistryUpdateTrackingBridgeTxDoesNotBumpLastAccess`.

**8** (real DB errors laundered into `ErrTrackingNotFound`): fixed — `UpdateTrackingBridgeTx`/`UpdateTrackingStep` now only map `errors.Is(err, aggkitdb.ErrNotFound)` to `domain.ErrTrackingNotFound`; anything else propagates wrapped. Added `TestSQLiteRegistryUpdateMethodsPropagateRealDBErrors`.

**9** (`scan_state` written but never read): fixed — `GetActivity` now returns the persisted `scan_state` (via a new `persistedWarnings`) as its warnings, so a network's last known failure surfaces on later calls too, not just the one that hit it. Added `TestSQLiteActivityStoreWarningSurvivesAcrossCalls`.

**10** (stores never `Close()`d): fixed — `runTracker` now closes both the registry and the activity store on shutdown via a small `closeOnShutdown` helper (type-asserts `io.Closer`, no-op for the in-memory adapters).

**11** (stale retention comments): covered by the DBPath fix above — comments now reflect that those knobs are in-memory-adapter-only.

`go build ./...`, `go vet ./bridgetracker/... ./proxy/...`, `golangci-lint run ./bridgetracker/... ./proxy/...`, and `go test ./bridgetracker/... ./proxy/...` all green.

arnaubennassar (approved):

PR #1854, opened 2026-09-18
fix: per-destination autoclaim cursors, bridge-service health sync status, bridgetracker emitter verification (#1651, #1689, #1751)

### LINKED ISSUE #1651: Add edge cases for L1 -> L2 autoclaim
- Add a new claimer while others have been operating (test cursor)

### LINKED ISSUE #1689: bridge service: imporve /health endpoint
check not only that the service is reachable, but that it's synced and other checks to ensure healthiness

Bridge service provide `/sync-status` end-point with that information, but the idea is that the `/health` returns a sync status like "error", "pending", "done"

### LINKED ISSUE #1751: bug: tracker bridges: Verify the BridgeEvent emitter address
For any successful transaction containing a log with this signature, the source accepts the log without checking l.Address against the canonical bridge contract for the requested network. An unrelated or malicious contract can emit the same ABI-compatible event and be reported as a real bridge, after which the tracker queries GER and claim state using attacker-controlled destination and deposit values; resolve the network's bridge address and reject logs emitted elsewhere.

arnaubennassar:
## 🔄 Changes Summary
- **#1651 — autoclaim cursors**: `autoclaim_ler_cursor` is re-keyed to `(source_network, destination_network)` so each destination's bridge detector tracks its own LER cursor independently, instead of sharing one cursor per source. A claimer added after others no longer inherits/corrupts an unrelated destination's progress.
- **#1689 — bridge-service health**: the `/health` endpoint now reports per-source sync status (L1 info tree, injected GERs, claims, etc.) instead of pure liveness, backed by a short-lived cache (`HealthCheckCacheTTL`, default 2s) and a bounded compute timeout (`DefaultHealthCheckComputeTimeout`, 3s) so repeated probes stay cheap and a black-holed dependency can't hang the endpoint.
- **#1751 — bridgetracker emitter check**: a `BridgeEvent`-shaped log is now rejected unless it actually came from the configured bridge contract, resolved from `Tracker.BridgeAddrs` or, when a network is absent from that map, via `bridgeservicefinder.Finder.BridgeAddress`. Previously any contract could emit a matching log and have it silently accepted as a real bridge transaction.

## ⚠️ Breaking Changes
- 🔌 **(#1751)**: any network absent from `Tracker.BridgeAddrs` now requires `bridgeservicefinder.Finder.BridgeAddress` to resolve the bridge address. A resolved-address mismatch now correctly rejects a real bridge tx as `ErrBridgeTxNotABridge`, where previously a `BridgeEvent`-shaped log from **any** emitter was silently accepted.
- #1689 is additive only: `/health` always returns HTTP 200 and existing callers are unaffected.
- #1651 adds a schema migration (`autoclaim0003`) — see Notes for the upgrade/downgrade procedure.

## 📋 Config Updates
- No new TOML fields are exposed.
- `HealthCheckCacheTTL` (default 2s) and `DefaultHealthCheckComputeTimeout` (3s) are Go constants/struct defaults, not wired to any TOML section today.
- `Tracker.BridgeAddrs` is unchanged as a config field — only its semantics changed: it is now an override rather than the only source of the bridge address (the Finder now backstops networks it doesn't cover).

## ✅ Testing
- 🤖 **Automatic**: targeted e2e selection run and green against `anvil-2chains` (4/4):
  - `TestAutoClaimClaimerAddedAfterOthers` (#1651)
  - `TestBridgeServiceHealthSyncStatus` (#1689)
  - `TestBridgeTrackerL1ToL2` (regression)
  - `TestBridgeTrackerNotABridge` (#1751)
  - Full unit test suite green (`make test-unit`).
  - `make build` and `make lint` clean.
  - The full ~60-minute e2e matrix was left to CI rather than run locally.

## 🐞 Issues
- Closes #1651
- Closes #1689
- Closes #1751

## 🔗 Related PRs
- N/A

## 📝 Notes
- **Migration `autoclaim0003`** re-keys `autoclaim_ler_cursor` to `(source_network, destination_network)` and parks existing rows in `autoclaim_ler_cursor_legacy`. The detector seeds the new per-destination rows from the legacy cursor on first poll, for whatever destinations are configured at that moment, then deletes the legacy row.
- **Upgrade in two steps**: upgrade the binary first, and add any new claimers in a **later** restart. A claimer added in the **same** restart as the upgrade will be seeded from the parked legacy LER and will **not** backfill (this is unavoidable — the destination set is only known at runtime). A `Warnf` log line is emitted for each legacy seed so this is observable.
- **Downgrade**: binary-only rollback is **not** supported. Run `autoclaim0003`'s `Down` migration first — otherwise the old per-source query will silently adopt one destination's cursor and then hard-error on every subsequent save. Also note `RunMigrationsDown(..., 1)` rolls back `basedb0001`, not `autoclaim0003`, because migration IDs sort lexicographically — target the migration explicitly by name.
- **Backfill volume**: a newly added destination with `StartL1Block = 0` will request full history in a single poll; set `StartL1Block` to bound this.
- **Deliberately left out of this PR**:
  - `swag` swagger generation is nondeterministic for bridgetracker (duplicated `time.Duration` enum entries on regeneration) — worth its own ticket.
  - `docs/autoclaim.md` documents `AggClaim.BridgeServiceFinder.HealthCheckPath`'s default as `/health`, while the code default is actually `/` — this is pre-existing and unrelated to this PR.


arnaubennassar:
@claude

arnaubennassar:
Thanks for the review. Addressed in f2b7781a. Taking each point in turn — three changes made, three declined with reasoning, and two of the "required" items are based on code that is in the diff but wasn't read.

## 🔴 Required issues

### 2 & 10 — Health cache synchronization

`healthCheckCache` **is** in this diff, in the same file the comment cites (`bridgeservice/bridge.go`, ~line 1678 onwards). It is a `sync.Mutex` plus an in-flight channel:

```go
type healthCheckCache struct {
	ttl      time.Duration
	mu       sync.Mutex
	value    healthCheckCacheEntry
	expires  time.Time
	inflight chan struct{}   // non-nil while a computation runs; closed when it finishes
}
```

`getOrCompute` takes `mu`, serves the value if `time.Now().Before(h.expires)`, otherwise joins the existing `inflight` channel if one is set, and only starts a computation when it is not. Concurrent callers therefore cannot trigger multiple simultaneous `computeSyncStatus` calls — that is the single-flight.

The "minor gap" in the coverage assessment isn't accurate either; these tests already exist in `bridgeservice/bridge_test.go`:

- `cache hit within TTL - underlying Bridger called once for N requests`
- `cache expiry - recomputes after TTL elapses`
- `concurrent requests single-flight into one computation` (blocks the first `GetContractDepositCount` so a single-flight failure shows up as a second call)
- `TestHealthCheckCache_PanicInComputeDoesNotWedgeCache`

That last one is worth flagging: an earlier revision of this PR *did* have a real bug here — `compute()` was called outside a `defer`, so a panic skipped `h.inflight = nil` / `close(ch)` and permanently wedged both `/` and `/health` with an unbounded goroutine leak. It was found in review and fixed before this PR opened; the test hangs the full 2s timeout if the `defer` is removed. Verified again just now under the race detector:

```
$ go test -race -count=1 ./bridgeservice/ -run TestHealthCheck
ok  	github.com/agglayer/aggkit/bridgeservice	1.241s
```

No code change.

### 3 — SQL injection

Agreed there's no vulnerability — the query is parameterised, as the comment itself concludes. I took the `LIMIT 1` suggestion anyway since it costs nothing and makes the single-row intent explicit (`source_network` is the legacy table's primary key, so it already matched at most one row). Changed.

### 1 — Migration test count

Declining, deliberately. The exact count is a tripwire, not an oversight: `require.Len(t, migrations, 3)` is what forces whoever adds `autoclaim0004` to come back to this test and add the matching `migrations[3].ID` / `SQL` assertions. Relaxing it to `GreaterOrEqual` is precisely what would let a new migration land with no assertions about its ID or ordering — and ordering matters here, because these IDs are applied in sequence and sort lexicographically (see the downgrade note below, where that sort order is load-bearing).

This also isn't new: the test asserted `2` before this PR and was bumped to `3` as part of it. That's the intended workflow.

## 🟡 Suggested improvements

### 4 — Health check timeout not configurable

Good catch, and a real asymmetry: `HealthCheckCacheTTL` was already a `Config` field with a Go default, while the compute timeout read the package var directly. Fixed — `Config.HealthCheckComputeTimeout` now exists and falls back to `DefaultHealthCheckComputeTimeout` (3s) when unset.

New test `TestHealthCheckHandler_ComputeTimeoutIsConfigurable` mutation-verified: if the constructor ignores the configured value, it fails after exactly the 3.00s default instead of the configured 30ms.

One correction to the premise: the 3s default was never "sized for large paginated response bodies" — that's the description of `ReadTimeout` (up to 5m), which is exactly what the health path must *not* inherit. That reasoning is in `DefaultHealthCheckComputeTimeout`'s doc comment.

Still not wired to TOML, and deliberately so: there is no `[BridgeService]` TOML section at all — `bridgeservice.Config` is assembled ad hoc in `cmd/run_bridgeservice.go` from `cfg.PublicREST.*`. Introducing one is a config-surface change beyond this PR's scope. `docs/bridge_service.md` now states this accurately (it previously called `HealthCheckCacheTTL` a "Go constant", which was wrong — it's a `Config` field).

### 5 — Bridge address resolver error classification

Declining for now, because the distinction doesn't exist at the source. `finder.BridgeAddress` has no "network not supported" path: it checks the per-network override, then falls back to the rollup manager's own `BridgeAddress()` on-chain read. The only error it returns is a transport/RPC error:

```go
return common.Address{}, fmt.Errorf("reading rollup manager's bridge address: %w", err)
```

Classifying that as permanent would reintroduce the exact failure mode this PR fixes, in reverse: a transient RPC blip would permanently reject a legitimate bridge tx as "not a bridge". Treating it as transient-and-retry is the safe direction. If a genuinely permanent "unsupported network" error is ever added to the finder, classification should be added at the same time — worth a follow-up issue against `bridgeservicefinder`, not this PR.

### 6 — Legacy seed log level

Keeping `Warnf`, deliberately. The asymmetry matters: if a claimer *was* added in the same restart as the upgrade, it silently inherits the parked cursor and never backfills its history — unrecoverable without operator action, and this log line is the only place that becomes diagnosable after the fact. Under-warning costs silent permanent data loss for that destination; over-warning costs one line per source, exactly once ever (the legacy row is deleted by the same transaction, so it never fires again). The message already names the destinations and points at the docs section.

### 7 — Test contract deployment warning

Added. The contract header already described it as "throwaway e2e test tooling", and it lives under `test/contracts/` alongside the other mocks, but an explicit line is cheap insurance:

```solidity
// TEST-ONLY: this contract exists purely to be rejected by that check. It must never be
// deployed to a production or public network -- it deliberately emits bridge-shaped logs it
// has no authority to emit, and nothing here is part of the aggkit runtime.
```

I left `compile.sh` alone — every contract it builds is test-only by virtue of the path, so a warning on just this one would imply the others aren't.

### Docs — "Before You Upgrade" checklist

Reasonable; the upgrade/downgrade material is already in the PR body and `docs/autoclaim.md`, so I'd rather not fork it into a third copy that can drift. The two operational rules, restated once: **upgrade first and add new claimers in a later restart**, and **never roll the binary back without running `autoclaim0003`'s Down first** (and target that migration explicitly — `RunMigrationsDown(..., 1)` rolls back `basedb0001`, because IDs sort lexicographically).

## Verification

```
make lint                    → 0 issues
go build ./... && go vet ./... → clean
go test -count=1 ./bridgeservice/... ./autoclaim/... → all ok
go test -race ./bridgeservice/ -run TestHealthCheck → ok
```

joanestebanr:
@claude review

arnaubennassar:
Thanks — addressed the blocking item and two of the suggestions in d9d68417.

## 🔴 Issue #5 — claimer added during upgrade (the merge blocker)

Implemented, with one deliberate deviation from the proposal.

**What landed.** New `Storage.PendingLegacyLERCursorSources` reports the sources whose pre-upgrade cursor is still parked in `autoclaim_ler_cursor_legacy`. `autoclaim/runtime.Start` calls it and, while any remain, logs:

```
autoclaim0003 upgrade pending: source networks [3 9] still hold a pre-upgrade LER cursor that the next
poll will fan out to the currently enabled destination claimers [20 22]. Any of those destinations added
in THIS restart will inherit that cursor and will NOT backfill history before it -- which cannot be
undone by restarting again. If you intended to add a claimer, stop now, run this upgrade with the
previous claimer set, and add the new claimer in a later restart (see docs/autoclaim.md 'Legacy upgrade seed').
```

It fires only while legacy rows remain — first start after upgrading, never again once a poll consumes them — so it's a reliable "you are inside the risky window" signal rather than routine noise, and it names the exact destination set about to inherit the cursor. That's the last moment the operator can still stop before the seed is irreversible.

**The deviation, and why.** The proposal's step 2 was to warn or fail *"if any new claimers are configured that weren't present in the last pre-upgrade run"*. That comparison can't be made: the pre-upgrade destination set is precisely the runtime fact that no pre-`autoclaim0003` database records. `autoclaim_ler_cursor` was keyed by `source_network` alone — that's the bug — so there is nothing on disk saying which destinations were claiming from a source before the upgrade. (`autoclaim_bridge_cursor` has per-destination rows, but only for the L1→L2 detector, so it can't answer the L2→Lx question, and inferring from it would be a heuristic inside a data-loss warning — worse than no answer.)

So a **fatal** is also off the table: gated on a comparison we can't make, it would either never fire or block every legitimate upgrade — including the overwhelmingly common one where no claimer changed. Warning with the exact source and destination lists is the strongest honest signal available. Both the code comment and `docs/autoclaim.md` now state this reasoning explicitly.

**Tests** (`TestWarnOnPendingLegacyLERCursors`, `TestPendingLegacyLERCursorSources`), both mutation-verified:
- removing the `len(sources) == 0` guard → fails (would warn in the steady state, destroying the signal's meaning)
- listing all claimers instead of only enabled ones → fails on `[20 21 22]` vs `[20 22]`
- also covers: silent after seeding, storage without the capability, and a lookup error warning without ever blocking startup

## 🟡 Issue #6 — readiness probe

Documented rather than adding an endpoint. A second endpoint that can return non-2xx recreates the exact hazard the always-200 design avoids: `bridgeservicefinder` gates routing on a 2xx at `DefaultHealthCheckPath` (`/`), so any probe path that *can* go non-2xx risks being pointed at by a finder or ingress check and evicting an instance that is merely catching up. `docs/bridge_service.md` now says that outright and shows how to gate on the body instead.

One correction to the suggested YAML: `httpGet` and `exec` are mutually exclusive within a single probe, so that snippet wouldn't apply — the `exec` block silently wins and `successThreshold: 1` is already the default. The documented form:

```yaml
readinessProbe:
  exec:
    command: ["/bin/sh", "-c", 'curl -sf localhost:8080/health | jq -e ".sync_status != \"error\""']
  periodSeconds: 10
  failureThreshold: 3
livenessProbe:
  httpGet:
    path: /health
    port: 8080
```

Liveness stays on plain `httpGet`: it should restart a wedged process, not a lagging one.

## 🟢 Observation #8 — `HealthCheckPath` doc default

Fixed, since two passes have now flagged it and it's a one-line correction in a file this PR already touches. `docs/autoclaim.md` said `/health`; the code default is `bridgeservicefinder.DefaultHealthCheckPath` = `/`. The row now also notes the bridge service serves the same always-200 handler on both paths — which this PR is what makes true.

## Declined, with reasoning

**#1 — migration test.** The proposed variant keeps `migrations[0..2]` assertions but relaxes the length, which is exactly what drops the tripwire: `autoclaim0004` would then land with nothing asserting its ID or position. Ordering is load-bearing here (IDs sort lexicographically, which is what makes `RunMigrationsDown(..., 1)` roll back `basedb0001` instead of the intended migration), so I'd rather a new migration be forced through this test than pass silently. One deliberate line of friction per migration.

**#2 — TOML wiring.** Agreed on the value, but there is no `[BridgeService]` TOML section at all today — `bridgeservice.Config` is assembled ad hoc in `cmd/run_bridgeservice.go` from `cfg.PublicREST.*`. Introducing one is a config-surface change with its own review surface, not a rider on a three-issue PR. Both fields are `Config` fields now, so that follow-up is a plumbing change with no further refactoring needed.

**#3, #4, #7** — no action needed; noting agreement with the re-assessment.

## Verification

```
make lint                                        → 0 issues
go build ./... && go vet ./...                   → clean
go test -count=1 ./autoclaim/... ./bridgeservice/...  → 17 packages ok, 0 failures
```

On the earlier red `test-unit`: that was `l1infotreesync/TestProcessor_ConcurrentProcessBlockAndReorg`, a package this PR doesn't touch. The identical test failed on `develop` at this PR's base commit `3210b09c` (run 35199716467), same line, same `sql: transaction has already been committed or rolled back`. The reorg goroutine's `cancel()` can land mid-transaction, so `database/sql` rolls the tx back and the next statement returns `sql.ErrTxDone` — but the test only tolerates `"context canceled"`. Pure timing, hence CI-load-dependent. Green on re-run; worth its own one-line fix in that package.

joanestebanr (approved):
LGTM

